PR #17, opened 2019-04-14
fix vision images

smithomaster:


smithomaster:
ok, should ge good to go, added in correct code blocks and notes also

PR #32, opened 2019-05-16
CameraServer class: Reformat to styleguide and update

PeterJohnson:
- Use digraph for graphs
- Make code examples more complete and include C++ versions


Daltz333:
Looks great! Could you put a space in-between each text segment and a section and you're good to go!
The space separating content and headers was a recommendation by Tyler. 
```
Section
=====

Text here
```

PR #33, opened 2019-05-16
Identify and process targets: clean up

PeterJohnson:
- Use latex for math equations
- Use |deg| for degrees to unbreak latex output


Daltz333:
Also looks good. Could you put a space separating content and headers? Thanks :)

PR #48, opened 2019-05-19
Update FIRST branding on docs

Daltz333:


AustinShalit:
Instead of the FIRST logo, it should be the FRC logo?  That way FTC teams don't come looking for info.

PR #52, opened 2019-05-19
Update branding

Daltz333:
Was previously prematurely merged 😅


Kevin-OConnor:
I'm out of office for 2 weeks so wont be able to take a look. As long as it's one of our official logos it should be fine for now. I have some other style stuff I was playing with on my machine to make it look close to the other FIRST pages that I can work on when I get back.

PR #60, opened 2019-05-20
Fetch the WPILib repository when running build script

Daltz333:
This PR will automatically fetch (and update) WPILib to a local folder relative to conf.py. This PR should undergo some review on whether we actually need it or not. 

**Advantages:** 
- Can reference files and examples straight from the repository.

**Disadvantages:**
- Increased local repo size
- Adds a bit of bloat to conf.py

PS: Ignore the extra commits, those have already been merged and git is just being weird.


AustinShalit:
@Daltz333 Does sphinx get mad if an imported file does not exist or it cannot find the spot to begin at?  Maybe we should change one example in the docs to use this as an example of how it would work.

Daltz333:
Makes the repo have too much bloat. An alternative approach would be creating a sphinx extension to do this from a remote repository.

PR #63, opened 2019-05-25
Git Basics

Daltz333:
Will include the basics of git
- Creating a repo
- Commits
- Merges
- Branches
- Resetting head
- Forks


ThadHouse:
This needs to have a warning about not changing the default gitignore that the the templates include. I've seen too many teams remove it or change it because they think it's wrong, and it goes bad.

source/docs/software/basic-programming/git-getting-started.rst
```diff
@@ -0,0 +1,250 @@
...
+- Ability to manage *commits* in various ways, including combining them
+- Various other features, see `here <https://git-scm.com/about>`__
+
+Prerequisites
+-------------
+
+You have to download and install Git from the following links:
+
+- `Windows <https://git-scm.com/download/win>`_
+- `Mac OS X <https://git-scm.com/download/mac>`_
```
ArchdukeTim: Mac comes with git I believe, and if not, just typing `git` into terminal will prompt you to install Xcode command line tools, which installs git.

source/docs/software/basic-programming/git-getting-started.rst
```diff
@@ -0,0 +1,250 @@
...
+- `Linux <https://git-scm.com/download/linux>`_
+
+.. note:: You may need to add Git to your path
+
+Git Vocabulary
+--------------
+
+Git is composed of several primary elements:
+
+- *Repository:* the actual local of your code, including a ``.git`` folder in the root directory
```
ArchdukeTim: “The actual local of your code” doesn’t sound correct

source/docs/software/basic-programming/git-getting-started.rst
```diff
@@ -0,0 +1,250 @@
...
+
+Repository
+----------
+
+A Git repository is a data structure containing the structure, history, and files of a project.
+
+Git repositories consist of:
+
+- A ``.git`` folder. This folder contains the various information about the repository. 
+- A ``.gitignore`` file. This file contains the files or directories that you do *not* want included when you commit.
```
ArchdukeTim: To be pedantic, a repo doesn’t *need* a gitignore file

source/docs/software/basic-programming/git-getting-started.rst
```diff
@@ -0,0 +1,250 @@
...
+- Files and folders. This is the main content of the repository.
+
+Creating the repository
+^^^^^^^^^^^^^^^^^^^^^^^
+
+You can store the repository locally, or through the cloud. `Github <https://github.com/>`_ is a popular free hosting service. Numerous developers use it, and that's what this tutorial will use.
```
ArchdukeTim: Locally ~or~ and in the cloud

ArchdukeTim: Might be better to say locally or remotely, which could be the cloud, but also on other medium like a flash drive (which is useful to know for using git at competitions)

source/docs/software/basic-programming/git-getting-started.rst
```diff
@@ -0,0 +1,250 @@
...
+This command retrieves the history or commits from the remote repository. When the remote contains work you do not have, it will attempt to automatically merge. See :ref:`docs/software/basic-programming/git-getting-started:Merging`.
+
+Run: ``git pull``
+
+Git Add
+^^^^^^^
+
+This command adds a selected file(s) to a commit. To commit every file/folder that isn't excluded via *gitignore*.
+
+Run: ``git add .``.
```
ArchdukeTim: This only adds every file if you’re in the root directory. It’s also a terrible practice because people will accidentally add files they don’t mean to

source/docs/software/basic-programming/git-getting-started.rst
```diff
@@ -0,0 +1,250 @@
...
+2. Confirm it was added via: ``git remote -v``
+3. Pull changes from upstream: ``git fetch upstream``
+4. Merge the changes into head: ``git merge upstream/upstream-branch-name``
+
+Gitignore
+---------
+
+.. important:: It is extremely important that teams **do not** modify the .gitignore file that is included with their robot project. This can lead to offline deployment not working.
+
+A .gitignore file is commonly used as a list of files to not automatically commit with ``git add .``. Any files or directory listed in this file will **not** be commited.
```
ArchdukeTim: Ignored files will not show up with gitstatus either

source/docs/software/basic-programming/git-getting-started.rst
```diff
@@ -0,0 +1,250 @@
...
+4. Merge the changes into head: ``git merge upstream/upstream-branch-name``
+
+Gitignore
+---------
+
+.. important:: It is extremely important that teams **do not** modify the .gitignore file that is included with their robot project. This can lead to offline deployment not working.
+
+A .gitignore file is commonly used as a list of files to not automatically commit with ``git add .``. Any files or directory listed in this file will **not** be commited. 
+
+Hiding a Folder
```
ArchdukeTim: I recommend linking https://www.atlassian.com/git/tutorials/saving-changes/gitignore

It explains gitignore pattern matching well

source/docs/software/basic-programming/git-getting-started.rst
```diff
@@ -0,0 +1,250 @@
...
+
+Fill out the appropriate information, and then click *"Create repository"*
+
+.. image:: images/image5.png
+
+You should see a screen similar to this
+
+.. image:: images/image6.png
+
+Now you'll want to open a powershell window and navigate to your project directory. An excellent tutorial can be found `here <https://programminghistorian.org/en/lessons/intro-to-powershell>`__
```
ArchdukeTim: What if I’m not on windows?

ThadHouse:
We do have some git things. Like the default templates are setup for git, and its built into vscode. We even debated including it in the installer at one point, although that might be too far. But I think enough people use git or should use git where we should have some documentation on it.

ThadHouse:
On that note, one thing this should say in the gitignore section is the VS Code templates include a properly set up .gitignore file, so if using the VS Code generated templates there is no no need to go find your own.

Daltz333:
> 
> 
> We should consider if we want to take on supporting tools that we do not directly support (support meaning here write code for).

I second Thad's response, we should have at least minimal documentation how to use git.

ArchdukeTim:
What about having a git document that has links to other already written tutorials that can go much more in depth? No need to reinvent the wheel, right?

source/docs/software/basic-programming/git-getting-started.rst
ThadHouse: This would be one where I would likely stray towards being more useful then pedantically correct. All good git repos need a gitignore.

source/docs/software/basic-programming/git-getting-started.rst
ThadHouse: We should focus our instructions on windows, as that is by far our most used platform, and most linux/mac users know the terminal. We can have a bit of guide on the other terminals, but we should focus mainly on windows.

Also, we should in here mention `ctrl + ~` in VS Code opens an integrated terminal. Since most teams are VS Code, I don't mind having a bit of platform specifics in there, and its super useful.

source/docs/software/basic-programming/git-getting-started.rst
Daltz333: I'll replace 
``Git repositories consist of:``
with
``Git repositories usually consist of:``

source/docs/software/basic-programming/git-getting-started.rst
Daltz333: Mac does not come with git, and the official installer is preferred in this case.

source/docs/software/basic-programming/git-getting-started.rst
Daltz333: I added a clarification.

source/docs/software/basic-programming/git-getting-started.rst
```diff
@@ -0,0 +1,260 @@
...
+- `Windows <https://git-scm.com/download/win>`_
+- `Mac OS X <https://git-scm.com/download/mac>`_
+- `Linux <https://git-scm.com/download/linux>`_
+
+.. note:: You may need to add Git to your `path <https://www.google.com/search?q=adding+git+to+path>`__
+
+Git Vocabulary
+--------------
+
+Git is composed of several primary elements:
```
Oblarg: I would say, instead, "Git revolves around several core commands."

source/docs/software/basic-programming/git-getting-started.rst
```diff
@@ -0,0 +1,260 @@
...
+
+- **Repository:** the data structure of your code, including a ``.git`` folder in the root directory
+- **Commits:** a particular saved state of the repository, this includes all files and additions
+- **Branches:** a means of separating various commits, having a unique history. This is primarily used for separating development and stable branches.
+- **Push:** update the remote repository with your local changes
+- **Pull:** update your local repository with the remote changes
+- **Cloning:** retrieving a local copy of a repository to modify
```
Oblarg: Inconsistent parallel structure; some of these are nouns, some of them are the actual command name, some of them are gerunds.  Make this consistent.

source/docs/software/basic-programming/git-getting-started.rst
```diff
@@ -0,0 +1,260 @@
...
+- `Linux <https://git-scm.com/download/linux>`_
+
+.. note:: You may need to add Git to your `path <https://www.google.com/search?q=adding+git+to+path>`__
+
+Git Vocabulary
+--------------
+
+Git is composed of several primary elements:
+
+- **Repository:** the data structure of your code, including a ``.git`` folder in the root directory
```
Oblarg: ""Repository" and "branches" should go in an earlier section that explains the core notion, probably.

auscompgeek:
Parts of this seem like a "how to use GitHub", which I don't think is appropriate for a how-to on git. Version control by itself, without any remote hosts, is equally valuable for keeping backups of previous versions of code if you're a team of one programmer.

Daltz333:
I think the cloud portion of git is as equally important, if not more so, than simply knowing how to use it.

auscompgeek:
Sure I guess, but then whatever we write shouldn't make it seem like we're pushing a particular service as _the_ (single) hosting service when there are others.

Daltz333:
There are indeed, but having tutorials introducing every other alternative would be excessive. I could include names of alternatives, but thats about the extent it could go. In my experience, Github is the most popular one.

Daltz333:
Closing, as I'll create a new PR with a different approach.

Daltz333:
Reopened per Thad's request. Reviewing and cleaning up content and then will merge.

source/docs/software/basic-programming/git-getting-started.rst
```diff
@@ -22,16 +24,16 @@ You have to download and install Git from the following links:
...
-- **Commits:** a particular saved state of the repository, this includes all files and additions
-- **Branches:** a means of separating various commits, having a unique history. This is primarily used for separating development and stable branches.
+- **Commit:** a particular saved state of the repository, this includes all files and additions
+- **Branch:** a means of separating various commits, having a unique history. This is primarily used for separating development and stable branches.
 - **Push:** update the remote repository with your local changes
 - **Pull:** update your local repository with the remote changes
-- **Cloning:** retrieving a local copy of a repository to modify
-- **Forking:** duplicating a pre-existing repository to modify, and to compare against the original
-- **Merging:** combining various changes from different branches/commits/forks into a single history
+- **Clone:** retrieving a local copy of a repository to modify
```
ArchdukeTim: Change these to define the noun, not the verb

Daltz333: Are you looking at the correct history? They've been changed to their noun form.

ArchdukeTim: The word "clone" has, but the definitions define different parts of speech. For commit, you define what a commit is, rather that what committing is, but for clone, you define what cloning is, rather than what a clone is

PR #66, opened 2019-05-26
Accept new articles

Daltz333:
Most of the screensteps have been ported. New articles are OK to be accepted, but revisions should be heavily preferred.

### DIFF
--- a/source/contributing.rst
+++ b/source/contributing.rst
@@ -3,7 +3,7 @@
 Contributing to frc-docs
 ========================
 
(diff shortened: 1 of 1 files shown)


ThadHouse:
This should state that any new content needs to be first approved before it can be accepted. Things like new sensors and other new info docs need to be approved first. Porting and corrections can be accepted without issue.

Daltz333:
@ThadHouse. What would this approval process look like?

ThadHouse:
Likely something on the PRs, but for the contributing doc likely say something like not everything can be accepted, and we will post a reason if not accepted, but no hard feelings if things get closed.

Sensors really are the big ones. Anything on the official docs needs to be on offically supported sensors.

Daltz333:
@ThadHouse Is the above commit more appropriate?

PR #68, opened 2019-05-28
Azure Cleanups

Daltz333:
Just some small visual cleanups. Also removed the archive task since it didn't really do anything.


ThadHouse:
Please keep the archive files. You can use it to download what gets built to verify things work correctly.

Issue #61, opened 2019-05-21
Reorganize TOCs


Daltz333:
Current structure breaks topics between software, hardware, and networking. See PDF output. 

Figure out a new nesting scheme for nested TOCs

PR #71, opened 2019-06-05
TOC Document Restructure

Daltz333:
Fixes some navigation problems with the sidebar. 

This

Becomes 

As well as an index page for each sub-category.

This *will* be the last and final major change to the index and categorization system to frc-docs for the foreseeable future. A separate pull request will be made limiting the depth of local tocs as well as their structure.  

Closes https://github.com/wpilibsuite/frc-docs/issues/61

This PR has also been updated to also include
- Changing the copyright to "2019 FIRST"
- Changing the author to "WPILib"
- Titles only option enabled
- renamed to FIRST Robotics Competition as Sphinx automatically appends "documentation"
- collapse navigation set to true (possibly help users to prevent cluttering their sidebar)


Kevin-OConnor:
The home page still looks pretty overwhelming here. Thoughts on limiting maxdepth to 1 for everything on the main index page? I'd also like to consider the titles_only theme option in conf.py to prevent users from accidentally making their sidebar a mess (could be considered separately but felt related)

Daltz333:
I'm fine with limiting it to titles only. Maxdepth of 1 does make it harder to Ctrl + F.

Daltz333:
This pull request was more for reorganizing the TOCs so it didn't break in the sidebar. PDF navigation is slightly broken (hardware is under software), but there's not a feasible way to fix that as RTD.

Daltz333:
Not against level 1 depth, just want to see how it flows first. Unable to make any changes for a few hours, as I'm out of town.

Daltz333:
> 
> 
> The home page still looks pretty overwhelming here. Thoughts on limiting maxdepth to 1 for everything on the main index page? I'd also like to consider the titles_only theme option in conf.py to prevent users from accidentally making their sidebar a mess (could be considered separately but felt related)

I've enabled the titles only and the sticky navigation options. After some tinkering with maxdepth of 1, I stand by keeping it at level 2 for the software and hardware articles. Limiting it to 1 removes **a lot** of information

PR #77, opened 2019-06-06
Use the readthedocs docker image to build

ThadHouse:


ThadHouse:
Their image is too big and specialized, so it is not feasible to use it directly.

PR #87, opened 2019-06-13
Various configuration tweaks

Daltz333:
- Fix readthedocs build failing (sphinx_tabs was incompatible with pdf generation)
- Generate PDF and epub files on azure pipelines

Modifications
--------------
- Suppress unknown mimetype for EPUB generation. Image files generated with graphviz do not have a known mimetype and should be suppressed. Does **not** effect the final result.
- Adds epub to the list of valid builders for the sphinx_tabs plugin. This results in code output looking like the below
- This also adds extra dependencies to build PDF and EPUB for appveyor
- Also bumps Sphinx tabs version up 1.1.11 which fixes compatibility with SphinxV2

Example EPUB/PDF tab code content (unfortunately, I do not believe this is configurable):

Java
C++

```
java content here
```

```
cpp content here
```

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -91,14 +91,22 @@ def setup(app):
 
 # -- Options for latex generation --------------------------------------------
 
-latex_engine = 'pdflatex'
+latex_engine = 'xelatex'
 
 latex_elements = {
-	'preamble': r'''
-	\usepackage[utf8]{inputenc}
-	\usepackage{fdsymbol}
-	\usepackage{newunicodechar}
-	\newunicodechar{⌀}{\ensuremath{\diameter}}'''
+    'fontpkg': r'''
+	\setmainfont{DejaVu Serif}
+	\setsansfont{DejaVu Sans}
+	\setmonofont{DejaVu Sans Mono}''',
+    'preamble': r'''
+	\usepackage[titles]{tocloft}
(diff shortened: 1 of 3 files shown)


source/conf.py
```diff
@@ -20,6 +20,7 @@
 project = 'FIRST Robotics Competition'
 copyright = '2019, FIRST'
 author = 'WPILib'
+version = '2019.4.1'
```
auscompgeek: Do we actually want these docs to track WPILib releases? I'd imagine the only thing that'd change between season updates is any release notes.

Kevin-OConnor: Probably not. It feels like one more thing we have to remember and coordinate while working on releases for not much win.

Daltz333: EPUB generation requires a version be set. An alternative could be years. 2019, 2020, etc. Thoughts?

Kevin-OConnor: Hmmm...yeah, just year feels more representative than WPILib version which I don't think we will keep up to date. But we could also just fill in a garbage version (like 1.0 or something).

Daltz333:
We'd have to update the copyright year anyways. Want to use a filler version anyways?

Kevin-OConnor:
> 
> 
> We'd have to update the copyright year anyways. Want to use a filler version anyways?

No, year seems fine.

Daltz333:
Ready for merge @Kevin-OConnor

PR #108, opened 2019-07-03
Remove Background Gradient

Daltz333:
Due to popular request, removes the background gradient to focus on content, not have the background distract the reader.

### DIFF
--- a/source/_static/css/frc-rtd.css
+++ b/source/_static/css/frc-rtd.css
@@ -52,11 +52,6 @@ margin-left:300px;
 background:#fcfcfc;
 }
 
-.wy-nav-content-wrap
-{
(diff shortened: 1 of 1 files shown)


AustinShalit:
I do not like the default grey.  Can we leave it as a color?  Maybe the light blue that was the start of the gradient?


Screen Shot 2019-07-03 at 3 35 11 PM
Screen Shot 2019-07-03 at 3 35 16 PM

AustinShalit:
Maybe it would look better if we used the grey somewhere else too, like under the logo area.

AustinShalit:
Actually, maybe we should hold off on doing this because [frc-events](https://frc-events.firstinspires.org) uses a similar gradient. I think we want to have consistent styling.

calcmogul:
Consistently bad style is still bad style. FIRST's own website doesn't even use this gradient; the colors are closer together chromatically on their site.

AustinShalit:
The gradient is exactly the same as the one on the main site.

calcmogul:
It's the same pattern, but it stretches across the whole page instead of just to the content. The brighter segments of it are covered by the page content whereas in rtd the gradient only stretches to the content, so the whole gammot of colors are shown. FIRST's website gradient looks better than RTD's because the brighter colors are hidden.

Kevin-OConnor:
I would be fine with changing the gradient range so it looks more like what's on the FRC home page, but not getting rid of it.

Daltz333:
I'm not inclined towards further modifying the design, this PR was only opened for public appeal anyways.

PR #109, opened 2019-07-03
Updated Victor SP image

Bankst:


Daltz333:
On second thought, could this image be resized down to the size of the original image? That's a big jump from 128KB to 2MB

PR #123, opened 2019-07-08
Add notice about Windows Lint Checking

Daltz333:
Lint Checking is unavailable due to https://bugs.launchpad.net/doc8/+bug/1756704

This PR adds a small notice in the build instructions saying it's unavailable.


Daltz333:
Adding onto this, I'll open another pull request actually adding linting to the make.bat file whenever it gets fixed. Not much point ATM considering it can't even be used.

Issue #88, opened 2019-06-14
Update azure build status in readme


ThadHouse:
https://dev.azure.com/wpilib/Documentation/_apis/build/status/wpilibsuite.frc-docs?branchName=master

Kevin-OConnor:
Should update the readme azure badge to point at this, but might as well wait until merge of develop into master

PR #145, opened 2019-07-19
Update Readme

### LINKED ISSUE #137: Add clarification that the FIRST logo is not Creative Commons
FIRST logo is not creative commons, and that should be specified somewhere

Daltz333:
Open to conversation on anything else that needs to be added or changed. 
Closes #137 #88


Daltz333:
Any reason to keep it in the readme when there is this article? https://frc-docs.readthedocs.io/en/latest/docs/contributing/build-instructions.html

AustinShalit:
Oh.  I need to think about this a bit more.  I think we may want to remove those pages of the documentation and instead have all of that information in the readme or other files that are not put onto the site itself.

Daltz333:
Could always have both?

PR #159, opened 2019-08-05
Move and Cleanup roboRIO imaging guide

Daltz333:
- Fix formatting
- Follows style guide
- Move article and images underneath the roboRIO section


source/docs/software/roborio-info/imaging-your-roborio.rst
```diff
@@ -1,60 +1,50 @@
...
 .. image:: images/imaging-your-roborio/usb-connection.png
 
-Connect a USB cable from the roboRIO USB Device port to the PC. This requires
-a USB Type A male (standard PC end) to Type B male cable (square with 2 cut
-corners), most commonly found as a printer USB cable.
+Connect a USB cable from the roboRIO USB Device port to the PC. This requires a USB Type A male (standard PC end) to Type B male cable (square with 2 cut corners), most commonly found as a printer USB cable.
 
-.. note:: The roboRIO should only be imaged via the USB connection. It is not\
- recommended to attempt imaging using the Ethernet connection.
+.. note:: The roboRIO should only be imaged via the USB connection. It is not\ recommended to attempt imaging using the Ethernet connection.
```
AustinShalit:
```suggestion
.. note:: The roboRIO should only be imaged via the USB connection. It is not recommended to attempt imaging using the Ethernet connection.
```

source/docs/software/getting-started/index.rst
```diff
@@ -12,7 +12,6 @@ Getting Started
    macos-setup
    linux-setup
    3rd-party-libraries
-   imaging-your-roborio
```
AustinShalit: I think we need this here.  This is an important getting started step.

source/docs/software/roborio-info/index.rst
```diff
@@ -4,6 +4,7 @@ roboRIO
 .. toctree::
    :maxdepth: 1
 
+   imaging-your-roborio
```
AustinShalit: See comment above.  Leave in getting started.

source/docs/software/getting-started/index.rst
Daltz333: The Getting Started section should really be renamed to WPILib getting started, as it is much more pertinent to WPILib setup. This enforces the per topic structure.

source/docs/software/roborio-info/imaging-your-roborio.rst
```diff
@@ -90,7 +90,7 @@ Troubleshooting
 If you are unable to image your roboRIO, troubleshooting steps include:
 
 - Try running the roboRIO Imaging Tool as Administrator by right-clicking on the Desktop icon to launch it.
-- Try accessing the roboRIO webpage with a web-browser at ``http://172.22.11.2/`` and/or verify that the NI network adapter appears in your list of Network Adapters in the Control Panel. If not, try re-installing the NI Update Suite or try a different PC.
+- Try accessing the roboRIO webpage with a web-browser at ``https://172.22.11.2/`` and/or verify that the NI network adapter appears in your list of Network Adapters in the Control Panel. If not, try re-installing the NI Update Suite or try a different PC.
```
AustinShalit: This actually will not work because the rio does not have a certificate. http is correct.

source/docs/software/getting-started/index.rst
sciencewhiz: This section had everything that programmers need to do at a Robot Quick Build session. If you move imaging out, it breaks that. 

If you wanted it to be just WPILib getting started, there's other things that should be moved out as well, like radio programing and updating the PCM and PDP.

Daltz333: Yes, those should be moved out as well. The idea was to create an actual getting started article directing users on systems. Even on ScreenSteps, the getting started article was never a "1 to boom robot".

sciencewhiz: What was missing?

Daltz333: This was discussed in a WPILib meeting awhile ago on the reasoning of the per-topic structure. Getting-Started may have been a "how-to" guide for rookie teams but caused numerous other problems. Every single article in getting-started was duplicated in various parts of the documentation (not linked to a single page but duplicated so inaccuracies between them). There was also the issue of coming to ScreenSteps looking for specific things and being unable to find them due to the poor organization and the "how-to" structure. If you come to frc-docs wanting to know how to use shuffleboard, you shouldn't have to dig through 50+ pages to find a single article. Per-topic alleviates most of the problems with very little cost. The trade-off that was decided was to create an actual introduction article to the FRC environment *without* cluttering up the docs. 

The reasons outlined above are why this pull-request has not been merged yet. I simply haven't had the time to create the article with the content referencing things in the various places.

sciencewhiz: I'm ok with that, but don't think this should move until the other parts are moved in a separate PR.

PR #174, opened 2019-08-24
Add intro book amazon links/ISBN

Daltz333:
Adds a generic amazon link to the books, as well as gives them their ISBN numbers

image


calcmogul:
Should we explicitly mention the books can be obtained from a local library or something? Idk if students know libraries are a thing nowadays.

PR #175, opened 2019-08-24
Add cpp example to using-speed-controllers

mcm001:
It was previously a TODO which was left in and merged and went unnoticed.


mcm001:
Is the formatting from #235ca14 what you're looking for?

PR #194, opened 2019-09-14
Exclude kauailabs from linkcheck

Daltz333:
KauaiLabs website consistently has downtime, and various website issues. This excludes all urls with their root domain from the link check. In the mean time, I have sent them an email informing them of these issues.

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -48,7 +48,7 @@ todo_include_todos = False
 linkcheck_anchors = False
 
 # Linkcheck Exclusions
-linkcheck_ignore = ['.*kauailabs.com.*']
+linkcheck_ignore = [r'.*kauailabs.com.*']
 
 # Autosection labels prefix document path and filename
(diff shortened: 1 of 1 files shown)


AustinShalit:
Is this good now?

Daltz333:
Yep.

source/conf.py
```diff
@@ -47,6 +47,9 @@
 # Disable following anchors in URLS for linkcheck
 linkcheck_anchors = False
 
+# Linkcheck Exclusions
+linkcheck_ignore = [r'.*kauailabs.com.*']
```
AustinShalit: This will ignore anything with `kauailabs.com` in the address. Is that what we want to do or do we want to be more selective?

sciencewhiz: That's probably more open then it needs to be, but I think it's also not going to cause a problem. It's highly unlikely that someone else makes a domain name that contains kauailabs.com that we would reference in the docs.

PR #202, opened 2019-09-23
Add translation instructions

moesoha:
Add some instructions on localization, and the Chinese translation of `index` as an example.

And here is a team called "F&D Studio`. In the past years, we've translated some content on the "ScreenSteps" into Chinese. And we've started working on the Chinese translation of this doc today, more pages are coming.


Kevin-OConnor:
Hi @moesoha! It's great to see your enthusiasm on this! For the 2020 season we are concerned about adding the complexity of adding translations along with the whole site being new. I expect there to be considerable churn in content still over the next few months as we continue to get the new site settled and am concerned about keeping translation up to date.

Translation is one of the major features we were excited about with moving to RTD, but are currently thinking that our target would be 2021, not 2020.

moesoha:
OK, that's fine. Just keep this pr open, until the doc is stable.
And in fact, you needn't worry about outdated translation. This is based on `gettext`, so that when the original text changed, it'll create a new entry for it. The new content will stay in English if no new translations added corresponding to the new text.

Daltz333:
@moesoha are you still interested in continuing with this?

Now that there has been some time to mature, I'd look to coordinate moving forward with providing a translation framework.

moesoha:
Of course, then what's next?

AustinShalit:
@moesoha We are looking into using Transifex to translate the project that way folks do not need to learn git in order to contribute translations. Do you mind taking a look at the test project we have created here: https://transifex.com/wpilib/frc-docs? There should be a help translate button which will let you sign up.

I wouldn't start translating everything but if you don't mind looking at the tools they have and providing some feedback to us that would be amazing.

moesoha:
@AustinShalit I've sent the request to join the team, waiting for your approval

AustinShalit:
Approved. Take a look around and let me know what you think. As I mentioned above, I wouldn't start translating everything because we are still working out the technical details.

Daltz333:
I'm going to create a python script to replace the tx config command, as it's bugged for windows slug generation.

PR #206, opened 2019-09-26
Disable linkcheck

Daltz333:
Build keeps failing due to azure timing out/exceeding maximum number of retries. Linkcheck can just be done manually every now and again.


Daltz333:
#207 supercedes this

PR #207, opened 2019-09-26
Move LinkCheck into its own build job

AustinShalit:
Supersedes #206 

This way we can tell if only broken links are causing the build to fail.


Daltz333:
Can we also split up linting into it's own build? That way the actual ``make html`` will run. There is no need to split up HTML/PDF/EPUB builds. If one fails, it'd fail them all.

PR #208, opened 2019-09-27
Combine install docs into 1 tabbed doc

Kevin-OConnor:
Combine install docs to reduce article count/confusion in getting started workflow. Does mean a small sacrifice of section headers as they are not supported inside tabs but seems worth it.


source/docs/software/getting-started/wpilib-setup.rst
```diff
@@ -0,0 +1,208 @@
...
+      .. figure:: images/mac/UntarredRelease.png
+         :alt:
+
+      Using Finder (or command line) copy the contents of the folder to a new
+      folder in your home directory, ~/frc2019 as shown below.
+
+      .. figure:: images/mac/MovedFiles.png
+         :alt:
+
+    **Run the ``ToolsUpdater.py`` script**
```
AustinShalit: This does not render the ToolsUpdater.py properly.

Kevin-OConnor: Good catch.

PR #213, opened 2019-09-29
Structure Path Following section

Daltz333:
Separates the distinction between PathFinder and the newer path following classes.


source/docs/software/wpilib-tools/path-planning/index.rst
```diff
@@ -1,13 +1,9 @@
-Autonomous Path Planning
-========================
+Path Following
+==============
+
+There are two main implementions of path-followering/trajectory generation. Pathfinder and the new WPILib trajectory classes.
```
AustinShalit: We should provide some guidance here about what to use.

Daltz333:
Thats a controversial opinion right now :sweat_smile:

calcmogul:
WPILib traj gen is technologically better than PFv1 in every way. It's easier to use, supports constraints, and doesn't blow up on waypoints 90 degrees to each other. Idk what's so controversial.

source/docs/software/wpilib-tools/path-planning/pathfinder/index.rst
```diff
@@ -0,0 +1,13 @@
+Autonomous Path Planning
```
AustinShalit: Maybe Pathfinder should be in the title?

PR #223, opened 2019-10-04
Fixed a typo of "notrecommended" on "Imaging your roboRIO" page.

Javaru:
In the connection note, "It is not recommended to..." was rendered as "It is notrecommended to..." (i.e. missing a space after the word 'not') due to a "keep line with next" slash immediately after the word 'not' in the source.


Daltz333:
The better solution would be to update this page to follow the style guide. Anyways, I'm pretty sure another PR fixes this. 

Sorry bout that, thanks anyways.

Daltz333:
Fixed in #159

PR #234, opened 2019-10-06
Remove FRC CAN Device Specifications document

yellowjaguar5:
It appears that the FRC CAN Device Specifications document got stranded in the `hardware` directory, but its accompanying article now lives in `software/can-devices`. This PR links the document with its related article. 

Also, can the `frc-can-device-specifications` document just be deleted? The article with the same name seems to be a copy of it.


Daltz333:
Honestly the PDF can just get deleted IMO, however this is more up to FIRST.

Kevin-OConnor:
If the article has all of the info from the doc then the doc can go. It was posted to screensteps when we didnt have time to port the content from the word doc for something used so infrequently

AustinShalit:
Do we need to remove the link too?

yellowjaguar5:
I took the link out of the article because I removed the document if that's what you meant.

PR #244, opened 2019-10-07
Update style guide to include admonitions and lists

### LINKED ISSUE #243: Add to styleguide
Styleguide should be updated to be specific about how to format role directives such as "note" and "warning" as our formatting does not match Sphinx examples.

Should also specify desired indent (or in this case not) for lists.

Daltz333:
Closes #243


AustinShalit:
Do we need a note about lists?

PR #247, opened 2019-10-08
Collapse contributing and issues for clarity.

sciencewhiz:
Clarify that this is Control System documentation, as opposed to
all the other documentation at
https://www.firstinspires.org/resource-library/frc/competition-manual-qa-system

### DIFF
--- a/source/index.rst
+++ b/source/index.rst
@@ -3,10 +3,10 @@
    You can adapt this file completely to your liking, but it should at least
    contain the root `toctree` directive.
 
-FIRST Robotics Competition Documentation
-========================================
+FIRST Robotics Competition Control System Documentation
+=======================================================
 
-Welcome to the FIRST Robotics Competition Documentation! This documentation will be the Official FIRST® Robotics Competition documentation for the 2020 season and beyond.
+Welcome to the FIRST Robotics Competition Control System Documentation! This documentation will be the Official FIRST® Robotics Competition control system documentation for the 2020 season and beyond.
 
 .. note:: It's advised to continue using `ScreenSteps <https://wpilib.screenstepslive.com/s/4485>`__ until the official WPILib release for 2020.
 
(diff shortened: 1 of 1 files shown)


Daltz333:
Note that the conf.py may need to be updated as well. I personally think this is lengthy, and should be obvious.

source/index.rst
```diff
@@ -79,17 +79,13 @@ Welcome to the FIRST Robotics Competition Documentation! This documentation will
...
 .. toctree::
    :maxdepth: 1
-   :caption: Contributing
+   :caption: Contributing to Documentation
 
    docs/contributing/contribution-guidelines
    docs/contributing/style-guide
    docs/contributing/build-instructions
    docs/contributing/top-contributors
+   Report a documentation issue <https://github.com/wpilibsuite/frc-docs/issues>
```
Daltz333: I don't like how less obvious this becomes.

Daltz333: This is be kept in it's own toctree to emphasis issue distinction, and the URL can be changed to point to the allwpilib repo, where we can move issues around as needed.

Kevin-OConnor:
> 
> 
> Note that the conf.py may need to be updated as well. I personally think this is lengthy, and should be obvious.

What would need to be updated in conf.py? I don't see anything

Daltz333:
> 
> 
> > Note that the conf.py may need to be updated as well. I personally think this is lengthy, and should be obvious.
> 
> What would need to be updated in conf.py? I don't see anything

project_name may need to be updated, not that it automatically appends "documentation" to the end. This is also what I believe determines the browser meta name.

Kevin-OConnor:
> project_name may need to be updated, not that it automatically appends "documentation" to the end. This is also what I believe determines the browser meta name.

It appears to be used to append to the name of the page in the ``<title>`` tag. So if we update the name of the index page as I have proposed in #249, the title that shows in a browser tab is "FIRST Robotics Competition Control System - FIRST Robotics Competition documentation." I don't have a problem changing it to something else, but nothing after "Competition" shows in Firefox or Chrome anyway

Daltz333:
I actually kinda like that, it's not redundant.

PR #249, opened 2019-10-08
Fix FIRST branding in docs

Kevin-OConnor:
Updates _FIRST_ branding to italicize all instances of _FIRST_ and add the registered mark after the first instance of _FIRST_ on the page. Also added registered mark to instances of FRC on pages being touched anyway, but did not do a complete update for FRC.


sciencewhiz:
We should probably reference the FIRST branding standards in the styleguide.

PR #251, opened 2019-10-08
Small graph fix

Oblarg:
Yes, I should have looked at this more carefully before merging it.


Kevin-OConnor:
There's a hanging paren in the (non-note) paragraph below the Scheduler run graph, any chance you want to fix that too real quick?

Oblarg:
Yeah, I'll do that when I get home

PR #221, opened 2019-10-03
Recreates diagrams in GraphViz

Modelmat:
Fixes #220 

Thoughts on:
- ~~SVG output rather than PNG?~~ Browser support & labview metadata
- Using transparent (not white) backgrounds

Checklist:
- [x] [Types of Sensors](https://frc-docs.readthedocs.io/en/latest/docs/software/sensors/sensor-overview-software.html)
	The graph does not look exactly the same but it's as close as I could make it.
image
- [ ] [Vision Workflows](https://frc-docs.readthedocs.io/en/latest/docs/software/vision-processing/introduction/strategies-for-vision-programming.html)
    GraphViz doesn't seem to render this very nicely - https://gist.github.com/Modelmat/f9d9aa40af391e6a6c6b09c7bd4d3523
- [x] [Sources and Sinks](https://frc-docs.readthedocs.io/en/latest/docs/software/vision-processing/introduction/cameraserver-class.html)
	Done before this PR
- [x] [Subsystems and Commands](https://frc-docs.readthedocs.io/en/latest/docs/software/commandbased/what-is-command-based.html#id2)
image
- [x] [Scheduler Control Flow Diagram](https://frc-docs.readthedocs.io/en/latest/docs/software/commandbased/what-is-command-based.html#id3)
	This requires a complete remake (inaccurate). ~~Waiting on #191~~ ~~Done through #191~~ ~~Slight issues with existing; the end should instead draw an arrow back up to the start.~~ Complete
- [x] [Command Group with Concurrency](https://frc-docs.readthedocs.io/en/latest/docs/software/commandbased/command-groups.html#id4)
image
- [x] [NAT](https://frc-docs.readthedocs.io/en/latest/docs/networking/networking-introduction/networking-basics.html#public-vs-private-ip-addresses)
image

PR #260, opened 2019-10-11
Makes the "Scheduler Run Sequence" diagram flow correctly.

Modelmat:
The new graph looks like:
image

I was unable to shift the "False" label away from on top of the line.


Modelmat:
Oops, this was meant to go as part of #221

PR #261, opened 2019-10-11
Bump year to 2020

Daltz333:


sciencewhiz:
IANAL, but I don't think the copyright date should be updated before 2020. The other one is fine.

PR #262, opened 2019-10-11
Add links to beta API docs

Daltz333:


AustinShalit:
Beta is over so I think we can close this.

PR #272, opened 2019-10-16
Fix year in installing Update Suite doc

sciencewhiz:


Modelmat:
Should we not have ahead at all?
This would mean we would have to update it every year (a hassle)

source/docs/getting-started/getting-started-frc-control-system/frc-update-suite.rst
```diff
@@ -157,7 +157,7 @@ NI Activation Wizard (2)
 |NI Activation Wizard 2|
 
 Enter the serial number. Click “Activate”. Note: If this is the first time
-activating the 2019 software on this account, you will see the message
+activating the 2020 software on this account, you will see the message
```
AustinShalit: See comment about making this year agnostic

Issue #227, opened 2019-10-05
Duplicated Content and Images


Modelmat:
The below pages contain duplicate images and content. There may be other pages that contain duplicate content.

- [Adding Widgets to SmartDashboard](http://docs.wpilib.org/en/latest/docs/software/wpilib-tools/smartdashboard/smartdashboard-intro.html#adding-widgets-to-the-smartdashboard) contains the same image as at the top of the page, [Introduction](http://docs.wpilib.org/en/latest/docs/software/wpilib-tools/smartdashboard/smartdashboard-intro.html#smartdashboard-introduction)
   + The images are [smartdashboard-intro-1](https://github.com/wpilibsuite/frc-docs/blob/master/source/docs/software/wpilib-tools/smartdashboard/images/smartdashboard-intro-1.png) and [smartdashboard-intro-6](https://github.com/wpilibsuite/frc-docs/blob/master/source/docs/software/wpilib-tools/smartdashboard/images/smartdashboard-intro-6.png)
- [Victor SP](https://docs.wpilib.org/en/latest/docs/getting-started/getting-started-frc-control-system/control-system-hardware.html#victor-sp) and [Victor SPX](https://docs.wpilib.org/en/latest/docs/getting-started/getting-started-frc-control-system/control-system-hardware.html#victor-spx)
  + [victor-sp-motor-controller.png](https://github.com/wpilibsuite/frc-docs/blob/master/source/docs/getting-started/getting-started-frc-control-system/images/control-system-hardware/victor-sp-motor-controller.png) and [victor-spx-motor-controller.png](https://github.com/wpilibsuite/frc-docs/blob/master/source/docs/getting-started/getting-started-frc-control-system/images/control-system-hardware/victor-spx-motor-controller.png)
- [Camera Webpage](https://docs.wpilib.org/en/latest/docs/software/vision-processing/introduction/configuring-an-axis-camera.html#camera-webpage) and [Manual Camera Configuration](https://docs.wpilib.org/en/latest/docs/software/vision-processing/introduction/configuring-an-axis-camera.html#manual-camera-configuration)
[...]

PR #280, opened 2019-10-18
Remove duplicated images

katzuv:
#227 
I fixed all the duplicates shown in the issue.
However, I wanted to find all duplicated in the docs. So I used [this](https://www.pythoncentral.io/finding-duplicate-files-with-python/) Python script that finds duplicate files (not only images). It can get as arguments a folder or several folders.
Using [this](https://superuser.com/a/426004) Linux Terminal command I found all directories holding images in the `source` directory. The final command looked like this:
```bash
python3 dupFinder.py $(find . -path '*/images/*' -type d)
```
That showed there are actually more duplicates (8 of them to be exact).
Maybe this check should be added to the CI or something. Maybe even write a script to fix this automatically.

Anyway, here are my fixes.


Modelmat:
I believe those extra duplicates were in different sections of the docs. I was told that moving them out would require placing them in some top-level assets directory which apparently is unwieldy.
(As opposed in the local images/) directory.

katzuv:
> I believe those extra duplicates were in different sections of the docs.

Yes, you are right. Can you use relative imports in `.rst` files?

Modelmat:
They are relative, unless prefixed with /, which corresponds to Sphinx's basedir (iirc).

I don't think it would be a good idea to use image files from other sections' image folder - this seems like a bad idea.

katzuv:
Yeah, you're probably right about this.
So I think all the samd-folder duplicates were fixed, I'll double check it in the morning (it's already night here in Israel 🙃).

katzuv:
OK, so I checked again, and these are all the needed fixes. Can someone review this PR?

katzuv:
Thanks, @AustinShalit!

AustinShalit:
Yup. Just waiting for @Daltz333 to confirm that this looks ok.

PR #281, opened 2019-10-19
Use forward slashes instead of backslashes

Daltz333:
Fixes it from appearing as ``JavaC++``


AustinShalit:
There is also one more in the status light page.

PR #283, opened 2019-10-19
Add notes for windows install clarification

sciencewhiz:
Installation location change
Difference between all users and current user


sciencewhiz:
Redundant with what, the changelog? It's much better to have specific notes where the user will run into them rather then just hiding everything in the changelog. For example, see the pathweaver swapped paths issue.

Daltz333:
The second note block becomes redundant in a year, and offers little benefit when the first note block states that the installation is installed to ``wpilib/2020``. I don't see it any benefit over just having the first note block. Plus, users should rarely if ever have to interact with this directory at all.

sciencewhiz:
Yes, there is overlap between the notes in that they both list the path. However, the intent is the second note can be removed after a year, while the first note will stay forever. I changed the second note to apply to all OSes and moved it to the end.

PR #293, opened 2019-10-24
Add alternative document URLs to the README

Daltz333:
Sometimes schools will wildcard block websites (like my old school blocking anything with "networking" in the URL 🤢). This adds the list of domains to the README, and states that the WPILib domain is the preferred domain.


sciencewhiz:
Why is it prefered to use the wpilib.org domain? Does it matter at all?

Daltz333:
It doesn't necessarily, just to maintain consistency.

PR #294, opened 2019-10-25
Add scrollbar to the sidebar

### LINKED ISSUE #292: [artf5911] New 2020 Online Docs Site - Left Panel Sometimes Lacks Scrollbar, which is a Problem if on a Laptop with no "ScrollWheel"
>The New 2020 Online Docs site uses a website layout with a left pane providing an index to the documentation site.  At least in the situations I tried (Google Chrome on Windows 10 on multiple different computers), the left pane does not show a scrollbar unless the screen is zoomed out far enough to fit the entire pane on the screen (which, ironically, is the situation where the scrollbar isn't actually needed!)

>The attached screen snips help show the problem:

WPILib-Docs-no-scrollbar
WPILib-Docs-scrollbar

>The first screen snip shows the scrollbar for the left pane, when the window is zoomed out completely.

>The second screen snip shows the window at a normal zoom level, and the scrollbar for the left pane disappears!

>On a computer with a scrollwheel mouse, this isn't a show-stopper, as the scrollwheel can be used to scroll the pane.  However, on a laptop (which typically does not have a scrollwheel) the navigation panel becomes practically useless and very confusing, particularly since it auto-expands sections and re-navigates automatically when clicking on links in the right panel.

>This problem seems like a Web UI issue, rather than being due to the content of the pages, but hopefully one that can be fixed somehow...

Daltz333:
Closes #292 
- Refactors the CSS source
- Increases sidebar width by 20px to accommodate the scrollbar (and margins of appropriate elements to adjust for this)
- Adds a small 10px margin above the top of the navigation to account for the colored top bar

image

### DIFF
--- a/source/_static/css/frc-rtd.css
+++ b/source/_static/css/frc-rtd.css
@@ -84,4 +84,12 @@
 	.header-bar {
 		display: none;
 	}
+
+	.wy-nav-content-wrap {
+		margin-left: 0px;
+	}
+
+	.wy-nav-side {
+		width: 300px;
+	}
 }
--- a/source/_static/css/frc-rtd.css
+++ b/source/_static/css/frc-rtd.css
@@ -1,68 +1,87 @@
 #link-bar a:visited {
-color:#FFF;
+	color: #FFF;
 }
 
 #link-bar li {
-float:left;
-padding:15px;
+	float: left;
+	padding: 15px;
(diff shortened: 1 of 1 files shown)


AustinShalit:
This works on macOS. If someone could please test on Windows.

PR #295, opened 2019-10-25
Fix inaccuracies in benchtop-test-program-cpp-java

### LINKED ISSUE #290: [artf5909] In Docs for "Creating your Benchtop Test Program (C++/Java)" Incorrect Shortcut for "Open Project..."
>In the "Opening The New Project" (see https://frc-docs.readthedocs.io/en/latest/docs/getting-started/running-a-benchtop/creating-benchtop-test-program-cpp-java.html#opening-the-new-project) section of the "Creating your Benchtop Test Program (C++/Java)" online documentation, there is the following sentence:

>"We can choose to do that now or later by typing Ctrl-O (Command+O on macOS) and select the folder where We saved our project."

>Ctrl-O is not the correct keyboard shortcut to "open a project."  The correct shortcut to open a project (at least on Windows 10) is Ctrl-K, Ctrl-O.  Rather, Ctrl-O is a shortcut for "open a file" which is similar, but not quite the same.  Suggest fixing either the shortcut to be Ctrl-K, Ctrl-O (on Windows; I don't know about macOS), or changing the instructions to be "open a file" in which case a file will need to be specified, too (e.g.  "Robot.java")

### LINKED ISSUE #291: [artf5910] Online Docs for "Simple Autonomous Example" in "Creating your Benchtop Test Program (C++/Java)" Inconsistent Between Text Description and Java
>In the "Simple Autonomous Example" (seehttps://frc-docs.readthedocs.io/en/latest/docs/getting-started/running-a-benchtop/creating-benchtop-test-program-cpp-java.html#simple-autonomous-example) section of the "Creating your Benchtop Test Program (C++/Java)" online documentation, there is the following sentence:

>"The value is negative for forward motion because of the convention for joysticks where a negative Y-axis value corresponds to moving the stick away from you (forward)."

>The C++ code does use a call to "m_robotDrive.ArcadeDrive(-0.5, 0.0);" but the Java code instead calls "m_robotDrive.arcadeDrive(0.5, 0.0); // drive forwards half speed" which is not consistent with the text description.

>The inconsistency should be fixed.  (Not clear to me which way the change should be made...)

>I should also note that unless there is a difference somewhere in the underlying WPILib code for m_robotDrive.arcadeDrive between C++ and Java (which probably wouldn't be good), since different parameters are used in these to examples, this line of code (with one using +0.5 and the other using -0.5) will give different behaviors in C++ and Java -- one will result in the robot driving forward and one will result in the robot driving backwards.  That should probably be fixed, too...

simple-autonomous-example

Daltz333:
Closes #291 
Closes #290 

- Remove the comment about joystick conventions as this is irrelevant since this is a constant value into the ArcadeDrive class, joystick values are the part that is inverted and this doesn't use a Joystick.
- Fix negative to positive in code


Modelmat:
The keyboard shortcut is not Ctrl + K .

It is Ctrl + K followed by Ctrl + O.

Also consider using the `.. kbd`` directive here? (I mentioned this is previous unrelated issue but whether we should?)

source/docs/getting-started/running-a-benchtop/creating-benchtop-test-program-cpp-java.rst
```diff
@@ -39,7 +39,7 @@ Once all the above have been configured, click "Generate Project" and the robot
 Opening The New Project
 -----------------------
 
-After successfully creating your project, VS Code will give the option of opening the project as shown below. We can choose to do that now or later by typing Ctrl-O (Command+O on macOS) and select the folder where We saved our project.
+After successfully creating your project, VS Code will give the option of opening the project as shown below. We can choose to do that now or later by typing Ctrl-K (Command+O on macOS) and select the folder where we saved our project.
```
sciencewhiz: Should it also be Command+K on macOS?

AustinShalit: Nope. Mac is able to select either a file or directory with a single dialog.

source/docs/getting-started/running-a-benchtop/creating-benchtop-test-program-cpp-java.rst
```diff
@@ -39,7 +39,7 @@ Once all the above have been configured, click "Generate Project" and the robot
 Opening The New Project
 -----------------------
 
-After successfully creating your project, VS Code will give the option of opening the project as shown below. We can choose to do that now or later by typing Ctrl-O (Command+O on macOS) and select the folder where We saved our project.
+After successfully creating your project, VS Code will give the option of opening the project as shown below. We can choose to do that now or later by typing :kbd:`Ctrl-K` then :kbd:`Ctrl-O` (or just Command+O on macOS) and select the folder where we saved our project.
```
AustinShalit:
```suggestion
After successfully creating your project, VS Code will give the option of opening the project as shown below. We can choose to do that now or later by typing :kbd:`Ctrl-K` then :kbd:`Ctrl-O` (or just :kbd:`Command+O` on macOS) and select the folder where we saved our project.
```

PR #314, opened 2019-11-02
Fix compressor C++ code

sciencewhiz:
Also clarify that "more control" of compressor just means you can keep it from turning on


source/docs/software/actuators/pneumatics.rst
```diff
@@ -46,7 +46,7 @@ all be queried from the Compressor object.
     allows for integrated closed loop control of a compressor. Creating any
     instance of a Solenoid or Double Solenoid object will enable the
     Compressor control on the corresponding PCM. The Compressor object is
-    only needed if you want to have greater control over the compressor or
+    only needed if you want to the ability to turn off the compressor or
```
Daltz333:
```suggestion
    only needed if you want the ability to turn off the compressor or
```

PR #315, opened 2019-11-03
Re-enable RTD search

Daltz333:
RTD has "fixed" their search to remove duplicates. We'll need to periodically purge the RTD build environment (no effect on the end user) every now and again.


AustinShalit:
How often is periodically?

Daltz333:
When the search issues become noticeable.

sciencewhiz:
The issue before was when an article was moved, both the old and new location showed up in the search results.

PR #317, opened 2019-11-05
Implement control system basics and PID articles

Daltz333:
From @calcmogul book

Content concerns direct toward Tyler.

I'll fix structure, grammar, or rST-isms.

- [x] Implement Chapter One
- [x] Implement PID articles

### DIFF
--- a/source/docs/software/wpilib-overview/known-issues.rst
+++ b/source/docs/software/wpilib-overview/known-issues.rst
@@ -3,6 +3,11 @@ Known Issues
 
 This article details known issues (and workarounds) for FRC Control System Software.
 
+WPILib Projects created with 2020.1.1 have a project version of Beta2020-2
+--------------------------------------------------------------------------
+
+Any projects created have a project version of `Beta2020-2` set in wpilib_preferences.json. This is not an issue, everything will still work properly. This will be fixed and update your projects in the 2nd release.
+
 C++ Intellisense - Files Open on Launch Don't Work Properly
 -----------------------------------------------------------
 


source/docs/software/basic-programming/control-system-basics.rst
```diff
@@ -0,0 +1,53 @@
...
+
+How can we prove closed-loop controllers on an autonomous car, for example, will behave safely and meet the desired performance specifications in the presence of uncertainty? Control theory is an application of algebra and geometry used to analyze and predict the behavior of systems, make them respond how we want them to, and make them robust to distrubances and uncertainty.
+
+Controls engineering is, put simply, the engineering process applied to control theory. As such, it's more than just applied math. While control theory has some beautiful math behind it, controls engineering is an engineering discipline like any other that is filled with trade-offs. The solutions control theory gives should always be sanity checked and informed by our performance specifications.We don't need to be perfect; we just need to be good enough to meet our specifications.
```
prateekma: Need space after period after the word `specifications`.

source/docs/software/basic-programming/control-system-basics.rst
```diff
@@ -0,0 +1,53 @@
...
+.. todo:: insert figure 1.2 here
+
+Block Diagrams
+--------------
+
+When designing or analyzing a control system, it is useful to model it graphically. Block diagrams are used for this purpose. They can be manipulated and simplified systematically.
+
+.. todo:: insert figure 1.3 here
+
+The open-loop gain is the total gain from the sun node at the input (the circle) to the output branch. this would be the system's gain if the feedback loop was disconnected. The feedback gain is the total gain from the output back to the input sum node. A sum node's output is the sum of its inputs.
```
prateekma: Should be sum node, not sun node.

calcmogul:
The only other thing I'd say is important is an intro to what a motion profile is (the rest of the book's things are optional). Chapter 16 of the book is a bit rough around the edges, but the minimum viable product of that would be saying what a motion profile is and why we use it, what a trapezoid profile is, what a trapezoid profile looks like, and that we pretty much always use trapezoid profiles in FRC instead of S-curve profiles because the latter provides no mathematical benefit (first order motor models can follow a trapezoid profile perfectly with finite inputs, but that explanation is optional).

PR #320, opened 2019-11-06
Add CAN Mfg ID for Studica

Kevin-OConnor:


ThadHouse:
@Kevin-OConnor should I add this to wpilib too?

Kevin-OConnor:
> 
> 
> @Kevin-OConnor should I add this to wpilib too?

Sure. I think the current intended use is non-FRC with VMX-pi so I don't know if they plan to use WPIlib for it or not, but it definitely doesn't hurt.

Issue #297, opened 2019-10-26
Add Spark MAX Blink codes to status light quick reference?


mcm001:
Other 3rd party vendors have their blink codes on [the reference](https://frc-docs.readthedocs.io/en/latest/docs/hardware/hardware-basics/status-lights-ref.html), but not the spark max yet.

Starlight220:
#330 should be linked to this.

PR #330, opened 2019-11-12
Add SparkMax Color Codes

RishabRao:
Added sparkMax color codes

#297


AustinShalit:
I think that this image does not convey the information as clearly as the image on screen steps.

sciencewhiz:
How about using an animated gif so the blinking lights from Rev's website aren't lost? It would probably require a static png for pdf and the changes here: https://stackoverflow.com/questions/45969711/sphinx-doc-how-do-i-render-an-animated-gif-when-building-for-html-but-a-png-wh

Daltz333:
It's been decided that MP4s are the best solution due to play/pause support. Unfortunately, we're trying to figure out where to store these outside of git.

Starlight220:
Why is this marked with the `Waiting On: WPILib` label? What is it waiting on?

Daltz333:
We are still deciding on how to handle animations, as we are worried about 
- Students with flashing sensitivity (so we need play/pause support that some browsers don't support gif with)
- LaTeX/PDF support
- Where do we store these large gif files?

ItayZiv:
Are the gifs actually needed though? Wouldn't it be possible to just create a diagram showing the blink codes without actually needing the blinking?

Daltz333:
I think if we could just create a tablet or vector image containing descriptions and a brief color (or color codes such as a line through yellow to represent flashing yellow) would be sufficient. If this PR wants to get updated to reflect that, or anyone else wants to take it on.

Starlight220:
I might have the time to do this, but I'll probably need help with the format, etc.

Daltz333 (approved):

PR #343, opened 2019-11-20
Fix RMUsb link

Daltz333:


sciencewhiz:
The original link works for me. I don't think linking to the beta page is correct.

Daltz333:
Oh that's weird. That link was 100% invalid for me and azure an hour ago.

PR #346, opened 2019-11-22
Limit image file size, allow gifs in styleguide

Daltz333:


AustinShalit:
I'd like to get more feedback from others on this PR because this will affect teams that use PDFs.

Daltz333:
@AustinShalit I will expand open a PR to set PNG as default for latex builder. Instructions on including two image files will be included in this one.

Daltz333:
#408 supercedes this

PR #347, opened 2019-11-23
Fix PDP getting constructed with non 0 CAN ID

ThadHouse:
Confusing, especially with the warning right below


sciencewhiz:
I think that warning is old and should be removed. Supporting non-0 CAN IDs was fixed a few years ago. https://github.com/wpilibsuite/allwpilib/commit/39130721d0477b0e57fbaf31c16975a0c943a347

ThadHouse:
We should probably actually reword it to you must have id 0 in order to be seen by the ds

PR #350, opened 2019-11-23
Add vendor reference documentation

PeterJohnson:
This is reference material for vendors (or teams acting as vendors) regarding
how to integrate with the rest of the WPILib ecosystem.

Still a WIP, only covers simulation GUI integration so far.

### DIFF
--- a/source/docs/vendor/getting-started.rst
+++ b/source/docs/vendor/getting-started.rst
@@ -1,0 +1,3 @@
+Vendor Getting Started
+======================
+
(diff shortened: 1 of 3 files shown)


PeterJohnson:
Key additional topics: detailed technical description (library structure for C++ and JNI, what libraries should depend on what, the vendor deps JSON format, how vendor libraries get deployed to the robot, cross compilation), how to use the vendor gradle template, offline install requirements including where files need to get installed to on team machines

Daltz333:
@PeterJohnson should we close this and put these docs in allwpilib instead, as per an earlier meeting about storing developer documentation there?

Daltz333:
Determined that this will be moved and redone in allwpilib

PR #375, opened 2019-12-15
Add feedforward intro

calcmogul:


Oblarg:
I feel like this should be part of a separate section that is devoted purely to concepts, rather than to implementation.

calcmogul:
Three small paragraphs doesn't really warrant a whole separate page.

Oblarg:
There's more than can be said than those three small paragraphs, though, and it could be located in the same place as the conceptual PID docs.

calcmogul:
That's all there is to say really, unless you want to get into plant inversion and a lot of other messy stuff.

Oblarg:
It's worth going over the derivations for the models for common feedforwards, I think.

AustinShalit:
@Oblarg Can you open an issue to add more information. I agree that this information is useful here.

PR #383, opened 2019-12-22
Tweaks to New for 2020 suggested by Joe H.

Kevin-OConnor:
Adjust the LED bullet and add one about control packet caching per review by Joe H.

### DIFF
--- a/source/docs/software/wpilib-overview/new-for-2020.rst
+++ b/source/docs/software/wpilib-overview/new-for-2020.rst
@@ -1,7 +1,7 @@
 New for 2020
 ============
 
-A number of improvements have been made to FRC Control System software for 2020. This article will describe and provide a brief overview of the new changes and features as well as a more complete changelog for C++\\Java WPILib changes.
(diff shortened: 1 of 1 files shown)


source/docs/software/wpilib-overview/new-for-2020.rst
```diff
@@ -1,14 +1,15 @@
 New for 2020
 ============
 
-A number of improvements have been made to FRC Control System software for 2020. This article will describe and provide a brief overview of the new changes and features as well as a more complete changelog for C++\Java WPILib changes.
+A number of improvements have been made to FRC Control System software for 2020. This article will describe and provide a brief overview of the new changes and features as well as a more complete changelog for C++\\Java WPILib changes.
```
Daltz333: Why not just use a front slash like everywhere else? Aka: /

Kevin-OConnor: Because I forgot that's what we did everywhere else. I'll update

PR #386, opened 2019-12-23
Update RamseteController constructor docs

prateekma:


Oblarg:
Should mention that these gains have been found to work well *when units are in meters*.

PR #391, opened 2019-12-23
Update odometry advice to use relativeTo method

Oblarg:
Per previous discussion, this is better advice than resetting the odometry.

### DIFF
--- a/source/docs/software/trajectory-end-to-end/creating-following-trajectory.rst
+++ b/source/docs/software/trajectory-end-to-end/creating-following-trajectory.rst
@@ -104,7 +104,7 @@ Creating the RamseteCommand
 
 .. todo:: link to ``relativeTo`` api docs once they're uploaded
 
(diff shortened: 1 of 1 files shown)


prateekma:
A robot centric to field relative transformation should use the `transformBy` method, not the `relativeTo` method. Using `relativeTo` will cause all poses to be negative since the origin relative to any positive pose is a negative transform.

Oblarg:
`transformBy` takes a `Transform2D` rather than a `Pose2D`, though; are they implicitly convertible?

prateekma:
If the origin is at (0,0), then yes. Alternatively, you can pass in `robotPose.minus(startPoseOfRobotCentricTraj)` which automatically creates the transform.

AustinShalit:
@prateekma This look good now?

prateekma:
Yes

PR #400, opened 2019-12-30
Add usage reporting info for Kotlin teams

prateekma:
Let me know if this belongs in a better directory than where I put it.


Kevin-OConnor:
I'm not sure this actually belongs in our docs at all. It seems like this should go with whatever documentation is telling people how to use Kotlin in the first place

prateekma:
There isn't any documentation telling teams how to use Kotlin for FRC; it's just something that teams have figured out themselves. I thought it would be a good idea to include this in the official docs to get an accurate number for teams using Kotlin. I assume this is something that is desirable since the number was added to usage reporting.

sciencewhiz:
Here's two:

https://www.chiefdelphi.com/t/paper-kotlin-for-frc-robot-programming/166496
https://www.chiefdelphi.com/t/kotlin-for-frc-vscode-extension/335602

prateekma:
I suppose. I doubt that many Kotlin teams are aware of those threads and I don't think anything other than frc-docs will reach out to more Kotlin teams;l (reaching out to more teams means more accurate data from usage reporting); however, I understand the hesitation due to providing resources for an unsupported language. I am closing this PR now, but someone can reopen this if anyone changes their mind.

sciencewhiz:
The gradlerio kotlin example readme might be another place for it.

PR #408, opened 2019-12-31
Supports GIFs and specify maximum image size

Daltz333:
- Gifs are now allowed, they require extra steps.
- Limits image sizes to 3MB to avoid bloat. Resolutions with a file size higher than 3MB are not seen on output and just provide bloat.
- Recommends that images be artifact free.


Daltz333:
This will either be closed or modified in favored of low size embedded MP4s. This allows pause/play capability and stricter browser controls.

PR #453, opened 2020-01-09
Add a friendly 404 page

Daltz333:
Adds the ``sphinx-notfound-page`` extension. Please note that the 404 page only views properly when deployed. A viewable version can be found below.

https://daltz-private.readthedocs.io/en/404-redirect/demo.html


sciencewhiz:
Where is the emoji from? It looks like the facebook one, which we can't use.

Kevin-OConnor:
Fine with using the XKCD as it properly complies with the license. Can we put the text above it though (other than the credit)? It probably looks fine on desktop, but on mobile it's a decent scroll to get to our real message.

PR #457, opened 2020-01-12
Revamp PathWeaver docs

prateekma:


prateekma:
Revamped the most important stuff that users might be concerned about. Didn't really touch path groups as that didn't need updating much.

PR #460, opened 2020-01-12
Add known issue for C++ command based buttons

AustinShalit:


sciencewhiz:
Seems like the why is too much detail for a known issue.

Oblarg:
There is a workaround; simply create an ordinary `Button` class and pass the joystick button state as a lambda to the constructor.

PR #461, opened 2020-01-12
Add known issue for RobotBuilder custom components

AustinShalit:


sciencewhiz:
I would use extensions rather then custom components.

PR #467, opened 2020-01-14
Added Keyboard Shortcut for MacOS installation 

RishabRao:
added the cmd + shift + H shortcut to the finder tab to get to home. Makes it easier to get to home for less experienced users.


sciencewhiz:
The way it's worded, it sounds like pressing cmd+shift+h will magically copy the stuff to the right folder.

AustinShalit:
I would recommend teams use the Go menu in the menu bar rather than a shortcut.

RishabRao:
> I would recommend teams use the Go menu in the menu bar rather than a shortcut.

I have seen instances of the home folder not showing up on the finder windows leading to confusion from programmers on my team. If the shortcut is also provided they have an alternative.

AustinShalit:
Yup. Which is why I think the go menu will work the best.

Daltz333:
@RishabRao any update on the addressed comments?

Daltz333:
Closing as stale.

PR #471, opened 2020-01-15
Add docs on transforming trajectories

prateekma:


source/docs/software/advanced-control/trajectories/transforming-trajectories.rst
```diff
@@ -0,0 +1,61 @@
+Transforming Trajectories
+=========================
+Trajectories can be transformed from one coordinate system to another using the ``transformBy`` and ``relativeTo`` methods. These are useful for converting field-oriented trajectories into a robot-relative trajectories and vice versa.
+
+.. note:: It is recommended to always generate trajectories with waypoints relative to the field as opposed to waypoints relative to the current robot pose. However, in certain cases like aligning on-the-fly with Vision, it might be easier to generate a robot-relative trajectory and then transform it into a field-relative trajectory. Field-relative trajectories are recommended because they are easier to follow, as the robot pose via odometry is defined with respect to the field.
```
sciencewhiz: Unless there are errors in transformation, why is it easier to follow a field-relative trajectory?

prateekma: Because path following uses odometry and odometry is typically relative to the field. For a robot relative trajectory, you have to keep resetting your odometry to zero.

agasser: How can a robot follow a field-relative trajectory unless it starts at the trajectory's start point?

prateekma: You typically don't. If your starting location is a little off, Ramsete corrects for this. You transform a robot-relatuve trajectory to a field-relative trajectory using the robot pose so that the trajectory's starting location is the same as the robot pose.

agasser: > You transform a robot-relative trajectory to a field-relative trajectory using the robot pose so that the trajectory's starting location is the same as the robot pose.

Did you mean to say this the other way around?
> You transform a _field_-relative trajectory to a _robot_-relative trajectory using the robot pose so that the trajectory's starting location is the same as the robot pose.

See the confusion in your PR? The proposed doc says, "Field-relative trajectories are recommended because they are easier to follow," but you typically do not follow a field-relative trajectory. You follow a robot-relative trajectory, but it may have originated as a field-relative trajectory. It may be just the wording.

prateekma: You do always follow field-relative trajectories. Trajectories rarely start at (0,0,0). Can you point out exactly where in the proposed doc the confusion lies?

prateekma: When you create a trajectory in PathWeaver, it is field-relative. The only place you might generate a robot-relative trajectory is when using vision processing with the camera on the robot. 99% of teams won't use on-the-fly trajectory generation so they will only interact with field-relative trajectories.

If you are a team that generates on-the-fly trajectories, you should transform the trajectory to be field-relative because your odometry is defined with respect to the field. Because path following uses odometry, doing this is desirable.

You would typically not convert a field-relative (think PathWeaver) trajectory into a robot-relative trajectory, but that functionality does exist if someone wants to use it for whatever reason.

agasser: This is confusing because the robot's current pose should be a position that is relative to the field:
> It is recommended to always generate trajectories with waypoints relative to the field as opposed to waypoints relative to the current robot pose.

Take out this statement, it doesn't make sense to say a field-relative trajectory is easier to follow. A trajectory cannot be reliably followed unless the robot's pose matches the starting pose.
>Field-relative trajectories are recommended because they are easier to follow, as the robot pose via odometry is defined with respect to the field.

Proposed wording:
> It is recommended to always generate trajectories with waypoints relative to the field as opposed to waypoints starting at (0, 0, 0°). However, in certain cases like aligning on-the-fly with Vision, it might be easier to generate a trajectory with waypoints starting at (0, 0, 0°) and then transform it into a trajectory that is relative to your robot's current pose on the field.

Here is my rationale:
The confusion is in _robot-relative_. I think your intended meaning is that the trajectory and the robot start at (0, 0, 0°). However, I read _robot-relative_ to mean the starting pose is the robot's current pose.

Consider this scenario. I generate a trajectory in PathWeaver from (1, 1, 90°) on the field to (2, 2, 90°). I now have a field-relative trajectory. How do I go about using this with my robot when it's sitting at (4, 4, 90°)? I see three options, but please tell if I am incorrect:

1) I drive my robot to the starting position of (1, 1, 90°) and then run the field-relative trajectory as is. Per my definition, this resulting trajectory is _both_ field-relative and robot-relative because the starting pose is relative to the field, and the starting pose is my current pose. (If I'm a little off, at say (.9, 1.2, 86°) Ramsete should correct and get me to (2, 2, 90°).)
2) I run the trajectory starting at the robot's current pose by transforming the field-relative PathWeaver trajectory to a robot-relative trajectory using my current pose of (4, 4, 90°). This results in a trajectory with a start of (4, 4, 90°) and an end of (5, 5, 90°). Again, per my definition this resulting trajectory is _both_ field-relative and robot-relative.
3) My final option is to reset my robot's current pose to (0, 0, 0°), then transform the PathWeaver field-relative trajectory to a robot-relative trajectory using (0, 0, 0°) as the starting point. Per both of our definitions, this is a robot-relative trajectory and not field-relative. This is the worse scenario because I have now lost my robot's current position relative to the field.

prateekma: > The confusion is in robot-relative. I think your intended meaning is that the trajectory and the robot start at (0, 0, 0°). However, I read robot-relative to mean the starting pose is the robot's current pose.

Ah, I get the source of the confusion now. I can see how the words "robot-relative" can be interpreted as such (which is perfectly valid), and this is where our definitions of "robot-relative" differ.

Scenario 1 sounds correct with your definition of robot-relative, but not mine, as I established in the previous paragraph. Scenario 2 is also reasonable in that you could transform the trajectory such that the starting poses match -- you are correct in that the resulting trajectory is field-relative and robot-relative (according to your definition of robot-relative). Your description of Scenario 3 is also accurate.

I now foresee the confusion that the whole "field-relative" and "robot-relative" terminology might bring, even to experienced users. I think it would generally be a good idea to simply outline how `transformBy` and `relativeTo` work in relation to coordinate systems in general, without presenting the idea of field-relative vs. robot-relative. What are your thoughts on this? @agasser

prateekma:
@agasser I pushed an update to make the docs talk more about the `transformBy` and `relativeTo` methods with respect to coordinate systems vs. robot-relative and field-relative. Hopefully this will be less confusing. Please let me know with your feedback.

prateekma:
(LinkCheck broken because some random link that is not associated with this PR decided to not work anymore)

Oblarg:
I think it could benefit from some pictures, to be honest.  I also think that it might be worth mentioning that the two methods are, in a sense, inverse operations (even though they take different argument types, there's an obvious canonical conversion from a pose to a transform, i.e. subtracting the origin).

agasser:
@prateekma The new wording is much more clear, I like it.

I agree with @Oblarg. a picture would help a lot. I thought I understood what `relativeTo` does, but when I ran it I didn't get the result I expected. The example isn't complete enough to describe the scenario and the expected result.

prateekma:
@Oblarg @agasser Added some diagrams that should hopefully clear things up.

Oblarg:
Looks good.

agasser:
Those pictures are hugely helpful! Thanks for adding them.

PR #483, opened 2020-01-18
Add reference to importing JSON in trajectory tutorial

prateekma:


sciencewhiz:
It would help to have a link from pathweaver into the trajectory docs as well.

prateekma:
@sciencewhiz That already exists here: https://docs.wpilib.org/en/latest/docs/software/wpilib-tools/pathweaver/integrating-robot-program.html

PR #485, opened 2020-01-18
Use cameraserver.CameraServer instead of wpilibj.CameraServer

PeterJohnson:
The latter is deprecated.


sciencewhiz:
Should probably also change from IterativeRobot to TimedRobot

PR #491, opened 2020-01-20
Remove hardcoded path to use getDeployDirectory()

dzil123:
I don't know C++, so I can't port the Java change.


sciencewhiz:
I think it's still useful to mention where the data is deployed to

Here's the C++ deploy description:

```Files placed in this directory will be deployed to the RoboRIO into the
  'deploy' directory in the home folder. Use the 'frc::filesystem::GetDeployDirectory'
  function from the 'frc/Filesystem.h' header to get a proper path relative to the deploy
  directory.```

PR #495, opened 2020-01-21
Fix line numbers

Oblarg:
These had been thrown off by an include change


Oblarg:
Yes; I thought I had changed them; did I miss a few?

PR #499, opened 2020-01-21
move roborio docs to hardware

Oblarg:


sciencewhiz:
@Daltz333 is there a way to add a redirect in this PR? I don't think we should be moving articles in season without a redirect.

Daltz333:
@sciencewhiz I can add a redirect, but it's a bit less relevant now that we have the stable and latest versions of the documentation. Would you still like me to add the redirect?

Daltz333:
Closing as stale, feel free to re-open.

PR #508, opened 2020-01-24
Add note about LQR gains being aggressive

calcmogul:
971's "Design for Controlability" presentation goes into a lot more detail on the points raised here. https://www.youtube.com/watch?v=VNfFn-gcfFI

### DIFF
--- a/source/docs/software/wpilib-tools/robot-characterization/analyzing-feedback.rst
+++ b/source/docs/software/wpilib-tools/robot-characterization/analyzing-feedback.rst
@@ -10,6 +10,8 @@ These can be used to calculate optimal feedback gains for a PD or P controller f
 
 .. important:: These gains are, in effect, "educated guesses" - they are not guaranteed to be perfect, and should be viewed as a "starting point" for further tuning.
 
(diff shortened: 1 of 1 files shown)


Oblarg:
Perhaps add "If the system cannot be made more rigid, the gains can always be lowered."

sciencewhiz:
If you filter a sensor with agressive gains, isn't it even more likely to go unstable due to the phase delay?

calcmogul:
Depends on what kind of filtering and how much.

prateekma:
I think that lowering the gains will be the quickest and most effective way to reduce the aggressiveness of the LQR gains for most users -- so this suggestion should definitely be added as well.

calcmogul:
An IIR filter with a fast pole does well. Kalman filters combine this with a model to help with phase delay, but Kalman filters my team has applied to fast, aggressive control loops generally trust the measurements much more than the model (it's just how the relative weights for the process noise vs measurement noise have worked out).

sciencewhiz:
Is this superseded by #550>

AustinShalit:
@calcmogul 

>Is this superseded by #550

calcmogul:
Yes.

PR #511, opened 2020-01-24
improve viewing diagnostics article

Oblarg:
Adds section for common errors and their diagnostic plots.

PR #512, opened 2020-01-24
add EMI warning in mag encoder docs

Oblarg:


sciencewhiz:
I think this is overly alarmist. It should mostly affect measuring acceleration, which most people don't do (besides the characterization tool). Based on that, I think it's better in the characterization docs.

Oblarg:
It's definitely not "mostly affecting acceleration."  Look at the magnitude of the velocity noise in the plots.

image

Daltz333:
Closed as #511 addresses this

Issue #514, opened 2020-01-24
Test mode and dashboard documentation unclear and out of date


agasser:
There are several places in the documentation that reference `addActuator()` or `addSensor()` but these methods no longer exist.
- [Enabling Test mode (LiveWindow)](https://docs.wpilib.org/en/latest/docs/software/wpilib-tools/smartdashboard/test-mode-and-live-window/enabling-test-mode.html)
- [Using Test Mode](https://docs.wpilib.org/en/latest/docs/software/wpilib-overview/using-test-mode.html)

It's also unclear if subsystems, actuators, and sensors _need_ to be added with code, or if they will automatically populate.
- [This page](https://docs.wpilib.org/en/latest/docs/software/wpilib-overview/using-test-mode.html#adding-test-mode-code-to-your-robot-code) says to be sure to add your subsystems.
- [This page](https://docs.wpilib.org/en/latest/docs/software/wpilib-tools/smartdashboard/test-mode-and-live-window/enabling-test-mode.html#explicitly-vs-implicit-test-mode-display) says everything will automatically populate, but you can customize the grouping.
- [This page](https://docs.wpilib.org/en/latest/docs/software/wpilib-tools/shuffleboard/advanced-usage/shuffleboard-commands-subsystems.html#subsystems-in-test-mode) says RobotBuilder will generate the code but you can do it yourself with `setName()` but it's not clear if I do that on just my subsystems or on all of my actuators and sensors.
- There is a [Shuffleboard](https://github.com/wpilibsuite/allwpilib/blob/master/wpilibjExamples/src/main/java/edu/wpi/first/wpilibj/examples/shuffleboard/Robot.java) example, but it doesn't seem to make use of the subsystem grouping, although I haven't tried executing it.

This [Test Mode in SmartDashboard](https://docs.wpilib.org/en/latest/docs/software/wpilib-overview/using-test-mode.html#test-mode-in-smartdashboard) section seems to be missing a screenshot. It says, "The above sample code produces the following output when the Driver Station is put into Test mode then enabled.", but there is no screenshot or output. Also, the [code sample](https://docs.wpilib.org/en/latest/docs/software/wpilib-overview/using-test-mode.html#adding-test-mode-code-to-your-robot-code) has a markup error so the tabs don't render correctly.

[...]

PR #521, opened 2020-01-26
Fix tab and removed method names

agasser:
This is related to issue #514. It fixes a couple of the basic issues.


sciencewhiz:
Overall, I think this article needs a bigger overhaul. Putting a subsystem to SmartDashboard doesn't add the items to LiveWindow, they are added automatically. Putting a subsystem to SmartDasbhoard shows the scheduling status of that subsystem (what the default command is, and which commands its running).

agasser:
> Overall, I think this article needs a bigger overhaul.

I agree, that's why I wrote up #514 to describe my confusion with these features. In the interim, it didn't seem bad to correct some simple problems.

PR #536, opened 2020-02-03
Ubuntu install info for robot characterization

retrodaredevil:
When installing the robot characterization with `pip3 install frc-characterization` on Ubuntu, you also have to install the `python3-tk` package. This PR adds clarification that you have to use the `pip3` command on Ubuntu distributions and that you will also have to manually install `python3-tk`.


Daltz333:
Looks fine to me. Probably should be a ```.. note``` block since Ubuntu specific

retrodaredevil:
I will change it to a note block sometime today

retrodaredevil:
I changed it to a note block. Let me know if I need to change anything else @Daltz333

PR #538, opened 2020-02-06
Update linux documentation for latest release

williamhaley:
Updates the text and image to refer to the latest allwpillib update in GitHub since a couple have been pushed since the kickoff release

The Linux instructions could probably be made generic to avoid constant updates like this.

And perhaps the releases in GitHub shouldn't be named `Update`? I wonder if that's confusing to users who may assume they have to first download the 2019.1.1 kickoff release and incrementally update it.


williamhaley:
See screenshots of the build docs on my local. I didn't update the extraction screenshot, but could if that's important.

# screenshot 1
screenshot1

# screenshot 2

screenshot2

Daltz333:
The goal would be to be version agnostic, replace the version numbers with something like ```
**WPILIB\_Linux-VERSIONUM**
```

williamhaley:
Yeah, I was thinking a pattern would be better, but I thought explicit version numbers would be less ambiguous. I can update to make this more generic.

williamhaley:
Updated with a generic screenshot for extraction (hides the version number) and a generic GitHub release entry (I mangled the HTML to hide obvious year indicators) screenshot.

Updated the copy to use a generic version number placeholder. I was thinking of trying a wildcard pattern for `mv`, but I think that could be risky if users have multiple extracted folders in `Downloads` (or wherever). I thought it better to leave it to the user to fill in the blank, but I can update as needed.

PR #561, opened 2020-02-21
Add note that the build tools are not sufficient

virtuald:


sciencewhiz:
What parts are needed?

virtuald:
Dunno, @ThadHouse knows?

ThadHouse:
You need to install Visual Studio and select the C++ workload

PR #574, opened 2020-03-13
Bump up LinkCheck timeout

Daltz333:
Azure link checking is slow sometimes. Going to test this a couple times to see how it runs.


sciencewhiz:
https://www.microsoft.com/accessories/en-us/products/webcams/lifecam-hd-3000/t3h-00011#techspecs-connect takes around 18 seconds to load on my computer. I increased the timeout to 30 seconds and it passed.

sciencewhiz:
3 times in a row. I wonder if we can get rid of the ignore for kauailabs.com

PR #578, opened 2020-03-19
Limit top contributors to 10

sciencewhiz:
Should limit the linkchecks to github


source/docs/contributing/top-contributors.rst
```diff
@@ -2,4 +2,4 @@ Top Contributors
 ================
 
 .. ghcontributors:: wpilibsuite/frc-docs
-   :limit: 100
+   :limit: 20
```
Daltz333:
```suggestion
   :limit: 10
```

I think further reducing it to 10 is still quite substantial.

PR #580, opened 2020-03-21
Specify only one linkcheck worker

Daltz333:
Looks like having five threads had us over GitHub's request/second quota.


Daltz333:
Looks like this may need to be combined with #578 to stabilize linkcheck

Issue #263, opened 2019-10-12
Command Groups - Update Composition Diagram to Clarify Behavior of "ParallelCommandGroup"


PeterJohnson:
[Duplicated from artf5901 on the beta test site]

On the 2020 Online Documentation web page for Command Groups at https://frc-docs.readthedocs.io/en/latest/docs/software/commandbased/command-groups.html#recursive-composition-of-command-groups the diagram depicting the control flow of the "command group with concurrency" is a bit misleading.  In the example, there is a SequentialCommandGroup of DriveToGoal(), ParallelCommandGroup(RaiseElevator(), SetWristPosition()), and ScoreTube().  The diagram (as of the time of this tracker item) implies that the ScoreTube() command commences after the RaiseElevator() command finishes, regardless of the status of the SetWristPosition() command.    The diagram should be clarified to indicate that the ScoreTube() command does not commence until both the RaiseElevator() and SetWristPosition() commands have finished.  

I have attached a couple images which depict (a) the referenced image at the time of this writing, and (b) an updated image showing what the diagram could be clarified to look like.  (Both are attached.)  Obviously, the updated image would look better than what I have provided if the original editing tool to generate the image could be used instead of my crude markup of the original image in MS Paint.

![image](https://user-images.githubusercontent.com/297126/66708176-269bb680-ed01-11e9-8e6d-9a10329074fe.png)

![image](https://user-images.githubusercontent.com/297126/66708172-18e63100-ed01-11e9-8ac0-0d24f14bd4d0.png)

(Incidentally, the original image is a reasonable depiction of the control flow if the composition were ParallelDeadlineGroup(RaiseElevator(), SetWristPosition()), but I wouldn't recommend using the DeadlineGroup example as the basic example of parallel composition.)

Daltz333:
@oblarg

Oblarg:
Yeah, this should be fixed.  I'm terrible at making diagrams, though.

sciencewhiz:
Originally reported by @coach_streeter

Modelmat:
The original block diagram has been recreated in GraphViz. If anyone wants to PR it to my branch to add the second text feel free.
If not I will (next month).

PR #587, opened 2020-03-24
Change images for ParallelCommandGroup and Subsystems and Commands

ItayZiv:
Changes the image for ParallelCommandGroup flow chat to represent the actual behavior, where the final command happens after both commands in the CommandGroup finishes, also changes the image for Subsystems and Commands in "What is "Command-Based" Programming?" to look similar.
Both images use the same format as the rest of the doc (Lato).

fixes: https://github.com/wpilibsuite/frc-docs/issues/263

Edit:
I used SVGs because vector graphics are most fitting for this type of image, not sure if PNG is preferred here.


Daltz333:
Looks like PDF support is yelling about SVG. PNG should be fine in the meantime.

calcmogul:
To include SVGs in my LaTeX book, I have to convert them to PDFs with inkscape, then use `\includegraphics{}` on the PDF files. I use `inkscape -D -z --file=filename.svg --export-pdf=filename.pdf` to convert them, with some makefile rules to generate it all for me.

ItayZiv:
I changed the SVGs to quite high res PNGs, doesn't take a lot of space since its quite simple and should be good enough quality-wise, I'm still keeping the SVGs if they are needed at some point.

ItayZiv:
Also, I noticed the old command-based docs has much more pictures explaining how stuff work than the new one, is it worth adding more images to the new command-based docs?

PR #594, opened 2020-04-02
Move VS Code documents into their own section

Daltz333:
This PR moves the various articles and documents into their own section. WPILib overview still exists to include current issues, the introduction on wpilib, and 2020-game-data as those information are agnostic to build language and system.


AustinShalit:
What is WPILib? should probably not be in the VS Code section. I am also confused why the 2020 Overview is in the software directory.

Daltz333:
I am unsure of where the what-is-wpilib document should go. I can move the 2020-Overview up.

PR #596, opened 2020-04-02
Update doc8 dependency

Daltz333:


Daltz333:
doc8 still needs to push to pypi

PR #609, opened 2020-04-11
Use dynamic width for sidebar version picker

### LINKED ISSUE #608: Broken sidebar CSS in mobile view
Screenshot_20200410-135357_Firefox

Daltz333:
Closes #608


sciencewhiz:
Is there a way to test this on mobile?

RyanHir:
As you are using Firefox I would recommend using the built in Firefox mobile emulator as specified in https://developer.mozilla.org/en-US/docs/Tools/Responsive_Design_Mode

Daltz333:
A preview can be found at https://daltz-private.readthedocs.io/en/fix-css/

ezgif-6-480ef614ad0a

Daltz333:
Please note that this isn't supported on landscape mode of smaller devices, as it'll use 85% of the screenwidth, but that's an edge-case and not really a supported environment for webbrowsing on the web in general.

PR #610, opened 2020-04-13
Add SVG support

Daltz333:
Adds support through the imgconverter extension. No additional dependencies are required as it's bundled with Sphinx. This will be supported alongside drawio.


Daltz333:
Blocked by https://github.com/sphinx-doc/sphinx/issues/7477

azure-pipelines.yml
```diff
@@ -94,5 +94,11 @@ jobs:
     - script: pip install -r source/requirements.txt
       displayName: 'Install Python Requirements'
 
+    - script: sudo apt-get update
+      displayName: 'Update System Packages'
+
+    - script: sudo apt-get install -y librsvg2-bin
+      displayName: 'Install SVG requirements'
+
```
AustinShalit, line 102: Do we need this block? it should've been installed above.

Daltz333: It only installs for the build script, linkcheck for some reason runs librsvg2

PR #631, opened 2020-04-20
change RobotDrive to DifferentialDrive

### LINKED ISSUE #525: Use DifferentialDrive instead of RobotDrive
RobotBuilder has lots of outdated UI content, as well as code-generation. Additionally needs to be updated for the update to the new command based.

Starlight220:
Fixes #525.
Maybe rename the object?
The Subsystem class is named Drive Train - how can it be two words?


Daltz333:
Do you plan on updating this, or can I close this so other individually could potentially fix this?

Starlight220:
I don't really have time to fix this. Is there a reason to leave this for someone to take over my branch, or are the changes to negligible that whoever fixes this can easily redo them? If there is no point in having someone clone my branch, then this can be closed.

PR #635, opened 2020-04-22
Update gitignore to include common operating systems

AustinShalit:


AustinShalit:
_build can still get generated (it is on my computer because of Visual Studio plugin I use).

PR #200, opened 2019-09-20
Add details to Hardware section

yellowjaguar5:
See #40, #55 and #105.
Closes #285 

Todo:

- [x] Add image source files (.ai) to `images`/`source` directories.
- [x] Update image of roboRIO in all illustrations
- [x] Resize illustrations to be compatable with `imgconverter`
- [x] Each new image <100 KB
- [x] Add image of wiring to pneumatics article
- [x] Rename `image-1.svg` and `image-2.svg` to `mb1013-ultrasonic-sensor-to-roborio.svg`
- [x] Add illustration and replace this image `/hardware/hardware-basics/images/status-lights/rioLight.png`
- [x] Replace roborio images `/getting-started/getting-started-frc-control-system/images/imaging-your-roborio`
- [x] Resize images on Preemptive Troubleshooting page 
- [x] Resize images on coltrol-system-hardware page 
- [x] Split into 4 PRs: 
    1. Format hardware text and resize existing photos
    2. Add SVGs
    3. Fix up Status Light reference page
    4. Add info about IMUs

As always, suggestions are welcome and much appreciated.

PR #640, opened 2020-04-27
Format hardware section text and resize existing photos

yellowjaguar5:
This was split from #200.


yellowjaguar5:
@Daltz333 / @AustinShalit I believe I addressed all of your comments and this PR is ready for another review.

PR #646, opened 2020-05-01
Add WPILib RSN Videos

### LINKED ISSUE #645: Add RSN WPILib videos
Some WPILib developers presented at the RSN conference last week. We should add the videos to the documentation. They are located here: https://wpilib.org/blog/wpilib-team-presents-on-the-robosports-network

prateekma:
Closes #645


Daltz333:
Why not have the videos just in the index file?

prateekma:
I suppose that can also be done, but I felt that it would be cleaner to have one page with a link to the video and the slides.

Daltz333:
Alternatively, the page could be put first in the TOC as it introduced the content.

PR #673, opened 2020-05-11
Fix typos

ItayZiv:
Still gathering a few typos I/other have found.


Modelmat:
There's two more of the advanced controls > PID page:

```
plot. Onnce the system is closer to the reference, the controller will stop saturating and produce 
``` and
```This cab result in steady-state error as shown in figure 2.4```

File: source/docs/software/sensors/digital-inputs-software.rst [hunk 1]
```diff
@@ -146,7 +146,7 @@ Nearly all motorized mechanisms (such as arms and elevators) in FRC should be gi
             }
         }
 
-Homing an encodered mechanism
+Homing an encoded mechanism
```
sciencewhiz, line 149: I think encodered is intentional. encoded doesn't have the same connotation.

ItayZiv: Yea this was one I wasn’t sure about... I did it to get opinions and because technically encodered isn’t really a word and encoded is and it seems to me like it does fit

prateekma: Even though "encodered" might not be a word per se, it can generally be understood that it means "something with an encoder on it." I don't think "encoded" has that same meaning.

ItayZiv: I've gotten feedback from quite a few people that just omitting the word should be fine, and I think its fine since its placed in a section about encoders anyways

sciencewhiz:
Need to run make translate to fix the translation failure

Issue #672, opened 2020-05-11
PDF - SVG with embedded HTML tags


ItayZiv:
This issue isn't quite urgent since its related only to PDF (and like IE), but anyways...
Being vector-graphics SVGs cannot store data the same way as normal PNGs/JPEGs, this results in somewhat of an inconsistency between different programs that display SVGs. One of these inconsistencies comes in with HTML tags, which are used with more advanced programs that export to SVG, to format text, an example of a program that does this is draw.io (Which is the program the docs use for diagrams).
This is fine for all modern browsers, but for making PDFs we convert the SVG file to PNG for use in a PDF, the program used to convert (rsvg in our case, Inkscape suffers from this as well), does not support rendering HTML tags, in draw.io's case it causes a warning on all diagrams, but regardless any formatting that relies on HTML tags won't work.

Getting to the point, it's worth thinking about whether or not enforcing proper formatting is kept on PDF is important or not.
In draw.io, for example, any SVG can be exported without HTML tags (also removing the warning), by selecting everything and deselecting a checkbox when the diagram is done. If any formatting would break at PDF it would break when doing that as well, an example of formatting getting broken is as follows
The original diagram:
![originaldiagram](https://user-images.githubusercontent.com/29122290/81514407-a17c8c80-9337-11ea-9f25-9c7035e610bb.png)
After disabling formatting:
![brokediagram](https://user-images.githubusercontent.com/29122290/81514414-b1946c00-9337-11ea-87dc-5d3a207f0e9c.png)
The same diagram broken in a PDF:
![AcroRd32_03-05-2020_09_13_59_PM_de7e5c_1a5](https://user-images.githubusercontent.com/29122290/81514423-b9541080-9337-11ea-94d0-c83eee795235.png)

Since this regard PDF support its not urgent at all, but thinking of the way to go forward with this stuff is worth thought, while this issue is coming up right now with draw.io it has to do with SVGs in general. It is also possible that in the future rsvg or the like start supporting HTML tags, walkthrough it isn't something id count on.
This can all be ignored because of how minimal PDF support, diagrams still will appear, they just might have weird formatting issues.

PR #676, opened 2020-05-15
Remove Formatted Text from Drawio Diagrams

yellowjaguar5:
@ItayZiv noticed in #672 that drawio-generated SVG images are broken in the PDF build due to lacking support of the `Formatted text` SVG feature. This PR will remove formatted text from existing drawio diagrams and add a note to the style guide encouraging future contributions with drawio diagrams to turn off the `Formatted text` option.


Daltz333:
Stale @yellowjaguar5 ?

PR #690, opened 2020-05-19
Set resolution of images to 1200

Daltz333:
Maintains aspect ratio, no visible loss of quality in the documentation.


sciencewhiz:
Have you verified the no loss in qualify on a hidpi screen?

Daltz333:
I've noticed no visible loss of quality on my 1080p desktop monitor. There may be possible loss when viewing the source on larger displays, but that exceeds the use case.

PR #692, opened 2020-05-21
Minify SVGs

maniyar1:
Applied the `svgo` command-line tool to all svgs to reduce size by ~20%
on average. This would speed up page loading, though impact might vary
based on gzip compression. Ideally this is something that could also
happen automatically during build.

This might need to be re-run for PR #676
[Source for reference](https://victorzhou.com/blog/minify-svgs/)


AustinShalit:
Does this remove the meta data that makes the svgs editable?

maniyar1:
I'm not entirely sure what software is being used to create them, a quick look shows me that they are still easily editable in Inkscape. I do think that's a valid concern though, if needed I can try my hand at integrating it into the build system and keeping the metadata intact.

yellowjaguar5:
Unfortunately, applying the `svgo` does erase the `svg` metadata needed to edit the `drawio` `svg` diagrams. Also, `source/docs/software/basic-programming/diagrams/branches.drawio.svg` appears to have lost its formatting.
image

maniyar1:
Alright, that's understandable - I'll go through and (try to) add it to the build process for just the output, probably a better way to go anyway. As for the formatting, that's strange, it loads fine for me in Inkscape but bugs out just like the picture you sent in Firefox, I'll look into it.

ItayZiv:
It seems like on top of removing the “useless” drawio metadata. It might mess with some of the native html tags which drawio uses for formatting (not on my computer rn so cant verify)

maniyar1:
So after going down some rabbit holes I think I'll close the PR soon. I'm not exactly sure what's going on with the drawio svgs, xmllint using the svg 1.1 DTD gives me a lot of validity errors, but I'm not really confident in its analysis. 
Meanwhile, if I put the SVGs through [this sanitizer](http://svg.enshrined.co.uk/) and then minify them they work as intended, which is great. But I don't really like adding the dependency chain for somewhat marginal benefits (the gzipped files are technically "three times as small"...but when that's actually just 2KiB of savings I'm not sure if it's worth it). 
I can still put in the sanitizer and minify the SVGs in the build process if you think it's something that'll benefit the docs.

PR #694, opened 2020-05-24
Add instructions for setting up CI with robot code

### LINKED ISSUE #57: Add documentation on how to set up AzureCI for CI builds
Low on the priority list, but it gets asked often enough it would be work doing as part of the simulation work.

prateekma:
Closes #57 

Here is the sample repository that I used to make sure these instructions work: https://github.com/prateekma/MyRobotProject


AustinShalit:
Sweet. Can you open a PR to this repo too: https://github.com/wpilibsuite/Gearsbot

ItayZiv:
If it’s added to gearsbot, remoteLiteralInclude should be possible as well.

ItayZiv:
Also, it might be worth adding a note about grades build/check/test (and possibly this diagram from gradle’s user guide https://docs.gradle.org/current/userguide/img/javaPluginTasks.png , iirc the docs are CCv4)

sciencewhiz:
Since we don't have any instructions for writing a unit test, I think that should be included as part of this too.

prateekma:
I'm not sure if I would be qualified to write something for unit testing robot code. The only tests my team has done include testing general math/algorithm techniques and nothing related to the robot framework itself.

I think it would be better if someone else added that to a separate article within the same "Basic Programming" section.

ItayZiv:
I personally think putting in in another PR might be fine just because of how much extra work it is, also might be more fit under advanced-controls

PR #695, opened 2020-05-24
Update firewall screenshots to Windows 10

### LINKED ISSUE #656: Update networking/firewall screenshots for Windows 10
Use ..group-tab to distinguish between Windows 7 and Windows 10. Default to Windows 10

prateekma:
Closes #656


sciencewhiz:
This does not disable the domain firewall. To do that, you need to use Windows Defender Firewall with Advanced Security.

Since we still support Windows 7, I think the Windows 10 instructions should be added in addition too, rather then replace the Windows 7 instructions (although they should come first).

Daltz333:
You can do .. group-tab:: Windows 7

PR #696, opened 2020-05-24
Add image file size check

ItayZiv:
Fixes https://github.com/wpilibsuite/frc-docs/issues/693

PR #697, opened 2020-05-24
Reduce image size

yellowjaguar5:
To prepare for #696, existing images greater than 500 kB must be reduced in size. 
Conflicts with #690 will be resolved soon.


Daltz333:
I'd like to see this passing before merging the imagecheck PR

yellowjaguar5:
@Daltz333 Could you review this PR?

Daltz333:
I've taken a look at the differences between the original and your changes. Could you please elaborate on the changes you've made for history sake?

yellowjaguar5:
@Daltz333 For PNGs, I found all the images that were too large, made a few Adobe Photoshop batch actions to reduce resolution, and ran each image through the actions individually. For the drawio diagrams, I copied out embedded images, reduced their resolution, threw them into a base64 image encoder (like this one), and plopped the images back into the drawio file.

Regarding the checks in #696, this PR does not currently pass, but if specific paths are added to the exception list in `conf.py` and several of the drawio diagrams are reduced in size (or added to the exception list as well), it should pass the new `imagesize` check. 

Would you like me to try to reduce the drawio diagrams below 500 kB? It might make some images very low resolution and hard to understand. I personally would rather just add the larger drawio diagrams to the exception list in `conf.py`

Daltz333:
I'd be fine with adding to exceptions.

yellowjaguar5:
@Daltz333 Once `imagesize` check in #696 runs, this PR should now pass it.

Daltz333:
Ah, Could you revert the exclusion change? I'll merge this before the check PR that way nothing fails. The check PR can add the exclusions.

yellowjaguar5:
@Daltz333 Is this ready to merge?

PR #718, opened 2020-06-14
Implement Image Usage Checker

Daltz333:


scripts/imageusagechecker.py
```diff
@@ -0,0 +1,89 @@
...
+                else:
+                    excluded = True
+
+
+        if excluded:
+            continue
+
+        if not found:
+            print(source_pair[0] + " is not currently used!")
+            err = True
```
ItayZiv, line 81:
```suggestion
    for file_path in source_hashes.keys():
        if not any(exc in file_path.replace('\\', '/').lower() for exc in tuple(excluded_files)):
            if source_hashes[file_path] not in output_hashes.values():
                print(file_path + " is not currently used!")
                err = True
```
Untested, something like this *should* work

azure-pipelines.yml
```diff
@@ -111,6 +111,29 @@ jobs:
...
+
+      - script: pip install -r source/requirements.txt
+        displayName: 'Install Python Requirements'
+
+      - script: sudo apt-get update
+        displayName: 'Update System Packages'
+
+      - script: sudo apt-get install -y texlive-latex-recommended texlive-fonts-recommended texlive-latex-extra latexmk texlive-lang-greek texlive-luatex texlive-xetex texlive-fonts-extra dvipng librsvg2-bin
+        displayName: 'Install LateX'
+
```
ItayZiv, line 133:
```suggestion
```
No point in updating pip or sudo and installing requirements (and really no reason to install LaTeX) to run a python script

Daltz333: We have to run make HTML, and we need requirements otherwise the whole thing fails

scripts/imageusagechecker.py
```diff
@@ -0,0 +1,76 @@
...
+IMAGE_FORMATS = ['.png', '.jpg', '.svg']
+
+
+def clean_module_path(path):
+    if path.endswith(".py"):
+        path = path.replace("/", ".")
+    else:
+        path = path.replace("\\", "/")
+
+    return path[:-len(".py")]
```
ItayZiv, line 17:
```suggestion
    if path.endswith(".py"):
        path = path[:-len(".py")]

    return path.replace("/", ".").replace("\\", ".")
```

scripts/imageusagechecker.py
```diff
@@ -0,0 +1,74 @@
...
+    print("Configuration Path:", args["exclude_file"])
+
+    source_hashes = build_hashes(args["source"])
+    output_hashes = build_hashes(args["output"])
+    exclude_file = args["exclude_file"]
+
+    if exclude_file is not None:
+        excluded_regex = "|".join(list(importlib.import_module(clean_module_path(exclude_file)).exclude_patterns))
+    else:
+        excluded_regex = list()
```
ItayZiv, line 54:
```suggestion
        excluded_regex = ""
```
regex needs to be run on a str

scripts/imageusagechecker.py
```diff
@@ -0,0 +1,74 @@
...
+
+    err = False
+
+    print("Starting Image Scan\n")
+
+    for file_path in source_hashes.keys():
+        posix_path = file_path.replace('\\', '/').lower()
+        if re.search(excluded_regex, posix_path) is None:
```
ItayZiv, line 62: This does work, but id appreciate a link to https://github.com/sphinx-doc/sphinx/blob/275d93b5068a4b6af4c912d5bebb2df928416060/sphinx/util/matching.py#L67 for the sphinx impl in case issues arise in the future. (Sphinx accepts a glob-style match and translates that to regex, although in our use case using regex directly will work)
The impl could either be duplicated to here or since this test relies on sphinx anyways, it can be grabbed from sphinx.

scripts/imageusagechecker.py
```diff
@@ -0,0 +1,74 @@
...
+        for file in files:
+            filename, file_extension = os.path.splitext(file)
+
+            if file_extension in IMAGE_FORMATS:
+                filepath = os.path.join(root, file)
+                with open(filepath, 'rb') as afile:
+                    hasher = hashlib.md5()
+
+                    hasher.update(afile.read())
+                    located_files[filepath] = hasher.hexdigest()
```
ItayZiv, line 30:
```suggestion
                    located_files[filepath] = hashlib.md5(afile.read()).hexdigest()
```
Only optional, both should work

Daltz333: I'd rather keep the readability.

scripts/imageusagechecker.py
```diff
@@ -0,0 +1,79 @@
...
+    print("Directory One:", args["source"])
+    print("Directory Two:", args["output"])
+    print("Configuration Path:", args["exclude_file"])
+
+    source_hashes = build_hashes(args["source"])
+    output_hashes = build_hashes(args["output"])
+    exclude_file = args["exclude_file"]
+
+    if exclude_file is not None:
+        module_path = clean_module_path(exclude_file)
```
Modelmat, line 52: Why not use python's feature for this?
https://stackoverflow.com/a/67692/7349206

scripts/imageusagechecker.py
```diff
@@ -0,0 +1,79 @@
...
+
+    err = False
+
+    print("Starting Image Scan\n")
+
+    # It may be useful to use the Sphinx implementation in the future
+    # https://github.com/sphinx-doc/sphinx/blob/275d93b5068a4b6af4c912d5bebb2df928416060/sphinx/util/matching.py#L67
+    for file_path in source_hashes.keys():
+        posix_path = file_path.replace('\\', '/').lower()
+        if re.search(excluded_regex, posix_path) is None:
```
Modelmat, line 67: If this is a glob, why not use python's `glob` module?

scripts/imageusagechecker.py
```diff
@@ -0,0 +1,79 @@
...
+def clean_module_path(path):
+    if path.endswith(".py"):
+        path = path[:-len(".py")]
+
+    return path.replace("/", ".").replace("\\", ".")
+
+def build_hashes(dir):
+    located_files = {}
+
+    for root, subFolders, files in os.walk(dir):
```
Modelmat, line 20: The capitalisation here is wrong. (`sub_folders`).

scripts/imageusagechecker.py
ItayZiv: It might be possible, however the idea here was that no matter if a python file or a module path or whatever gets thrown it will work

scripts/imageusagechecker.py
ItayZiv, line 67:
```suggestion
        path = PureWindowsPath(file_path)
        if any(path.match(match) for match in globs):
```
Code for @Modelmat's suggestion
needs to import pathlib
`PureWindowsPath` is used because its not case sensitive, which is whats desired here

scripts/imageusagechecker.py
```diff
@@ -0,0 +1,79 @@
+import os
+import sys
+import hashlib
+import argparse
+import importlib
+import re
```
ItayZiv, line 6:
```suggestion
import pathlib
```

scripts/imageusagechecker.py
```diff
@@ -0,0 +1,79 @@
...
+
+    source_hashes = build_hashes(args["source"])
+    output_hashes = build_hashes(args["output"])
+    exclude_file = args["exclude_file"]
+
+    if exclude_file is not None:
+        module_path = clean_module_path(exclude_file)
+        globs = importlib.import_module(module_path).exclude_patterns
+
+        excluded_regex = "|".join(list(globs))
```
ItayZiv, line 55:
```suggestion
```

scripts/imageusagechecker.py
```diff
@@ -0,0 +1,79 @@
...
+    else:
+        excluded_regex = ""
```
ItayZiv, line 57:
```suggestion
        globs = []
```

scripts/imageusagechecker.py
Modelmat: Not exactly the right line, but for the other instance of this it should be `.join(globs)` because `.join` takes an iterable; all `list()` does it take an iterable -> list, so it wastes time and memory.

Issue #215, opened 2019-09-29
Update robotbuilder docs


Daltz333:
A lot of the examples still use IterativeRobot, and not TimedRobot. Should fix these. 

Additionally, the documentation needs to have updated images, and updated content for the RobotBuilder new command-based

yellowjaguar5:
@Daltz333 I believe this issue has been resolved.

PR #720, opened 2020-06-15
Use DifferentialDrive instead of RobotDrive

### LINKED ISSUE #525: Use DifferentialDrive instead of RobotDrive

yellowjaguar5:
Closes #525. I believe I replaced all instances properly (and in line with RobotBuilder output), but I would appreciate a verification from a reviewer.


sciencewhiz:
The problem is that the code that is there now is not what RobotBuilder generates now, even for Robot Drive. The code needs to be regenerated, not just find/replace.

Daltz333:
I'm going to tentatively close this as RobotBuilder documentation needs to be rewritten. See #215 for more information.

Issue #588, opened 2020-03-24
Update for Sphinx 3


Daltz333:
Blocked by
https://github.com/djungelorm/sphinx-tabs/pull/59
https://github.com/readthedocs/sphinx-notfound-page/pull/82
https://github.com/sphinx-doc/sphinx/issues/7369

Daltz333:
It may be needed to fork a couple of repos and maintain that. (sphinx-tabs and ogp)

PR #730, opened 2020-06-23
Update dependencies

Daltz333:


sciencewhiz:
Fixes #588

Daltz333:
Please note that Sphinx-tabs now throws a warning (doesn't fail build). This is tracked in https://github.com/djungelorm/sphinx-tabs/pull/59

It may be worth investigating forking sphinx-tabs and maintaining it to a degree.

PR #739, opened 2020-06-24
use azure templates in build

AustinShalit:


Daltz333:
We could probably use Python 3.7, I'm unsure about 3.8

PR #760, opened 2020-07-07
Add pressure transducer documentation

### LINKED ISSUE #627: Document Pressure Transducers
We should add some documentation about pressure transducers, what they do, and how to write code for them (AnalogInput & scale).

Starlight220:
Resolves #627 .
Probably needs some fixing with the API etc, reviews will help.


source/docs/software/actuators/pneumatics.rst
```diff
@@ -112,3 +112,24 @@ Double solenoids are controlled by the DoubleSolenoid class in WPILib. These are
         exampleDouble.Set(frc::DoubleSolenoid::Value::kOff);
         exampleDouble.Set(frc::DoubleSolenoid::Value::kForward);
         exampleDouble.Set(frc::DoubleSolenoid::Value::kReverse);
+
+
+Pressure Sensors
+----------------
+
+To read the pressure stored in the pneumatic system, connecting an analog pressure sensor is possible. These sensors connect to the Analog ports on the roboRIO, and can be read by the AnalogInput class in WPILib.
```
prateekma, line 120: Consider rewording the first sentence to "One can connect a pressure sensor to measure the pressure stored in a pneumatic system."

source/docs/software/actuators/pneumatics.rst
```diff
@@ -112,3 +112,24 @@ Double solenoids are controlled by the DoubleSolenoid class in WPILib. These are
...
+.. tabs:
+
+    .. code-tab:: java
+        import edu.wpi.first.wpilibj.AnalogInput;
+
+        AnalogInput pressureSensor = new AnalogInput(/* the AnalogIn port*/ 2);
+
+        //scaled units in the 0V-5V range
```
prateekma, line 128: Add space between `//` and start of comment

source/docs/software/actuators/pneumatics.rst
```diff
@@ -112,3 +112,24 @@ Double solenoids are controlled by the DoubleSolenoid class in WPILib. These are
...
+        double volts = pressureSensor.getVoltage();
+
+    .. code-tab:: c++
+        frc::AnalogInput pressureSensor {2};
+
+        //scaled units in the 0V-5V range
```
prateekma, line 134: Add space between `//` and start of comment

source/docs/software/actuators/pneumatics.rst
```diff
@@ -112,3 +112,24 @@ Double solenoids are controlled by the DoubleSolenoid class in WPILib. These are
...
+    .. code-tab:: java
+        import edu.wpi.first.wpilibj.AnalogInput;
+
+        AnalogInput pressureSensor = new AnalogInput(/* the AnalogIn port*/ 2);
+
+        //scaled units in the 0V-5V range
+        double volts = pressureSensor.getVoltage();
+
+    .. code-tab:: c++
+        frc::AnalogInput pressureSensor {2};
```
prateekma, line 132: Remove space between `pressureSensor` and `{2}`

sciencewhiz:
While pressure sensor is technically correct, I think it may be confusing with pressure switch. For that reason, I prefer the term pressure transducer. 

Add a reference to look at the sensor data sheet for how to turn voltage into pressure. While some are linear, I've used ones that aren't.

source/docs/software/actuators/pneumatics.rst
sciencewhiz: Specify Analog Input port On RoboRIO, since there are also Analog Output ports

Starlight220: Fixed both.

source/docs/software/actuators/pneumatics.rst
Starlight220: It was consistent with the other code samples. Should they also be changed?

Daltz333: What other code examples?

Daltz333:
The error is currently due to invalid code block syntax

Starlight220:
> Add a reference to look at the sensor data sheet for how to turn voltage into pressure. While some are linear, I've used ones that aren't.

Explain? I understood that it should be documented that teams should implement a function for `Volts -> PSI`, but something more concrete would be helpful.

source/docs/software/actuators/pneumatics.rst
Starlight220: The DoubleSolenoid, Solenoid, and Comperssor code examples. Literally every other code sample on the page.

sciencewhiz:
For example, this sensor provides the formula in the datasheet https://www.revrobotics.com/content/docs/REV-11-1107-DS.pdf
p = 250 * (V / 5.0) - 25

But this one does not provide a formula, but does provide a graph https://www.ebay.com/itm/Pressure-transducer-sender-150-psi-5V-stainless-steel-for-oil-fuel-air-water-/251362960044 from which you can derive a formula
p = 37.5 * (V - 0.5)

Fundamentally those are the same equation, just scaled for the maximum pressure of 200 PSI for the former and 150 PSI for the latter. However, other transducers may use other sensors and not have the same scaling or equation.

I was wrong when I said we used one that wasn't linear.

Starlight220:
So to add a concrete example featuring one of the sensors, or a general saying of "there's a formula, use it."
Anyway, what's the API for scaling/shifting AnalogInputs?

Starlight220:
thanks Austin for fixing my bad merge

source/docs/software/actuators/pneumatics.rst
```diff
@@ -112,3 +112,38 @@ Double solenoids are controlled by the DoubleSolenoid class in WPILib. These are
...
+        double psi = pressureTransducer.get();
+
+    .. code-tab:: c++
+
+        // product-specific voltage->pressure conversion, see product manual
+        // in this case, 250(V/5)-25
+        // the scale parameter in the AnalogPotentiometer constructor is scaled from 1 instead of 5,
+        // so if r is the raw AnalogPotentiometer output, the pressure is 250r-25
+        double scale = 250, offset = -25;
+        frc::AnalogPotentiometer pressureTransducer {/* the AnalogIn port*/ 2, scale, offset};
```
prateekma, line 146:
```suggestion
        frc::AnalogPotentiometer pressureTransducer{/* the AnalogIn port*/ 2, scale, offset};
```

Starlight220: @Daltz333 - what do you say about this change? All other cpp code is with 
```cpp
namespace::Type variableName {args}
```
format. 
If we want to remove the space before the arg braces, it should be done for all code on the page or even all code in frc-docs.

Daltz333: That is correct. Code should follow the WPILib style guide.

Starlight220: And wpiformat has no space before args?

PR #769, opened 2020-07-14
Add reference to SimulationPeriodic

sciencewhiz:


source/docs/software/commandbased/subsystems.rst
```diff
@@ -5,7 +5,7 @@ Subsystems are the basic unit of robot organization in the command-based paradig
...
+Subsystems can be associated with “default commands” that will be automatically scheduled when no other command is currently using the subsystem. This is useful for continuous “background” actions such as controlling the robot drive, or keeping an arm held at a setpoint. Similar functionality can be achieved in the subsystem’s ``periodic()`` method, which is run once per run of the scheduler; teams should try to be consistent within their codebase about which functionality is achieved through either of these methods. There is also a ``simulationPeriodic`` method that is similar to ``periodic`` except that it is only run during :doc:`Simulation </docs/software/wpilib-tools/robot-simulation/introduction>` and can be used to update the state of the robot. Subsystems are represented in the command-based library by the Subsystem interface (`Java <https://first.wpi.edu/FRC/roborio/release/docs/java/edu/wpi/first/wpilibj2/command/Subsystem.html>`__, `C++ <https://first.wpi.edu/FRC/roborio/release/docs/cpp/classfrc2_1_1Subsystem.html>`__).
```
Daltz333, line 8:
```suggestion
Subsystems can be associated with “default commands” that will be automatically scheduled when no other command is currently using the subsystem. This is useful for continuous “background” actions such as controlling the robot drive, or keeping an arm held at a setpoint. Similar functionality can be achieved in the subsystem’s ``periodic()`` method, which is run once per run of the scheduler; teams should try to be consistent within their codebase about which functionality is achieved through either of these methods. There is also a ``simulationPeriodic()`` method that is similar to ``periodic`` except that it is only run during :doc:`Simulation </docs/software/wpilib-tools/robot-simulation/introduction>` and can be used to update the state of the robot. Subsystems are represented in the command-based library by the Subsystem interface (`Java <https://first.wpi.edu/FRC/roborio/release/docs/java/edu/wpi/first/wpilibj2/command/Subsystem.html>`__, `C++ <https://first.wpi.edu/FRC/roborio/release/docs/cpp/classfrc2_1_1Subsystem.html>`__).
```

sciencewhiz: Is there a reason you only suggested simulationPeriodic() and not periodic()?

Daltz333: Didn't notice the other one.

PR #774, opened 2020-07-17
Rename translators document

Daltz333:


ItayZiv:
Shouldn't it be the same as it is with contributors?

Daltz333:
Contributors is sorted by top, which is fine. We don't want to do that with translators.

sciencewhiz:
If translators are limited to 20, isn't top translators (even if there's less then 20 today)

PR #779, opened 2020-07-27
Replace Azure with Actions

Daltz333:


.github/workflows/CI.yaml
```diff
@@ -0,0 +1,74 @@
...
+      run: |
+        make html
+    - name: Build PDF
+      run: |
+        make latexpdf
+    - name: Archive HTML
+      uses: actions/upload-artifact@v2
+      with:
+        name: frc-docs-html
+        path: build/html/
```
AustinShalit, line 33: This was before the make pdf call on azure to make sure the html artifact got uploaded even if the pdf failed.

PR #780, opened 2020-07-27
Force tag yearly releases

### LINKED ISSUE #722: Permalinks
Not sure if this should even be added or how, but it might be useful to have permalinks to certain articles (For example `docs.wpilib.org/permalinks/????/shuffleboard` would be linked to the index of shuffleboard no matter if it gets moved or renamed in the future).
This could also redirect automatically to the preferred stable/latest version of docs.
An example usage would be https://wpilib.org/ where right now there's a link to a specific page, where renaming or relocation would break the link.

Daltz333:
Closes #722 
Relies on #779

This commit pushes a yearly tag on every commit. This allows us to set "permalinks" that can still get updated.


.github/workflows/CI.yaml
```diff
@@ -0,0 +1,87 @@
...
+        make sizecheck
+  update-release:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push'
+    steps:
+    - uses: actions/checkout@v2
+    - uses: eine/tip@master
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
+        tag: '2021'
```
prateekma, line 83: You need to first manually create this tag.

Daltz333:
Closing as it was determined that this wasn't the correct solution

PR #784, opened 2020-08-02
Use experimental HTML4 writer by default

Daltz333:
- Fixes a bunch of issues ``sphinx_rtd_theme``


sciencewhiz:
Are there issue(s) open in spinx_rtd_theme?

Daltz333:
Doesn't work :(

PR #794, opened 2020-08-14
CI check links: fetch master instead of full checkout

TheTripleV:


TheTripleV:
This feels faster. I haven't actually measured the times of (checkout full repo) vs (checkout HEAD of pr and checkout HEAD of master)

Daltz333:
Check out is about 30-50s faster.

PR #806, opened 2020-08-23
Add Rediraffe Extension

### LINKED ISSUE #722: Permalinks

Daltz333:
This extension allows us to much more easily redirect content.
- Adds a check to verify that any deleted/renamed files get redirected.
- Navigates through a list of redirects to build a final redirect.

Closes #722


sciencewhiz:
Shouldn't this page https://frc-docs--806.org.readthedocs.build/en/806/docs/software/wpilib-overview/known-issues.html redirect to https://frc-docs--806.org.readthedocs.build/en/806/docs/2020-overview/known-issues.html? I get a 404?

PR #817, opened 2020-08-31
Add pip3 install to instructions

RyanHir:
On some Debian based distros, python3-pip is not preinstalled. This commit adds this to the build instructions in the readme to include the  `python3-pip` package in the required packages.


Daltz333:
I think it's better to just require Python 3.5 or greater, since we don't test with anything less than that.

- [ ]

RyanHir:
Uh oh, I will change this back to a draft as I fix that.

RyanHir:
It appears that file you requested already has the patch applied.

Daltz333:
Yep, just noticed. Okay. Waiting for build to complete.

PR #854, opened 2020-10-01
Display warning marking latest as unstable.

### LINKED ISSUE #828: Add Sphinx Extension for the "latest" version declaring not stable.
We don't need to create this ourselves, as it exists somewhere. I just cannot remember what it's called but I believe it's published by a maintainer of RTD. Should be a fairly easy issue.

Daltz333:
Closes #828

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -42,7 +42,23 @@ extensions = [
     'sphinxext.linkcheckdiff',
     'hoverxref.extension',
     'notfound.extension',
+    'versionwarning.extension',
 ]
+
+versionwarning_messages = {
+    "latest": """
+        The documentation you are currently viewing is for upcoming changes to WPILib.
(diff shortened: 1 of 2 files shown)


source/conf.py
```diff
@@ -42,8 +42,18 @@
     'sphinxext.linkcheckdiff',
     'hoverxref.extension',
     'notfound.extension',
+    'versionwarning.extension',
 ]
 
+versionwarning_messages = {
+    "latest": "The documentation you are currently viewing is for upcoming changes to the WPILib library." +
```
prateekma, line 49: Would "...is for upcoming changes to WPILib" be enough? Adding "library" to the end of "WPILib" seems a little redundant.

sciencewhiz:
I think the commit message would be better if it said something like "display warning if using latest version of docs and point to stable".

AustinShalit (approved):

PR #862, opened 2020-10-04
Add networking note

### LINKED ISSUE #372: Encourage Static IP for complex networks over mDNS
For coprocessors and roboRIO, etc.

Daltz333:
Closes #372


sciencewhiz:
I think there needs to be more explanation. Why is roborio and ds recommended to be DHCP, when further down there is a fully dhcp and fully static configuration, but no mixed dhcp and static configuration.

Daltz333:
I think the solution for this may be a more complicated rewrite of the networking article.

PR #866, opened 2020-10-05
Make link-check all a nightly task

### LINKED ISSUE #859: Schedule full linkcheck ci
Currently, the full linkcheck runs on every commit, which is inconsistent (between multiple times per day and once every few days). Instead of running on commit, the full linkcheck should be scheduled to run 1-2 times a week and should automagically open an issue for broken links.

AustinShalit:
Closes #859

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -1,11 +1,7 @@
 name: CI
-on:
-  pull_request:
-    branches:
-      - "*"
-  push:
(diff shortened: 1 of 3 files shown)


Daltz333 (commented):
Ideally I'd like to see a distinguish between link check and linkcheck diff. Perhaps just call it Daily Linkcheck?

Daltz333 (approved):

PR #888, opened 2020-10-11
Fix color formatting

Daltz333:
Apparently --color breaks on windows

image


AustinShalit (commented):
We don't build on Windows CI anyway. Should we just completely revert the windows changes?

AustinShalit (approved):

PR #889, opened 2020-10-11
Move 3rd-party-libraries into zero-to-robot

Daltz333:
This is a better location as 3rd-party-libraries has nothing to do with VS Code


Daltz333:
Probably going to move to step-3, as it more closely relates to using vscode, and this is necessary for the vast majority of teams (use CAN devices)

PR #896, opened 2020-10-18
Remove Redundancy in root TOC

Daltz333:
This PR removes the redundant "WPILib" in many of the root TOC captions. This isn't necessary as the LabVIEW section is quite clearly defined. It doesn't break any existing links and is fairly cheap.

Also, this fixes some line overflow due to translations being quite verbose.

### DIFF
--- a/source/index.rst
+++ b/source/index.rst
@@ -22,7 +22,7 @@ Welcome to the *FIRST*\ |reg| Robotics Competition Control System Documentation!
 
 .. toctree::
    :maxdepth: 1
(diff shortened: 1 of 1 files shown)


Daltz333:
@Kevin-OConnor What are your thoughts on this?

Kevin-OConnor (approved):
Seems fine. I believe NI also calls their library WPILib so the redundancy isnt even buying us that distinction

AustinShalit (approved):

PR #919, opened 2020-11-04
Clear up VS Code installation confusing docs

### LINKED ISSUE #918: Make installer docs more clear on what the select existing download vscode button in the installer does.
People will assume this button is used to select an existing copy of vscode, which is not the case.

Starlight220:
Fixes #918.
@ThadHouse I hope I didn't add anything technically incorrect. 

Thoughts?


Daltz333:
I don't think this is what he meant, I'm pretty sure he's talking about the "Select an Existing Copy" button.

Daltz333:
Closed in favor of #920

PR #928, opened 2020-11-12
Update Offline Preparation

sciencewhiz:
Remove outdated information about downloading VS Code
Clean up 3rd party information since they change was 4 years ago
Move offline installation higher in TOC


Daltz333 (commented):
We should use Java/C++

Daltz333 (approved):

PR #929, opened 2020-11-12
Update beta target dates

AustinShalit:


Kevin-OConnor:
Tasks 2 and 3 still need the year updated. Otherwise looks reasonable

Kevin-OConnor (approved):

PR #948, opened 2020-11-22
Clarify "Change Select Default Simulate Extension Setting"

sciencewhiz:


source/docs/software/vscode-overview/wpilib-commands-vscode.rst
```diff
@@ -16,7 +16,7 @@ To access these commands, press Ctrl+Shift+P to open the Command Palette, then b
 - **WPILib: Change Language Setting** - Change whether the currently open project is C++ or Java.
 - **WPILib: Change Run Commands Except Deploy/Debug in Offline Mode Setting** - Change whether GradleRIO is running in Online Mode for commands other then deploy/debug (will attempt to automatically pull dependencies from online). Defaults to enabled (online mode).
 - **WPILib: Change Run Deploy/Debug Command in Offline Mode Setting** - Change whether GradleRIO is running in Online Mode for deploy/debug (will attempt to automatically pull dependencies from online). Defaults to disabled (offline mode).
-- **WPILib: Change Select Default Simulate Extension Setting** - Change which simulation extensions are enabled by default.
+- **WPILib: Change Select Default Simulate Extension Setting** - Change whether simulate extensions are selected automatically (will only occur if the build defines a single simulation extension)
```
Daltz333, line 19: I don't actually think this is clearer at all. "Selected Automatically" seems more vague than "Enabled by default". I would perhaps keep the original verbiage and just add the parenthesis information

Daltz333 (changes requested):

source/docs/software/vscode-overview/wpilib-commands-vscode.rst
sciencewhiz: I'm open to different wording, but the old wording of saying "which" is wrong because it implies choice that doesn't apply.

sciencewhiz: Maybe Change whether simulation extensions are enabled by default (will only occur if the build defines a single simulation extension)

Daltz333: I like that wording.

Daltz333 (approved):

PR #955, opened 2020-11-24
Add Right-to-Left support

ofekashery:
The beginning of Hebrew support 🥳


AustinShalit:
This looks great! If it helps you at all (not sure if it will) take a look at the PR here: https://github.com/readthedocs/sphinx_rtd_theme/pull/986

Daltz333:
A very pleasant surprise!

ofekashery:
Thanks! I think it's ready.

By the way, it seems like something is broken with the [Hebrew docs](https://docs.wpilib.org/he/latest/) in production - it works for me locally.

Daltz333:
Hebrew isn't enabled 😄

Daltz333:
Something that would be nice, if things like admonitions (warnings/notes), hoverxref (hover tooltip content), and the extra things we have in the docs (code-tabs) could be RTL aligned.

ofekashery:
The admonitions and code-tabs are already RTL aligned.
Do you have an example of a page that uses the ``hoverxref``? I didn't find it.

image
image

Daltz333:
https://docs.wpilib.org/en/latest/docs/software/advanced-controls/introduction/tuning-pid-controller.html

Daltz333:
@Kevin-OConnor what are your thoughts on the CSS styling and changes

ofekashery:
Here is how it looks:
image

AustinShalit:
I am not sure why this would look this way but does the Zayin look backward to you?

image

ofekashery:
Actually, if I focus on that, it looks a bit backward, but it's part of Heebo's typography (We use Heebo as our primary font at _FIRST_ Israel).

AustinShalit:
Ah ok. I don't want to go messing with the typography.

Someone pointed out to me that the `C++` gets converted to `++C` in the image you posted. I think this is not standard?

ofekashery:
It's not expected but unfortunately we don't have much to do with it.
It's generally caused because the characters being written right-to-left, and the ``++`` are locale-free characters, so they are rendered after the ``C``. If there was another English char after the ``++`` it would be okay.

A possible solution is to wrap it in a ``span`` with left-to-right direction (``<span dir="ltr">C++</span>``), but I'm not sure how to do that on the translation side.

This issue seems to also exist in [other sites](https://campus.gov.il/course/course-v1-basmach-pc264/) that write ``C++`` in a Hebrew sentence...

Daltz333:
Ah. I'm gonna spend a bit of time testing locally tomorrow, so expect a review then

source/_static/css/frc-rtl.css
```diff
@@ -0,0 +1,75 @@
...
+
+.rst-versions .rst-current-version {
+  text-align: left;
+}
+.rst-versions .rst-current-version .fa-book, .rst-versions .rst-current-version .icon-book {
+  float: right;
+}
+.rst-other-versions {
+  text-align: right;
+}
```
AustinShalit, line 75:
```suggestion
}

```

AustinShalit (approved):

PR #959, opened 2020-11-29
Link to local api docs in installer included docs

### LINKED ISSUE #728: Point External Link at Different Location in Compile Time
Include content based on configuration
https://www.sphinx-doc.org/en/master/usage/extensions/ifconfig.html

TheTripleV:
closes #728 .

If docs are accessed from the wpilib install location, the links to the Java/C++ Api Docs will go to local versions instead of online versions.

Tested on Mac.


source/cpp-api-docs-redirect.rst
```diff
@@ -0,0 +1,18 @@
+C++ Api Documentation
+======================
```
Daltz333, line 2:
```suggestion
=====================
```

TheTripleV:
rebased off master

Daltz333:
Not sure if it's possible to add a check to confirm that the destination exists (locally), for cases where the RTD docs is manually downloaded

Daltz333:
I have confirmed that this works on Windows

prateekma (changes requested):
Change Api -> API

source/cpp-api-docs-redirect.rst
```diff
@@ -0,0 +1,21 @@
...
+            window.location.href = documentation_url_base + sep + "cpp" + sep + "index.html";
+        } else {
+            window.location.href = "https://first.wpi.edu/FRC/roborio/release/docs/cpp/index.html";
+        }
+    </script>
+
+.. image:: assets/cat-redirect.jpg
+   :alt: Cat Yawning
+
+Oops. Your browser doesn't support Javascript.
```
prateekma, line 19: Javascript -> JavaScript

prateekma (changes requested):

Daltz333 (dismissed):

TheTripleV:
The links to the api docs in `docs/software/advanced-controls/trajectories/ramsete.rst` also need to be fixed.
They link to specific pages in the api docs while this PR currently only supports linking to the root of the api docs.

Solution: 
In ramsete.rst, we can link to the built redirect html file instead of the rst. The desired path (relative to the root of the api docs) can be passed in the query portion of the url and the js can redirect correctly.

sciencewhiz:
I believe there are many more links to the api docs, especially in the command section.

TheTripleV:
I think all links can be automatically taken care of using sphinx's build-finished event.

1. The redirecting file will be updated to allow for redirecting to any page within the api docs.
2. On build-finished, all links to the api docs in frc-docs will be replaced with redirecting ones.

The only downside I see to this solution as opposed to redirecting in the wpilib installer is that this solution requires that javascript is enabled.

So I guess the decision is where to do the find and replace, here or in the wpilib installer?

Daltz333:
It can be done via JavaScript. I don't think we need to support a no-javascript environment, as the website is already severely broken without it.

source/java-api-docs-redirect.rst
```diff
@@ -1,16 +1,13 @@
-Java API Documentation
-======================
+WPILIB Java API Documentation
```
Daltz333, line 1: WPILib

TheTripleV:
Changes:
- All links to the api docs from frc-docs are properly redirected, including links to specific api pages
- The hash is propagated to the redirected url
- Code duplication reduced - the main redirecting js code only exists once now.

Functionality:
All links to the api docs are replaced with a link to a dummy file that conditionally redirects to local/online versions of the api docs. Links to the api docs of form `https://first.wpi.edu/FRC/roborio/release/docs/LANG/PAGE#HASH` are replaced with `api-docs-redirect.html?path=LANG/PAGE&hash=HASH` which is then parsed by the js in `api-docs-redirect.rst`

Note: When `index.rst` is read, `api-docs-redirect.html` has not been built yet, resulting in an error. This is why `cpp-api-docs-redirect.rst` and `java-api-docs-redirect.rst` exist. Users clicking on the api docs links in the nav sidebar will face a double redirect while all other api docs links only result in a single redirect.

TheTripleV:
Because of the double redirect on the sidebar, the browser back doesn't work. Need to implement another workaround. Will do after AoC.

TheTripleV:
Changes:
The dummy redirecting page has been removed. Now, on page load, all api docs links are changed in the DOM. This makes the redirects "instant" and doesn't mangle the browser history.

source/_static/js/api-docs-redirect.js
```diff
@@ -0,0 +1,24 @@
...
+}
+
+document.addEventListener('DOMContentLoaded', function() {
+    "use strict";
+    for (let link of document.links) {
+        if (link.href.startsWith("https://first.wpi.edu/FRC/roborio/release/docs/")) {
+            link.href = resolveApiDocsLink(link.href)
+        }
+    }
+}, false);
```
AustinShalit, line 24:
```suggestion
}, false);

```

Daltz333 (approved):

PR #964, opened 2020-11-30
Change ComandBased Example to Flywheel

### LINKED ISSUE #958: Correct StartEndCommand example
The example for StartEndCommand should be changed because the WPILib drive commands will abort since the motor safety isn't fed. https://docs.wpilib.org/en/latest/docs/software/commandbased/convenience-features.html#startendcommand

jasondaming:
Closes #958

### DIFF
--- a/source/docs/software/commandbased/convenience-features.rst
+++ b/source/docs/software/commandbased/convenience-features.rst
@@ -167,23 +167,23 @@ The ``StartEndCommand`` class (`Java <https://first.wpi.edu/FRC/roborio/release/
   .. code-tab:: java
 
     new StartEndCommand(
-        // Start driving forward at the start of the command
-        () -> m_robotDrive.arcadeDrive(kAutoDriveSpeed, 0),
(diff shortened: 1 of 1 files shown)


prateekma (changes requested):
Although motor safety is not enabled by default on individual motors, we should a note mentioning that this will not work if teams have explicitly enabled it.

jasondaming:
I feel like if you know enough to enable motor safety then you know much more than this example is aimed at.  This is not true of motor safety for drive as that is enabled by default and this could be used with true understanding of why motor safety would prevent it from running.

What causes this command to stop?  I think this is true of a couple of examples in the code but I don't think that this code makes sense completely.  We just have to weigh is it worth adding a decorator to increase the usefulness of the code while making it slightly more complex?  I am seriously tempted to add a .timeout so it actually has a reason to stop, thoughts?

prateekma (approved):

AustinShalit (approved):

PR #968, opened 2020-12-03
Improve the VS Code Debugging Page with Relevant Links

### LINKED ISSUE #355: Improve debugging page
[Debugging a Robot Program](http://frcdocs.wpi.edu/en/latest/docs/software/wpilib-overview/debugging-robot-program.html) talks about setting a breakpoint, but not single stepping.

jasondaming:
Closes #355

### DIFF
--- a/source/docs/software/vscode-overview/debugging-robot-program.rst
+++ b/source/docs/software/vscode-overview/debugging-robot-program.rst
@@ -29,13 +29,13 @@ Click in the left margin of the source code window (to the left of the line numb
 Debugging with Print Statements
 -------------------------------
 
-Another way to debug your program is to use print statements in your code and view them using the RioLog in Visual Studio Code or the Driver Station.  Print statements should be added with care as they are not very efficient especially when used in high quantities.  They should be removed for competition as they can cause loop overruns. 
+Another way to debug your program is to use print statements in your code and view them using the RioLog in Visual Studio Code or the Driver Station.  Print statements should be added with care as they are not very efficient especially when used in high quantities.  They should be removed for competition as they can cause loop overruns.
 
 .. tabs::
 
     .. code-tab:: java
 
-        System.out.print("example");      
+        System.out.print("example");
 
     .. code-tab:: c++
 
--- a/source/docs/software/vscode-overview/debugging-robot-program.rst
+++ b/source/docs/software/vscode-overview/debugging-robot-program.rst
@@ -39,7 +39,7 @@ Another way to debug your program is to use print statements in your code and vi
 
     .. code-tab:: c++
 
-        fmt::print("example");
+        wpi::outs() << "example\n";
 
 Debugging with Network Tables
 -----------------------------
--- a/source/docs/software/vscode-overview/debugging-robot-program.rst
+++ b/source/docs/software/vscode-overview/debugging-robot-program.rst
@@ -5,6 +5,8 @@ Inevitably, a program will not behave in the way we expect it to behave.  When t
 
 A debugger is a tool used to control program flow and monitor variables in order to assist in debugging a program. This section will describe how to set up a debug session for an FRC robot program.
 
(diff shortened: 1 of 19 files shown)


jasondaming:
Some structure questions I have:
- I think this page (https://code.visualstudio.com/docs/editor/editingevolved) is also really good for getting started coding in VS Code but I am unsure where it would go.  I guess in this (https://docs.wpilib.org/en/latest/docs/software/vscode-overview/vscode-basics.html) basics article, but maybe it is more advanced than that?
- Where do you think the OutlineViewer page should go?  Under Software Tools would make sense but it isn't nearly as big as all the others there.  It could go under Network Tables?
- Are there other links or critical debugging info that you think I should include here?

Daltz333:
I would keep the note that a debugger is recommended. We shouldn't recommend prints. An accidental print in Java robot code can cause some pretty severe loop overruns

jasondaming:
are you talking about the note in the blue box?  It felt repetitive with what I had under Debugging with Print Statements.  I have no problem adding it back if that is what you meant as it is an important point to ram home

jasondaming:
Who knows the most about OutlineViewer?  There are some featuers I am unsure what they do and I have never actually used it.  Is it okay to leave the addition of OutlineViewer in this PR?

source/docs/software/vscode-overview/debugging-robot-program.rst
```diff
@@ -5,72 +5,54 @@ Inevitably, a program will not behave in the way we expect it to behave.  When t
...
 
-|Watch Pane|
+        System.out.print("example");
 
-The Watch Pane can be used to monitor specific variables or expressions while debugging. To add an expression, right-click and select Add Expression.
+    .. code-tab:: c++
 
-Call Stack
-^^^^^^^^^^
+        fmt::print("example");
```
prateekma, line 42: This is not going to compile unless teams add `fmtlib` as a dependency. We should recommend `wpi::outs() << "example\n";` instead (from the `wpi/raw_ostream` header).

prateekma (changes requested):

source/docs/software/vscode-overview/debugging-robot-program.rst
jasondaming: blame @calcmogul I got that from him.  Fixed

calcmogul: I didn't intend for it to be in the docs. :stuck_out_tongue:

Daltz333:
The OutlineViewer should be a seperate article.

And yeah, I'd leave the note. People don't like to read things that aren't screaming at them.

Daltz333:
Ideally, I'd like to refactor this page a bit more. Can we use the 
```
:kbd:`Ctrl+Shift+P`
```
directive for keyboard shortcuts and migrate to inline image tags

IE: Replace
```
| my image |
```

with

```
.. image:: myimage.png
   :alt: What my image shows
```

Additionally, things like GUI text elements should be highlighted with the ``guilabel`` directive

```
:guilabel:`my label`
```

jasondaming:
With :kdb: should it be Ctrl or Control?  No spaces?  + or -?  I am going to clean up other similar issues.

Daltz333:
Control is Mac, Ctrl is Windows/Linux

jasondaming:
What about spaces and +/-?  I wasn't familiar with :kdb: should we add that to the style doc?

jasondaming:
What do you think TOC should be?  I like moving 3rd party up to after creating and before deploying (because you need to know to have them before you deploy).

I am not sure what else I would move above debugging both of the importing only needs to be done once so it should either go to the top or where it is and using test mode I think is in the wrong section.  It should probably go to the driverstation section or something.  It really has nothing to do with VS Code Overview

Daltz333 (approved):

PR #987, opened 2020-12-08
Driver Station undocumented feature - override robot IP address

### LINKED ISSUE #986: Driver Station undocumented feature - override robot IP address
https://www.chiefdelphi.com/t/announcing-gloworm-an-inexpensive-and-open-source-vision-module/386370/72

Can we include that the DS can have the team number overridden as an IP address to point directly at a robot to control?
3rd party equipment relies on this function, its weird that it's not documented as something like an "Advanced Feature", or a throwaway sentence "In Non Competition Use ONLY, the Team Number can be overridden by the complete Robot IP address"  

See Ethernet section of: 

https://pdocs.kauailabs.com/vmx-pi/software/vmx-pi-for-frc-2020-robot-programming/vmx-pi-for-frc-documentation/controlling-a-vmx-pi-frc-robot-using-the-frc-driver-station/

jasondaming:
Closes #986


source/docs/software/driverstation/driver-station.rst
```diff
@@ -19,10 +19,12 @@ Setting Up the Driver Station
 
 .. image:: images/driver-station/ds-setup.png
 
-The DS must be set to your team number in order to connect to your robot. In order to do this click the Setup tab then enter your team number in the team number box. Press return or click outside the box for the setting to take effect.
+The DS should be set to your team number in order to connect to your robot. In order to do this click the Setup tab then enter your team number in the team number box. Press return or click outside the box for the setting to take effect.
 
 PCs will typically have the correct network settings for the DS to connect to the robot already, but if not, make sure your Network adapter is set to DHCP.
 
+.. note:: An IP address can also be entered in the :guilabel:`Team Number` box for use in rare situations.  This is not recommended for normal use!
```
Kevin-OConnor, line 26: Maybe instead of just saying "rare situations" here we elaborate a little more on what situations? They may not end up being all that rare. Situations I can think of: non-RoboRIO targets, troubleshooting, possibly simulation.

Kevin-OConnor (changes requested):
We should probably also document some of the other un-documented features of this box which may make sense in this same PR. Particularly, clicking on the arrow with a team number in the box should show you detected roboRIO's with that team number, shift-clicking the arrow should show you all roboRIOs detected on the network (both of these rely on mDNS).

sciencewhiz:
> We should probably also document some of the other un-documented features of this box which may make sense in this same PR. Particularly, clicking on the arrow with a team number in the box should show you detected roboRIO's with that team number, shift-clicking the arrow should show you all roboRIOs detected on the network (both of these rely on mDNS).

Elsewhere in the document it mentions mdns and clicking the arrow. That might be a better place to add a mention of the IP address.

AustinShalit (approved):

source/docs/software/driverstation/driver-station.rst
jasondaming: If there is no team number entered in the box (either when XXYY or when cleared) what will the dropdown show?  

I would in general argue that showing all the robots is a good idea for the single click and might be more intuitive.  What is the downside to always showing all of them?  Do we expect the robot to be on networks with that many roboRIOs?

jasondaming:
I agree moving everything down in the setup tab makes it much less important to heavily caution against it too!

source/docs/software/driverstation/driver-station.rst
Daltz333: How about
```
.. note:: An IP address can also be entered in the :guilabel:`Team Number` box as an alternative. This is not recommended for competition use!
```

AustinShalit (dismissed):
I think that this looks good now. @Kevin-OConnor?

Kevin-OConnor (dismissed):

Daltz333 (approved):

PR #1010, opened 2020-12-14
Note to the Post Install Section about the Tools Shortcuts Folder

### LINKED ISSUE #1006: Shuffleboard shortcut not showing up
I just installed the 2021 WPILib VsCode beta version. The installation went fine but the Shuffleboard shortcut is missing on my desktop. Actually I just have 2 shortcuts: 2021 WPILib VS Code and 2021 WPILib Documentation.

jasondaming:
Closes #1006

### DIFF
--- a/source/docs/zero-to-robot/step-2/wpilib-setup.rst
+++ b/source/docs/zero-to-robot/step-2/wpilib-setup.rst
@@ -108,7 +108,7 @@ Some operating systems require some final action to complete installation.
 
       .. image:: images/wpilib-setup/linux-enable-launching.png
 
-.. note:: The shortcuts are now placed on the desktop in a folder called "YYYY WPILib Tools" where ``YYYY`` is the year.
+.. note:: Installing desktop tools and rebooting will create a folder on the desktop called YYYY WPILib Tools, where YYYY is the current year. Desktop tool shortcuts are not available on Linux and MacOS.
 
 What is Installed?
 ------------------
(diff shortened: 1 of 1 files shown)


sciencewhiz (changes requested):
On Windows 10, you actually need to log out or reboot for the tools folder to appear. That should go in a new tab for Windows. 

Currently, macOS doesn't install any shortcuts and Linux only installs vscode, so the general note about the folder only applies to Windows.

jasondaming:
Okay so I am thinking instead of a note maybe I should do a tab structure with tabs for Mac, Windows 7, Windows 10 and Linux if we know what that does?

Daltz333:
I think a tab is excessive and isn't as exclamatory as a note.

jasondaming:
Well I don't think 4 different notes is a good idea so I am open to suggestions.

Daltz333:
This should be fine

```Installing desktop tools will create a folder on the desktop called YYYY WPILib Tools, where YYYY is the current year. Desktop tool shortcuts are not available on Linux and MacOS.```

sciencewhiz:
That note does not resolve the original issue because it doesn't tell them to reboot to see the icon. 

We already have a tab with different post install actions for Mac and Linux, I'm not sure why there's resistance for adding another tab for windows.

Daltz333:
Not a ton of resistance. Admonitions generally convey a bit more importance, while I think tabs can get missed.

jasondaming:
What about: "Installing desktop tools and rebooting will create a folder on the desktop called YYYY WPILib Tools, where YYYY is the current year. Desktop tool shortcuts are not available on Linux and MacOS."  I don't think the chance that win7 users reboot if they didn't see the folder is a problem as there are comparably fewer of them and rebooting after install is never a bad idea IMO.

sciencewhiz:
That text sounds fine to me

sciencewhiz (approved):

Daltz333 (approved):

AustinShalit (approved):

PR #1016, opened 2020-12-18
Robot Characterization: Track Width

### LINKED ISSUE #506: Robot Characterization: Track Width
There isn't any documentation of the track width characterization

jasondaming:
Closes #506

### DIFF
--- a/source/docs/software/wpilib-tools/robot-characterization/characterization-routine.rst
+++ b/source/docs/software/wpilib-tools/robot-characterization/characterization-routine.rst
@@ -56,11 +56,11 @@ Follow the instructions in the pop-up windows after pressing each test button.
 Trackwidth
 ^^^^^^^^^^
 
-The Trackwidth test will spin your robot in an attempt to measure the distance between your wheels.
+The trackwidth test will spin your robot to determine an empirical trackwidth. It compares how far the wheel encoders drove against the reported rotation from the gyroscope.
 
 .. note:: If your robot is having trouble turning during the Trackwidth test you should increase the :guilabel:`Rotation Wheel voltage (V):` value and run the test again.
 
-.. note:: Trackwidth values for high friction wheels (like pneumatic tires) may be measured incorrectly and a measured value should be used.
+.. note:: Trackwidth values for high friction wheels (like pneumatic tires) may be significantly different from the measured value.
 
 The entire routine should look something like this:
 
--- a/source/docs/software/wpilib-tools/robot-characterization/characterization-routine.rst
+++ b/source/docs/software/wpilib-tools/robot-characterization/characterization-routine.rst
@@ -56,11 +56,11 @@ Follow the instructions in the pop-up windows after pressing each test button.
 Trackwidth
 ^^^^^^^^^^
 
-The trackwidth test will spin your robot to determine an empirical trackwidth. It compares how far the wheel encoders drove against the reported rotation from the gyroscope.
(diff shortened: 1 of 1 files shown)


Daltz333:
This seems good to me. @calcmogul ?

source/docs/software/wpilib-tools/robot-characterization/characterization-routine.rst
```diff
@@ -53,6 +53,15 @@ Follow the instructions in the pop-up windows after pressing each test button.
...
    :alt: Running the various tests
 
+Trackwidth
+^^^^^^^^^^
+
+The Trackwidth test will spin your robot in an attempt to measure the distance between your wheels.
+
+.. note:: If your robot is having trouble turning during the Trackwidth test you should increase the :guilabel:`Rotation Wheel voltage (V):` value and run the test again.
+
+.. note:: Trackwidth values for high friction wheels (like pneumatic tires) may be measured incorrectly and a measured value should be used.
```
prateekma, line 63: This is misleading. The track width value that frc-char will calculate for drivetrains with high-friction wheels *is* the value that should be used everywhere in code -- it just won't match with the physical reality.

The physical measurement should *not* be used because it won't accurately represent the angular dynamics.

prateekma (changes requested):

source/docs/software/wpilib-tools/robot-characterization/characterization-routine.rst
jasondaming: That isn't what I understood from @calcmogul.   He also said don't do it with pneumatics but that is probably for a different note...maybe?

calcmogul: What Prateek said. And yeah, avoid pneumatic wheels because the encoders aren't as accurate, even with a correct empirical trackwidth.

jasondaming: So what do you think this note should say something like:
"Trackwidth values for high friction wheels (like pneumatic tires) may be significantly different from the measured value."

calcmogul, line 63: Might not be good as is, but here's a start:
```suggestion
The trackwidth test will spin your robot to determine an empirical trackwidth. It compares how far the wheel encoders drove against the reported rotation from the gyroscope. If the encoder distances are small relative to the change in heading, the empirical trackwidth will be smaller than the real value to reflect wheel scrub. A common cause for this would be high-friction wheels (e.g., pneumatic tires).

This empirical value should be used everywhere in code instead of the real value measured with a tape measure.

.. note:: If your robot is having trouble turning during the Trackwidth test you should increase the :guilabel:`Rotation Wheel voltage (V):` value and run the test again.
```

jasondaming: From your first paragraph it sounds like the trackwidth would be smaller.  In practice I observed a trackwidth >2X actual.

jasondaming:
@prateekma / @calcmogul: I made changes.  Because of the uncertainty when we were talking I didn't include a reason that it was different just that it could be.  What do you think?

prateekma:
I think it is important that we let teams know to use the value from characterization in their code even though it doesn't match physical reality. Just a note saying "it won't match" is vague and doesn't help users who are confused about what to use.

prateekma (approved):

Daltz333 (approved):

PR #1027, opened 2020-12-22
Add linting following styleguide

TheTripleV:
Adds a linter package that extends doc8 to add tests for conformity to the stricter frc-docs styleguide.

Checks included:
- CheckFilenames
  - make sure filenames are valid
  - make sure directory names are valid
- CheckTermSpelling
  - make sure terms are spelled correctly
- CheckHeadingUnderscores
  - make sure headings are used in the correct order
  - make sure heading underline lengths are correct
- CheckDoubleNewlines
  - make sure there are no double newlines
- CheckTrademarks
  - make sure the first instance of some terms are trademarked
- CheckFIRSTItalicized
  - make sure FIRST is italicized
- CheckBlankLinkAfterHeader
  - make sure headers are followed by a blank line
- CheckBlankLinkAfterDirective
  - make sure directives without options are followed by a blank line
- CheckIndentation
  - make sure indents follow styleguide (~todo: add list support~ done)
- CheckInteriorSpaces
  - make sure sentences only have 1 space separating them
- CheckImageAlt
  - make sure all images have alt text
- CheckSingleLineParagraph
  - make sure all paragraphs are on a single line

### Testing
I wrote all the boilerplate required for testing.
Tests are welcome.

### Future
This linter finds many issues in frc-docs. They should go in a separate PR once this one is finalized to reduce clutter.


TheTripleV:
The styleguide needs to be excluded as it explicitly does not follow the styleguide.

TheTripleV:
linter reports 3482 errors.

Daltz333:
For right now, can it be added to run as diffs that way we can slowly migrate as things get updated.

TheTripleV:
I think that should do it?

sciencewhiz:
Does this exclude code?

TheTripleV:
doc8 uses docutils directly so deleting code files shouldn't matter. Just in case, I changed it to only delete rst files.

sciencewhiz:
I meant, for example, if a there's two consecutive new lines within a .. code:: directive, would this flag it?

TheTripleV:
code directives are excluded from all checks except for the two consecutive lines test.

I had accounted for code-blocks and code-tabs but forgot about regular old code. I fixed that now.

sciencewhiz:
Is there a way to disable a specific linter for a block?

TheTripleV:
If you're asking if there are pre-processor flags of sort that allow users to disable linting for blocks, then no.

I think that any excluded blocks should be hardcoded into the checks.

TheTripleV:
Removed the linter source and updated ci to use ohnoyoudidnt

Daltz333 (approved):

PR #1035, opened 2020-12-27
Update remote literal includes to use tags

sciencewhiz:
Using master will cause broken docs if something changes in allwpilib
master.
Use the latest release tag (2020.3.2) when applicable, or the latest beta
tag (2021.1.1-beta-4) if needed.
Fix a few line number problems.


prateekma (changes requested):
Trajectory tutorial should use Beta 4 because there were some changes we made to resetting odometry. See: https://github.com/wpilibsuite/frc-docs/commit/6872609273e7864cfeba011182c81b60ab3b4239

Edit: Never mind, the appropriate changes are using the correct version.

Daltz333 (approved):

PR #1075, opened 2021-01-09
Add literal checker

Daltz333:
Adds a simple diff checker to verify that all literal includes are up-to-date. It goes for the dummy simple approach of "if anything in the file changes, update". 

TODO: Fix current literal includes to latest release


.github/workflows/CI.yaml
```diff
@@ -97,3 +97,23 @@ jobs:
...
+    - name: Install Dependencies
+      run: |
+        pip install -r source/requirements.txt
+    - name: Remove unchanged files
+      run: |
+        ( git diff --name-only origin/master ; git ls-files ) | sort | uniq -u | grep "^.*\.rst$" | xargs rm
+    - name: Check remote literal includes
+      run: |
+        make includecheck
+
```
AustinShalit, line 119:
```suggestion

```

Daltz333:
@calcmogul Mind taking a brief look at the python?

scripts/verify-literal-check.py
```diff
@@ -0,0 +1,100 @@
+import os
+import argparse
+import re
+
+import urllib.request
```
calcmogul, line 5: These should be sorted.

scripts/verify-literal-check.py
```diff
@@ -0,0 +1,100 @@
...
+# # URL_REGEX FROM:
+# https://mathiasbynens.be/demo/url-regex
+
+
+URL_REGEX = re.compile("((https?):((//)|(\\\\))+([\w\d:#@%/;$()~_?\+-=\\\.&](#!)?)*)", re.DOTALL)
+
+# Grab RemoteLiteralIncludes from .rst resources
+# Return the URLs in the RLIs
+def grab_rli_urls(path: str):
+
```
calcmogul, line 16: Remove the blank line.

scripts/verify-literal-check.py
```diff
@@ -0,0 +1,100 @@
...
+def verify_diff(rli_urls: list, target_branch: str):
+    err = False
+    for url in rli_urls:
+        source_url = rli_urls[url]
+        split_url = source_url.split("/")
+
+        if "raw" in split_url:
+            split_url.remove("raw")
+        
+        split_url[5] = target_branch
```
calcmogul, line 49: Why 5? Make this a constant or document it.

calcmogul (changes requested):

TheTripleV:
Currently as in CI.yaml, this won't properly work.
1. Since this only runs on changed files, rli's in unchanged frc-docs files who's source has been updated in allwpilib will be missed.
2. PRs unrelated to rli's will fail if a change in allwpilib is made.

I think this tool should be scheduled daily like full-link-check is. On fail, it should open in issue in this repo.

Daltz333:
Automatically opening an issue is a bit of a pain, I'd rather not interact with Google's API more than I have to. I agree, I plan for it to run daily.

In regards to #2. This is fairly complex problem, the proper method would be to store a couple of the current "lines" and compare them to see if the "lines" change. This sounds like non-trivial and I'd just be easier for contributors to just update the link. PRs are cheap

scripts/verify-literal-check.py
```diff
@@ -0,0 +1,100 @@
...
+# Grab RemoteLiteralIncludes from .rst resources
+# Return the URLs in the RLIs
+def grab_rli_urls(path: str):
+
+    # retrieve list of rst documents
+    documents = {}
+    for root, subdirs, files in os.walk(os.path.abspath(path)):
+        for document in files:
+            if ".rst" in document:
+                documents[os.path.join(root, document)] = os.path.join(root, document)
```
TheTripleV, line 22: They key and value are the same so it can just be a list/set/iterable?
So, something like this would be simpler.
```suggestion
from pathlib import Path # before
    documents = Path(path).glob("**/*.rst")
```

.github/workflows/CI.yaml
Daltz333: Can you resubmit this review? Github is giving me barf here after I force-pushed.

Daltz333:
If anyone would like to help update the RLI in the documentation, that would be tremendously welcome!

Daltz333:
Superceded by #1351

PR #1114, opened 2021-01-19
Encoder getDistance Issues

### LINKED ISSUE #1113: Encoder Documentation Issues
https://docs.wpilib.org/en/stable/docs/software/sensors/encoders-software.html#distance
image

=====

https://docs.wpilib.org/en/stable/docs/software/sensors/encoders-software.html#driving-to-a-distance

Calls `getDistance` not `getDistance()`

jasondaming:
Closes #1113

### DIFF
--- a/source/docs/software/sensors/encoders-software.rst
+++ b/source/docs/software/sensors/encoders-software.rst
@@ -127,13 +127,13 @@ Users can obtain the total distance traveled by the encoder with the :code:`getD
 
     .. code-tab:: java
 
-        // Configures an encoder to return a distance of 4 for every 256 pulses
-        encoder.setDistancePerPulse(4./256.);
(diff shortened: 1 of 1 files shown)


Daltz333 (approved):

sciencewhiz:
It might be clearer if the setting distance per pulse was in the GetDistance section, since setting the distance per pulse is a prerequisite for using GetDistance, rather then it hiding in the general Encoder section.

jasondaming:
To me it makes sense as is, setDistancePerPulse is part of the general encoder setup.  Creating the encoder is also a prerequisite for using it.  I don't think we should worry about the prereqs as long as they are explained above.  Get is a separate action that happens repeatedly rather than just at setup.

PR #1122, opened 2021-01-20
Invert GetY() of controllers where necessary as "up" is -Y not +Y

Modelmat:
This does not fix any of the `remoteliteralinclude` examples yet (these will need to be fixed by updating the examples within allwpilib and then bumping the include versions later).

This should probably be sanity checked by people with real robots (I can't do that).

### DIFF
--- a/source/docs/software/actuators/wpi-drive-classes.rst
+++ b/source/docs/software/actuators/wpi-drive-classes.rst
@@ -196,26 +196,26 @@ Like Arcade Drive, the Curvature Drive mode is used to control the drivetrain us
 
         public void teleopPeriodic() {
             // Tank drive with a given left and right rates
-            myDrive.tankDrive(leftStick.getY(), rightStick.getY());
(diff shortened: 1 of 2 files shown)


Daltz333:
Needs sanity checked.

AustinShalit (approved):
This looks correct to me.

Daltz333:
Did you/should this get checked on a real robot @AustinShalit

Modelmat:
The force push was just to add an explaining comment.

This has been testing in a simulation on Linux with an XboxController, but I would definitely like confirmation on an IRL robot (though multiple others have "confirmed" that this should be the behaviour).

Daltz333 (approved):
I verified this when my teams robot.

Modelmat:
I'm unsure if any of the remotelitetalinclides have been updated. Currently starting exam block so I'll add it to a to-do list and get back in ~2 weeks unless someone fixes it first.

(The examples affected will exist in a WPILIB PR).

PR #1126, opened 2021-01-21
Fix photo not found when clicked on

### LINKED ISSUE #1125: Photo not found when clicked on
The overview picture (first one [here](https://docs.wpilib.org/en/stable/docs/controls-overviews/control-system-hardware.html)) shows up when scrolling, yet gives a page not found when clicked on.

jasondaming:
Closes #1125

### DIFF
--- a/source/docs/controls-overviews/control-system-hardware.rst
+++ b/source/docs/controls-overviews/control-system-hardware.rst
@@ -11,7 +11,7 @@ Overview of Control System
 --------------------------
 
 .. figure:: images/frc-control-system-layout.svg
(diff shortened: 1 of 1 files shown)


Daltz333:
Does width need to be added?

jasondaming:
Yes.  If width is not added the image is not clickable.  I know you dislike having width there but IMO the benefit of being able to click on the image and zoom in ETC outweighs the cost

Daltz333:
That's fine then.

Daltz333 (approved):

sciencewhiz (approved):

PR #1140, opened 2021-01-24
Add detail that two nuts for romi is attached in the battery compartment

### LINKED ISSUE #1136: Add detail that two nuts for romi is attached in the battery compartment
Two nuts have to be attached to the RPi standoffs via the battery compartment. Just slot the nuts in and twist. This wasn't obvious at first and I spent a good 5 minutes looking for a way to attach them.

jasondaming:
Closes #1136


Daltz333 (approved):

Daltz333 (changes requested):
Actually, can you compress it a bit? Not sure why the size increased by 200KB for a tiny change

Daltz333 (approved):

PR #604, opened 2020-04-05
Remove software tools section

ItayZiv:
Remove software tools and split up the tools into the most fitting location. Not 100% sure about the locations...

fixes https://github.com/wpilibsuite/frc-docs/issues/505

PR #1141, opened 2021-01-24
Move Sim Tutorial to Simulation Directory

Daltz333:
Not sure if we want to do this mid-season, but lots of people are finding the flow confusion (jumping around simulation and sim tutorial). I think this is more intuitive.


jasondaming:
As I said in discord I think the PR #1139 that was pushed tonight will help people a ton.  I really think if this is the way we want to go we should push this with a breakup / move of the trajectory tutorial too.  But I am not still sold on breaking up the tutorials but I see both sides.

Daltz333:
This conversation closely relates to #604

jasondaming (approved):

Issue #492, opened 2020-01-20
Relay docs


sciencewhiz:
There isn't any software documentation for relays.

Starlight220:
How many teams still use relays?

AustinShalit:
In 2019, 187 teams instantiated 270 relays.

They are a supported feature and should be added to the documentation

Starlight220:
I agree with the fact that they need documentation, and more docs could increase the amount of users. 
However, what are FRC uses of relays? What relay-based FRC products exist?

AustinShalit:
A handful of teams use relays to power their compressor rather than the PCM so they can use an auto resetting breaker. Relays are also an easy way to turn an LED light ring on and off.

ItayZiv:
Relays big use are in various custom circuits

sciencewhiz:
Relays can also be used for low current motors.

PR #1144, opened 2021-01-24
Add relay description to actuator overview

### LINKED ISSUE #626: Add relay description to actuator overview
https://github.com/wpilibsuite/frc-docs/blob/master/source/docs/software/actuators/overview.rst

This page contains a brief description of motor controllers and pneumatics but nothing for relays. We should add something short and sweet describing their use.

jasondaming:
Closes #626

### DIFF
--- a/source/docs/software/actuators/overview.rst
+++ b/source/docs/software/actuators/overview.rst
@@ -7,6 +7,7 @@ This section discusses the control of motors and pneumatics through motor contro
 
 Motor Controllers
 -----------------
+
 A motor controller is responsible on your robot for making motors move. For brushed DC motors such as CIMs or 775s, the motor controller regulates the voltage that the motor receives, much like a light bulb. For brushless motor controllers such as the Spark MAX, the controller regulates the power delivered to each "phase" of the motor.
 
 .. note:: Another name for a motor controller is a speed controller.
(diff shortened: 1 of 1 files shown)


source/docs/software/actuators/overview.rst
```diff
@@ -36,13 +38,20 @@ Motor controllers come in lots of shapes, sizes and feature sets. This is the fu
...
 
 FRC Legal Pneumatics controllers
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
 - Pneumatics Control Module (P/N: am-2858, 217-4243)
 
+Relays
+------
+
+A relay controls power to a motor or custom electronics in an On/Off fashion.
```
Daltz333, line 52: How do you control a relay in code? Perhaps this deserves an article?

jasondaming: I think the relay example covers it well.  I don't think it warrants its own page.  Even if it did it wouldn't be this page right?

https://github.com/wpilibsuite/allwpilib/tree/master/wpilibjExamples/src/main/java/edu/wpi/first/wpilibj/examples/relay

sciencewhiz: I do think it deserves an article, but it definitely doesn't belong here (it would close #492). Specifically, the difference between kOn and kForward often confuses people. It's not intuitive that kOn makes a motor not turn.

sciencewhiz (approved):

PR #1157, opened 2021-01-26
Add macOS Mojave known issue and info on min version

prateekma:


Daltz333:
Should we specify this in the installer docs too?

Daltz333 (approved):

PR #1160, opened 2021-01-27
Add console commands for running Linux installer

calcmogul:


prateekma:
We should probably make the commands version agnostic so that we don't have to update them every release.

Daltz333 (approved):

PR #1162, opened 2021-01-28
Add information on cloning

### LINKED ISSUE #1161: Add information or example about git clone
https://docs.wpilib.org/en/stable/docs/software/basic-programming/git-getting-started.html has lots of good introductory information about starting to use git. I think a key use case is missing though in using git clone. It has how to start a git locally and then push it remote, but it does not have a walk through (or links to examples) of pulling a remote repository to local. 

I'm not imagining something very long or in-depth. Just something like the cloning an existing repository here https://firstmncsa.org/2018/09/07/vs-code-with-github/

Daltz333:
Closes #1161


AustinShalit:
Do you think cloning should be the first topic or later in the progression?

AustinShalit (approved):

PR #1163, opened 2021-01-29
Remove note about scaling on Retina displays

prateekma:
This was fixed in a recent allwpilib commit (https://github.com/wpilibsuite/allwpilib/pull/3135)


prateekma:
@sciencewhiz This should be good to go now.

Daltz333:
Should we wait until a release is cut?

sciencewhiz:
I think so, since we don't have separate latest and stable right now.

Daltz333 (dismissed):

sciencewhiz:
Thinking about it some more, I think I would prefer to add something to say to update to 2021.2.2 or later, rather then deleting it.

Daltz333 (approved):

PR #1176, opened 2021-01-31
Simulation - Add Keyboard Joystick Instructions

### LINKED ISSUE #1154: Simulation - Add Keyboard Joystick Instructions
See https://github.com/wpilibsuite/allwpilib/discussions/3129 for motivation.

Should include how to configure and how to use them.

jasondaming:
Closes #1154


prateekma (changes requested):
I think we should also document what the "Key Rate", "Decay Rate", and "Max Absolute Value" settings do.

jasondaming:
I thought about that but explicitly decided against it.  This page is getting super long and is a grab bag for a lot of things simulation related.  I think if we decided to split up this page and have a specific "joysticks" page then I wouldn't mind talking about all the things above.  What does everyone else think?

https://docs.wpilib.org/en/latest/docs/software/wpilib-tools/robot-simulation/simulation-gui.html

Daltz333:
Simulation can likely be structured similar to the Glass documentation with the sections for widgets and the such.

jasondaming:
For that matter is there value in continuing to keep Sim/Glass separate?  From a user standpoint they really aren't.  I think there will we a couple parts that diverge but I think the vast majority of the articles apply to both and would be duplication otherwise.

Daltz333:
@PeterJohnson what are your thoughts

Daltz333:
Marking as needs design decision, as I talk with Peter about merging glass and simulation docs.

Daltz333 (approved):

Daltz333:
These docs are only applicable for simulation anyways. Glass has no keyboard control.

PR #1198, opened 2021-02-07
Add missing motor declarations

### LINKED ISSUE #1100: m_leftMotor and m_rightMotor used in Drivetrain Sim Tutorial example code, but not referenced anywhere
The java example on [this page](https://docs.wpilib.org/en/stable/docs/software/examples-tutorials/drivesim-tutorial/updating-drivetrain-model.html) references `m_leftMotor` and `m_rightMotor` but the reader is never told what these are.

Daltz333:
Close #1100

### DIFF
--- a/source/docs/software/examples-tutorials/drivesim-tutorial/updating-drivetrain-model.rst
+++ b/source/docs/software/examples-tutorials/drivesim-tutorial/updating-drivetrain-model.rst
@@ -13,6 +13,9 @@ There are three main steps to updating the model:
 .. tabs::
    .. code-tab:: java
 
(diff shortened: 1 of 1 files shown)


prateekma (commented):
I think we should use PWMSparkMax because that's what's in the KoP.

Daltz333:
Done.

prateekma (approved):

sciencewhiz (approved):

sciencewhiz:
Also since a Neo is used for the simulated physics, so it's the only speed controller that would work ;-)

PR #1202, opened 2021-02-07
Clarify Encoder Class

### LINKED ISSUE #1191: Double Clarify that the Encoder Class is just for Encoders plugged into the RIO
https://docs.wpilib.org/en/latest/docs/software/sensors/encoders-software.html

We say what an encoder is and there are many times of encoders now that don't get plugged into the RIO so many are getting confused using this Encoder class because they have an encoder...just not the right type.  They should realize that when they try to construct and the parameter requires a digital line but lets make it extra super clear.

Daltz333:
Close #1191

### DIFF
--- a/source/docs/software/sensors/encoders-software.rst
+++ b/source/docs/software/sensors/encoders-software.rst
@@ -18,6 +18,8 @@ The Encoder class
 
 WPILib provides support for encoders through the :code:`Encoder` class (`Java <https://first.wpi.edu/wpilib/allwpilib/docs/release/java/edu/wpi/first/wpilibj/Encoder.html>`__, `C++ <https://first.wpi.edu/wpilib/allwpilib/docs/release/cpp/classfrc_1_1Encoder.html>`__).  This class provides a simple API for configuring and reading data from encoders.
 
+.. important:: The encoder class is only used for encoders that are plugged directly in the roboRIO! Please reference the appropriate vendors documentation for using third-party encoders.
+
 Initializing an encoder
 ^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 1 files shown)


source/docs/software/sensors/encoders-software.rst
```diff
@@ -18,6 +18,8 @@ The Encoder class
 
 WPILib provides support for encoders through the :code:`Encoder` class (`Java <https://first.wpi.edu/wpilib/allwpilib/docs/release/java/edu/wpi/first/wpilibj/Encoder.html>`__, `C++ <https://first.wpi.edu/wpilib/allwpilib/docs/release/cpp/classfrc_1_1Encoder.html>`__).  This class provides a simple API for configuring and reading data from encoders.
 
+.. important:: The encoder class is only used for encoders that are plugged directly in the roboRIO! Please reference the appropriate vendors documentation for using third-party encoders.
```
prateekma, line 21:
```suggestion
.. important:: The ``Encoder`` class is only used for encoders that are plugged directly into the roboRIO! Please reference the appropriate vendors' documentation for using third-party encoders.
```

prateekma (approved):

source/docs/software/sensors/encoders-software.rst
```diff
@@ -18,6 +18,8 @@ The Encoder class
 
 WPILib provides support for encoders through the :code:`Encoder` class (`Java <https://first.wpi.edu/wpilib/allwpilib/docs/release/java/edu/wpi/first/wpilibj/Encoder.html>`__, `C++ <https://first.wpi.edu/wpilib/allwpilib/docs/release/cpp/classfrc_1_1Encoder.html>`__).  This class provides a simple API for configuring and reading data from encoders.
 
+.. important:: The ``Encoder`` class is only used for encoders that are plugged directly into the roboRIO! Please reference the appropriate vendors' documentation for using third-party encoders.
```
sciencewhiz, line 21: I think third-party encoders should be clarified. It's how the encoder is connected, not the encoder itself. A Grayhill Encoder attached to the RoboRIO would use the Encoder class, but the same encoder attached to a CANCoder would use the CANCoder class and the same encoder attached to a SparkMax would use CANEncoder.

Daltz333: Clarified.

sciencewhiz (approved):

PR #1212, opened 2021-02-10
Fix formatting issue. Add Raspberry PI jni lib.

agasser:
Fix improperly formatted text in example.
Add Raspberry PI JNI library. The comment says it is adding all supported platforms. Additionally Raspberry Pi is a good use case for a standalone NetworkTables client.

### DIFF
--- a/source/docs/software/networktables/client-side-program.rst
+++ b/source/docs/software/networktables/client-side-program.rst
@@ -12,7 +12,7 @@ A basic client program looks like the following example.
    .. code-tab:: java
 
       package networktablesdesktopclient;
-      
+
       import edu.wpi.first.networktables.NetworkTable;
       import edu.wpi.first.networktables.NetworkTableEntry;
       import edu.wpi.first.networktables.NetworkTableInstance;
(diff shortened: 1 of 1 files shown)


source/docs/software/networktables/client-side-program.rst
```diff
@@ -10,7 +10,9 @@ A basic client program looks like the following example.
 .. tabs::
 
    .. code-tab:: java
+
       package networktablesdesktopclient;
+
```
sciencewhiz, line 15: Fix trailing white space
```suggestion

```

sciencewhiz (changes requested):

source/docs/software/networktables/client-side-program.rst
agasser: Fixed

Daltz333:
Should the comment get removed?

agasser:
> Should the comment get removed?

That's a good question. It is helpful to have all of the environments in the example, otherwise another section should be added to explain. In my use case I was using a Raspberry Pi and I had to solve it myself when I got an error attempting to run it.

Daltz333 (approved):

PR #1222, opened 2021-02-11
Add Installation Note

### LINKED ISSUE #1221: Emphasize that we install a separate VS Code than what the user has

Daltz333:
Close #1221


prateekma (approved):

sciencewhiz:
I think we should also mention that it installs separate from other years wpilib installs.

sciencewhiz (approved):

PR #1273, opened 2021-03-20
Recommend SI Units for frc-char analysis

### LINKED ISSUE #1270: Add a note for analyzing data to set Units to meters
"It's recommended to set units to meters, as WPILib uses metric as its preferred unit."

Verbiage could be better. Can be placed in the analyzing section of trajectory tutorial

Daltz333:
Close #1270

### DIFF
--- a/source/docs/software/wpilib-tools/robot-characterization/analyzing-data.rst
+++ b/source/docs/software/wpilib-tools/robot-characterization/analyzing-data.rst
@@ -1,6 +1,8 @@
 Analyzing Data
 ==============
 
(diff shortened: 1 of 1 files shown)


Oblarg (approved):
LGTM

sciencewhiz:
The original request on CD I thought was talking about the ramsete tuning parameters and how the magic numbers given only apply with SI units. While this is good, it doesn't really address that.

sciencewhiz:
Never mind, what I thought of already says meters.

Issue #1147, opened 2021-01-25
Add warning about AddCommands() footgun


Modelmat:
See https://github.com/wpilibsuite/allwpilib/issues/3109

jasondaming:
How / where do we want to document this?

Daltz333:
The solution should include a warning with an example of what's bad and how to work around it. Unfortunately, that's outside of my skills, so if anybody familiar with C++ wants to tackle. I am taking the "Good First Issue" tag away due to some required C++ semantics knowledge.

PR #1274, opened 2021-03-20
Add warning for C++ NPE

Oblarg:
Addresses https://github.com/wpilibsuite/frc-docs/issues/1147.

### DIFF
--- a/source/docs/software/commandbased/convenience-features.rst
+++ b/source/docs/software/commandbased/convenience-features.rst
@@ -26,6 +26,8 @@ While method references work well for passing a subroutine that has already been
 Lambda Expressions (C++)
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


Daltz333:
It was requested for an example of bad and how to fix it

Oblarg:
I'm not quite sure how to easily structure a code example for this without it being fairly long - the setup is kind of obscure.  I can't just plop an excerpt from a command constructor in there - that's about as confusing as the description itself.

Daltz333:
@Modelmat

Modelmat:
I was thinking either a <details> element within the page (or just adding the actual example to a section at the bottom of the page).

Or perhaps just linking the the https://github.com/wpilibsuite/allwpilib/pull/3111 PR which fixed it within an example.

Daltz333 (approved):

PR #1276, opened 2021-03-22
Add a common terminology glossary page

### LINKED ISSUE #198: Add a common terminology glossary page
There should be a description of deprecated somewhere. Most high school students don't know what it means, and when they see warnings they think their code is broken.

jasondaming:
Closes #198 

I need more ideas of things that need to be included!

### DIFF
--- a/source/docs/software/frc-glossary.rst
+++ b/source/docs/software/frc-glossary.rst
@@ -20,11 +20,11 @@ FRC Glossary
    C++
       One of the three officially supported programming languages.
 
-   depricated
+   deprecated
       Software that will be maintained for at least 1 year, but may be removed after that.
 
    DHCP
-      Dynamic Host Configuration Protocol, the protocol that allows a central device to assign unique IP addressess to all other devices.
+      Dynamic Host Configuration Protocol, the protocol that allows a central device to assign unique IP addresses to all other devices.
 
    FMS
       Field Management System, the electronics core responsible for sensing and controlling the FIRST Robotics Competition field.
@@ -70,4 +70,3 @@ FRC Glossary
 
    trajectory
       A trajectory is a smooth curve, with velocities and accelerations at each point along the curve, connecting two endpoints on the field.
-   
--- a/source/docs/software/frc-glossary.rst
+++ b/source/docs/software/frc-glossary.rst
@@ -56,8 +56,8 @@ FRC Glossary
    match
       A two (2) minute and thirty (30) second period of time in which an :term:`alliance` plays a FRC game.
 
-   NT
-      NetworkTables, a way to communicate key / value pairs, of data, between programs.
+   NetworkTables
+      A way to communicate key / value pairs, of data, between programs.
 
    RP
       Ranking Point, a way of ordering teams rewarding specific objectives in addition to winning the match.
--- a/source/docs/software/frc-glossary.rst
+++ b/source/docs/software/frc-glossary.rst
@@ -1,0 +1,73 @@
+.. include:: <isonum.txt>
+
+FRC Glossary
+============
+
+.. glossary::
+
+   accelerometer
+      A common sensors used to measure acceleration in one or more axis.
+
+   alliance
+      A cooperative of up to four (4) FIRST\ |reg| Robotics Competition teams.
+
+   auto
+      The first phase of each :term:`match` is called Autonomous (auto) and consists of the first fifteen (0:15) seconds.
+
+   COTS
+      Commercial off the shelf, a standard (i.e. not custom order) part commonly available from a vendor to all teams for purchase.
+
+   C++
+      One of the three officially supported programming languages.
+
+   depricated
+      Software that will be maintained for at least 1 year, but may be removed after that.
+
+   DHCP
+      Dynamic Host Configuration Protocol, the protocol that allows a central device to assign unique IP addressess to all other devices.
+
+   FMS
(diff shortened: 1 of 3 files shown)


source/docs/software/frc-glossary.rst
```diff
@@ -0,0 +1,73 @@
...
+   KOP chassis
+      The `AM14U4 <https://www.andymark.com/products/am14u4-kit-of-parts-chassis>`__ chassis distributed to every team in the :term:`KOP`.
+
+   LabVIEW
+      One of the three officially supported programming languages.
+
+   match
+      A two (2) minute and thirty (30) second period of time in which an :term:`alliance` plays a FRC game.
+
+   NT
```
Daltz333, line 59: > NT
Not really anywhere this is referenced as NT. Just leave it as NetworkTables

source/docs/software/frc-glossary.rst
```diff
@@ -0,0 +1,73 @@
...
+   auto
+      The first phase of each :term:`match` is called Autonomous (auto) and consists of the first fifteen (0:15) seconds.
+
+   COTS
+      Commercial off the shelf, a standard (i.e. not custom order) part commonly available from a vendor to all teams for purchase.
+
+   C++
+      One of the three officially supported programming languages.
+
+   depricated
```
Daltz333, line 23:
```suggestion
   deprecated
```

source/index.rst
```diff
@@ -33,6 +33,7 @@ Welcome to the *FIRST*\ |reg| Robotics Competition Control System Documentation!
    :caption: Programming Basics
 
    docs/software/what-is-wpilib
+   docs/software/frc-glossary
```
Daltz333, line 36: Thinking about it, maybe we should move this low down in the index. I don't think we want people to actively read it, but more use it as a page to reference for ``:term:``

source/docs/software/frc-glossary.rst
jasondaming: I don't think it is as rare as you think, from our docs:
Twice [here](https://docs.wpilib.org/en/latest/docs/software/vision-processing/roborio/using-multiple-cameras.html?highlight=NT#switching-the-driver-views)
Once [here](https://docs.wpilib.org/en/latest/docs/software/wpilib-tools/shuffleboard/custom-widgets/builtin-plugins.html?highlight=NT#networktables-plugin)
The rPi abbreviates is this way in the logs [here](https://docs.wpilib.org/en/latest/docs/software/vision-processing/wpilibpi/the-raspberry-pi-frc-console.html?highlight=NT#vision-status)
[Here](https://docs.wpilib.org/en/latest/docs/software/networktables/multiple-instances.html?highlight=NT#creating-multiple-instances-of-networktables)

There are probably more.

source/index.rst
jasondaming: makes sense

source/docs/software/frc-glossary.rst
Daltz333: I think in this specific case we should prefer NetworkTables, we can still reference to it when doing NT by doing ```:term:`NT <NetworkTables>` ``` 

We should probably avoid two letter abbreviations as they can quickly collide.

source/docs/software/frc-glossary.rst
```diff
@@ -0,0 +1,72 @@
+.. include:: <isonum.txt>
+
+FRC Glossary
+============
+
+.. glossary::
+
+   accelerometer
+      A common sensors used to measure acceleration in one or more axis.
```
prateekma, line 9: sensors -> sensor

source/docs/software/frc-glossary.rst
```diff
@@ -0,0 +1,72 @@
...
+      Dynamic Host Configuration Protocol, the protocol that allows a central device to assign unique IP addresses to all other devices.
+
+   FMS
+      Field Management System, the electronics core responsible for sensing and controlling the FIRST Robotics Competition field.
+
+   GradleRIO
+      The mechanism that powers the deployment of robot code to the roboRIO.
+
+   gyroscope
+      A device that measures rate-of-rotation.  It can add up the rotation measurements to determine :term:`heading` of the robot. (“gyro”, for short)
```
prateekma, line 36: I don't think the hyphens are necessary here. Just "rate of rotation" should work.

source/docs/software/frc-glossary.rst
```diff
@@ -0,0 +1,72 @@
...
+      Inertial Measurement Unit, a sensor that combines both an :term:`accelerometer` and a :term:`gyroscope` into a single sensor.
+
+   Java
+      One of the three officially supported programming languages.
+
+   KOP
+      Kit of Parts, the collection of items listed on the Kickoff Kit checklists, distributed to the team via FIRST Choice, or paid for completely (except shipping) with a Product Donation Voucher (PDV).
+
+   KOP chassis
+      The `AM14U4 <https://www.andymark.com/products/am14u4-kit-of-parts-chassis>`__ chassis distributed to every team in the :term:`KOP`.
```
prateekma, line 51: Should we mention that this is opt-out?

prateekma (changes requested):

prateekma (approved):

Daltz333:
Can we add a note somewhere in the styleguide that common used terms should get added here?

Daltz333 (approved):

PR #1285, opened 2021-04-01
Change instances of NT to use the :term: NetworkTables

### LINKED ISSUE #1282: Change instances of NT to use the :term: NetworkTables
I don't think it is as rare as you think, from our docs:
Twice [here](https://docs.wpilib.org/en/latest/docs/software/vision-processing/roborio/using-multiple-cameras.html?highlight=NT#switching-the-driver-views)
Once [here](https://docs.wpilib.org/en/latest/docs/software/wpilib-tools/shuffleboard/custom-widgets/builtin-plugins.html?highlight=NT#networktables-plugin)
The rPi abbreviates is this way in the logs [here](https://docs.wpilib.org/en/latest/docs/software/vision-processing/wpilibpi/the-raspberry-pi-frc-console.html?highlight=NT#vision-status)
[Here](https://docs.wpilib.org/en/latest/docs/software/networktables/multiple-instances.html?highlight=NT#creating-multiple-instances-of-networktables)

There are probably more.

jasondaming:
Closes #1282


Daltz333:
Please rebase

sciencewhiz:
Per the spec, Network Tables is correct, not NetworkTables.

jasondaming:
We need to get a lot more discussion in here about this then because NetworkTables is plastered everywhere, or does everyone else agree? @PeterJohnson

jasondaming:
That document also uses both Network Tables and network tables not sure why.

Daltz333 (approved):

PR #1288, opened 2021-04-02
Pin black to a specific version to fix check failures

sciencewhiz:


Daltz333 (approved):

prateekma:
Because of https://github.com/psf/black/issues/2079

PR #1297, opened 2021-04-04
Use pipenv for reproducability

### LINKED ISSUE #1296: Make builds more reproducible
frc-docs should consider using a python package management tool such as [Poetry](https://python-poetry.org/) or [PDM](https://pdm.fming.dev/) to pin all dependencies and make builds fully reproducible.

Daltz333:
Closes #1296


Pipfile
```diff
@@ -0,0 +1,28 @@
+[[source]]
+url = "https://pypi.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[packages]
+sphinx-rtd-theme = "==0.5.1"
```
auscompgeek, line 7: I dunno about you, but this feels weird. The point of having the Pipfile/Pipfile.lock split is so you _don't_ have to pin your dependencies in the Pipfile.

TheTripleV:
Can you alphabetize the pipfile please?

Pipfile
Daltz333: I'd rather keep them pinned, as I want to control when and which major dependencies get updated. I don't want to accidentally update a major dependency that runs "valid" but secretly changes behavior (this was done with docutils)

TheTripleV (approved):

TheTripleV (approved):

TheTripleV:
pipenv does not meet the requirements laid out by @Daltz333:
1. pin top level dependencies in pipfile
2. occassionally update the minor dependencies
3. manually choose when to update the top-level dependencies

Namely, requirement 2 is not under our control with pipenv. Updating the lockfile will update all dependencies.

Based on further testing, [pip-tools](https://github.com/jazzband/pip-tools) and [poetry](https://python-poetry.org/) do meet the all of the requirements.

I was not able to successful try [PDM](https://pdm.fming.dev/) without it crashing.

PR #1310, opened 2021-04-08
Add note about initial heading in HolonomicDriveController

prateekma:
Corresponds to https://github.com/wpilibsuite/allwpilib/pull/3290


prateekma:
See https://github.com/wpilibsuite/allwpilib/pull/3290#issuecomment-815399264

PR #1315, opened 2021-04-10
Fix secret syntax; use bool literal

AustinShalit:


Daltz333 (approved):

TheTripleV:
Would you also want to enable Push on workflow_dispatch? That way, a wpilib member manually triggering this workflow would make it publish to stable too.

Daltz333:
That's already there

AustinShalit:
This does not work.

TheTripleV:
workflow_dispatch doesn't push to stable if secrets.PUBLISH_STABLE isn't set.

PR #1322, opened 2021-04-11
Use URL for favicon

Daltz333:
Due to https://github.com/mozilla/discourse/issues/140


Daltz333:
urls are only supported in sphinx 4

PR #1333, opened 2021-04-23
Google search meta

Kevin-OConnor:
Add meta tag to verify for Google Search Console


Daltz333:
You can inject it via the meta tags that we use for opengraph. See the discord/chief Delphi example in conf.py

Kevin-OConnor:
> 
> 
> You can inject it via the meta tags that we use for opengraph. See the discord/chief Delphi example in conf.py

I'm fine with doing it that way instead. Is there a reason to prefer that?

Daltz333:
No reason. I wasn't actually aware of the <meta> tag.

PR #1343, opened 2021-04-28
Fixes syntax for C++ float double comparison

### LINKED ISSUE #1342: C++ unit test article should prefer EXPECT_DOUBLE_EQ() over EXPECT_EQ() with a tolerance
The examples as described don't actually compile. EXPECT_EQ() doesn't have a tolerance field. EXPECT_NEAR() does, but users should prefer EXPECT_DOUBLE_EQ() for checking equality of doubles.

jreneew2:
Fixes #1342. Should we include an example of EXPECT NEAR as java has a DELTA value while C++ doesn't?


calcmogul:
No, because EXPECT_NEAR() isn't recommended. Java not having an equivalent of EXPECT_DOUBLE_EQ() is a defect in JUnit imo.

prateekma (approved):

Daltz333 (approved):

PR #1345, opened 2021-04-30
Remove manual gradle project import instructions

sciencewhiz:
They won't work for 2022, and the import process will be too complicated
to do manually


Daltz333:
Sounds good

Daltz333 (approved):

PR #1358, opened 2021-05-12
Update Shuffleboard Gradle Example

### LINKED ISSUE #1325: Shuffleboard plugin sample build code uses gradle kts
https://docs.wpilib.org/en/stable/docs/software/wpilib-tools/shuffleboard/custom-widgets/creating-plugins.html

Nothing uses gradle kts, we don't recommend it, so we should not have our examples use gradle kts.

Daltz333:
Uses 2020 due to Shuffleboard not publishing 2021.

Closes #1325


ThadHouse (dismissed):

Daltz333:
Just rebased due to conflicts.

jasondaming (approved):

PR #1360, opened 2021-05-20
Create PathPlanning Section

Daltz333:
Migrate the following sections

- Robot Characterization
- Trajectory Following
- PathWeaver


source/docs/software/pathplanning/index.rst
```diff
@@ -0,0 +1,11 @@
+Path Planning
+=============
+
+Path Planning is the process of developing and "planning" paths that your robot can autonomously follow. These paths use the WPILib Trajectory Framework for generation and Ramsete for following. This section highlights the process of characterizing your robot for system identification, developing with the trajectory tracking API and visualizing your paths with PathWeaver. Users may also want to read the :ref:`generic trajectory following documents <docs/software/advanced-controls/trajectories/index:Trajectory Generation and Following with WPILib>` for non-commandbased usage.
```
jasondaming, line 4: Did you mean to say a Ramsete Controller?  Just Ramsete doesn't really work here especially for someone who doesn't know what that is.

jasondaming, line 4: I would remove the non-commandbased mention and mention for more detail.  People will generally go to the generic stuff to dive into more detail.

jasondaming (changes requested):

source/docs/software/pathplanning/index.rst
Daltz333: @jasondaming Eh, the Command-Based tutorial is very not useful for non-command teams for 3/4ths of it

jasondaming: okay then maybe mention both.  As is it sounds like the only reason people should go to the generic stuff is non-command based and that isn't true

source/docs/software/pathplanning/index.rst
```diff
@@ -1,7 +1,7 @@
...
+ "Path Planning is the process of creating and following paths that your robot can autonomously follow. These paths use the WPILib trajectory APIs for generation and a :ref:`Ramsete Controller <docs/software/advanced-controls/trajectories/ramsete:Ramsete Controller>` for following. This section highlights the process of characterizing your robot for system identification, developing with the trajectory tracking API and visualizing your paths with PathWeaver. Users may also want to read the :ref:`generic trajectory following documents <docs/software/advanced-controls/trajectories/index:Trajectory Generation and Following with WPILib>` for non-commandbased usage.
```
jasondaming: Using follow twice here sounds redundant.

jasondaming: "developing with the trajectory tracking API and visualizing your paths with PathWeaver."

1) developing seems odd
2) Pathweaver generates too

I propose: trajectories can be created using either the API or generated in PathWeaver then imported.

jasondaming: Still need to address the additional information in the last senatance

jasondaming (changes requested):

source/docs/software/pathplanning/index.rst
```diff
@@ -1,7 +1,7 @@
...
+Path Planning is the process of creating and following paths or trajectories. These paths use the WPILib trajectory APIs for generation and a :ref:`Ramsete Controller <docs/software/advanced-controls/trajectories/ramsete:Ramsete Controller>` for following. This section highlights the process of characterizing your robot for system identification, trajectory following with ramsete and usage of PathWeaver.. Users may also want to read the :ref:`generic trajectory following documents <docs/software/advanced-controls/trajectories/index:Trajectory Generation and Following with WPILib>` for additional information about the API and non-commandbased usage.
```
jasondaming, line 4: Can we just say trajectories rather than paths or trajectories.  There really isn't any option to "follow" a path right?

You used ramsete again and I don't think we should use that alone.

2 periods after PathWeaver

jasondaming (approved):

sciencewhiz (commented):
Robot Characterization shows up in the index for both tools and Path Planning, but PathWeaver only shows up in Path Planning

PR #1365, opened 2021-06-02
Fix RLIs for 2022

Starlight220:
Incremental fixing part 1.


Daltz333 (approved):
I went through and compared the contents of the changed Java and C++ rlis to the current site and verified that the examples maintained accuracy.

PR #1371, opened 2021-06-04
Add motor inversion information

### LINKED ISSUE #1355: Update examples and docs for #3340
Right side inversion is no longer a thing :tada:

https://github.com/wpilibsuite/allwpilib/pull/3340

Daltz333:
Closes #1355 

Also updated comment to coincide with the function parameter being renamed

### DIFF
--- a/source/docs/software/hardware-apis/motors/wpi-drive-classes.rst
+++ b/source/docs/software/hardware-apis/motors/wpi-drive-classes.rst
@@ -32,7 +32,27 @@ Drive Class Conventions
 Motor Inversion
 ^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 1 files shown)


Starlight220:
Perhaps use an RLI instead of an in-place code example? This is also a question for site-wide resources: should RLIs be encouraged with in-place code as a fallback?

I am for having primarily RLIs as they're easier to update, especially once Inspector is merged.

Daltz333:
There is no example in allwpilib that uses many of the convention functions. There is nothing to RLI and I'd rather merge now and RLI later. Creating examples in allwpilib is a difficult process for me. I would use them if they were available.

jasondaming (approved):

PR #1377, opened 2021-06-11
Set base url to root domain

Daltz333:
RTD no longer injects it for us (we still want to keep frcdocs.wpi.edu domains and not redirect them).


TheTripleV:
Does RTD not inject it just for frc-docs or for all rtd pages?
Sphinxext-opengraph currently autodetects html_baseurl (https://github.com/wpilibsuite/sphinxext-opengraph/blob/main/sphinxext/opengraph/__init__.py#L65).

Daltz333:
Re: https://github.com/readthedocs/readthedocs.org/issues/8250

AustinShalit (approved):

PR #1387, opened 2021-07-04
Add gradlew not found to debugging article

### LINKED ISSUE #1369: Add 'gradlew is not recognized...' error to common problems
The common cause is due to opening the wrong folder, such as opening `wrong-root` instead of `right-root` in this case:
```
- wrong-root/
   |- right-root/
      |- build.gradle
      |- gradlew.bat
```

Daltz333:
Close #1369


source/docs/software/basic-programming/reading-stacktraces.rst
```diff
@@ -547,3 +547,33 @@ In the example, the left motor controllers are plugged into PWM ports ``0`` and
                frc::PWMVictorSPX m_rearLeftMotor{1};
 
             };
+
+gradlew is not recognized...
```
prateekma, line 551: I don't think this should go in the stack trace documentation. Perhaps a different "general troubleshooting" document?

Daltz333: Would you be open to creating this document? I would not fill satisfactory creating a document with less than 20 lines of content.

Starlight220: Maybe this can be put with the VS Code C++ Intellisense issue? I think it's the `Known Issues` section.

Daltz333: That is an idea. Although, in the past we have only used the known issues article for things that are system wide/bugs that are unfixable. This would be categorized as a user error.

Daltz333:
I think it is fine to merge now and refactor into a more appropriate location later.

jasondaming (approved):

PR #1397, opened 2021-07-20
Minify PNG/JPG on HTML/PDF builds

### LINKED ISSUE #171: Compress images for PDF/EPUB
Should be a way to do this without running a script.

Daltz333:
Optimizes image quality of exported images on RTD with no noticeable differences.

Do we want to make this a proper extension or keep in RTD repo?

Close #171


source/conf.py
```diff
@@ -45,6 +48,7 @@
     "notfound.extension",
     "versionwarning.extension",
     "sphinx_panels",
+    "minify",
```
TheTripleV, line 51: It's not clear if this item is downloaded from pypi or is local (And I don't know how sphinx would resolve conflicts).
Could you change in-repo extensions to clearly denote that they are local and to prevent potential conflicts with pypi packages? Maybe create a separate `local_extensions` list that would only resolve in-repo extensions?

TheTripleV (changes requested):

Daltz333:
Research Food: https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images

Override ``image`` and replace with srcset and export multiple widths for browser to choose.

PR #1399, opened 2021-07-24
Generate responsive images

TheTripleV:


TheTripleV:
Apparently rav1e (another avif encoder) is faster than libaom. pillow_avif_plugin mentions rav1e in its source code but I don't know how well supported it is.
A standalone rav1e based avif encoder is available at kornelski/cavif-rs.
We should try this encoder to see if it improves build times.

TheTripleV:
Changes:
- refactor long nested functions
- use python's builtin mimetypes module
- Add support for svgs
  - svgs don't get rasterized fallbacks generated. Their size is used to prevent layout shift
  - Converting svgs to pngs seems to be the easiest way to get size (aspect ratio). svg2png.py takes care of that. I think it's pretty robust.
- Add support for VI snippets
  - VI snippets don't have webp's generated but do generate all pngs resolutions. The vi snippet data gets injected into all of the copies.

pyproject.toml
```diff
@@ -5,10 +5,14 @@ description = ""
...
 [tool.poetry.dependencies]
+python = "^3.6"
+beautifulsoup4 = "==4.9.3"
 doc8 = "==0.8.1"
 docutils = "==0.16.0"
 latex = "==0.7.0"
-python = "^3.6"
+Pillow = "==8.3.1"
+pillow-avif-plugin = "==1.2.0"
+pypng = "==0.0.20"
```
TheTripleV, line 15: These 2 dependencies should be removed.

PR #1410, opened 2021-08-08
Update RLIs for 2022

Daltz333:
Fixes various RLIs for 2022.

### DIFF
--- a/source/docs/software/advanced-controls/controllers/profiled-pidcontroller.rst
+++ b/source/docs/software/advanced-controls/controllers/profiled-pidcontroller.rst
@@ -114,16 +114,16 @@ A more complete example of ``ProfiledPIDController`` usage is provided in the El
 
   .. group-tab:: Java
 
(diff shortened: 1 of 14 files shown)


Daltz333:
Another round of RLI fixes for 2022. This batch is mostly inverting the right side in examples.

sciencewhiz:
Are these changes in the 2022 alpha tag? That would probably be easier to find and update in the future.

Daltz333:
Both are easily findable with a regex. I believe there are RLI breaking changes that have not been tagged yet.

sciencewhiz:
My worry isn't searching for the commit id when the commit id is already known, but rather figuring out which commit id to search for.

Daltz333:
Don't worry about the commit ID and use a regex with the prefix URL to help? I can update it to a tag, but there are RLIs that aren't tagged.

Daltz333:
I have walked through all changed documents and verified that all changes RLIs are the functional equivalent and are accurate for the existing version of documents.

For change RLIs in the future to a pinned tag, a simple regex can be used to capture the group between ``/allwpilib/`` and ``/``. This PR is good for approval and merge.

jasondaming (approved):
Looks good thanks for all the work!

PR #1416, opened 2021-09-08
Cleaning up static addressing info and clarifying details.

marshallmassengill:
I have this funny feeling that I'm opening a can of worms with this proposed change but this should clarify a few things.  Hopefully this helps.  If you don't use it, that's totally ok too. The only thing I'm not really happy about is the use of "field router" but that is what the device is called.  A more generic "Field Networking Equipment" could be substituted.

### DIFF
--- a/source/docs/networking/networking-introduction/ip-configurations.rst
+++ b/source/docs/networking/networking-introduction/ip-configurations.rst
@@ -46,7 +46,7 @@ It is also possible to configure static IPs on your devices to accommodate devic
 -  roboRIO - Static ``10.TE.AM.2`` would be a reasonable choice, subnet mask
    of ``255.255.255.0`` (default)
 -  Driver Station - Static ``10.TE.AM.5`` would be a reasonable choice,
-   subnet mask **must** be ``255.0.0.0`` to enable the DS to reach both the robot and FMS Server, without additionally configuring the default gateway. 
+   subnet mask **must** be ``255.0.0.0`` to enable the DS to reach both the robot and FMS Server, without additionally configuring the default gateway.
    If a static address is assigned and the subnet mask is set to ``255.255.255.0``, then the default gateway must be configured to ``10.TE.AM.4``.
 -  IP Camera (if used) - Static ``10.TE.AM.11`` would be a reasonable
    choice, subnet ``255.255.255.0`` should be fine
--- a/source/docs/networking/networking-introduction/ip-configurations.rst
+++ b/source/docs/networking/networking-introduction/ip-configurations.rst
@@ -26,7 +26,7 @@ This section describes networking when connected to the Field Network for match 
 On the Field DHCP Configuration
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-The Field Network runs a DHCP server with pools for each team that will hand our addresses in the range of 10.TE.AM.20 and up with subnet masks
+The Field Network runs a DHCP server with pools for each team that will hand out addresses in the range of 10.TE.AM.20 to 10.TE.AM.199 with a subnet mask of 255.255.255.0, and a default gateway of 10.TE.AM.4.
 
 -  OpenMesh OM5P-AN or OM5P-AC radio - Static ``10.TE.AM.1`` programmed by
    Kiosk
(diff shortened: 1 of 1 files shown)


source/docs/networking/networking-introduction/ip-configurations.rst
```diff
@@ -26,7 +26,7 @@ This section describes networking when connected to the Field Network for match
 On the Field DHCP Configuration
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-The Field Network runs a DHCP server with pools for each team that will hand our addresses in the range of 10.TE.AM.20 and up with subnet masks
+The Field Network runs a DHCP server with pools for each team that will hand out addresses in the range of 10.TE.AM.20 to 10.TE.AM.199 with a subnet mask of 255.255.255.0, and a default gateway of 10.TE.AM.4.
```
AustinShalit, line 29: While we are here we probably should clarify that the radio has its own DHCP server that passes out IPs to the robot local network. And the field only passes IPs to the local field network. No DHCP traffic should flow over the wireless link.

source/docs/networking/networking-introduction/ip-configurations.rst
```diff
@@ -39,13 +39,13 @@ The Field Network runs a DHCP server with pools for each team that will hand our
...
+It is also possible to configure static IPs on your devices to accommodate devices or software which do not support mDNS. When doing so you want to make sure to avoid addresses that will be in use when the robot is on the field network. These addresses are ``10.TE.AM.1`` for the OpenMesh radio, ``10.TE.AM.4`` for the field router, and anything greater than ``10.TE.AM.20`` which may be assigned to a device configured for DHCP or else reserved. The roboRIO network configuration can be set from the webdashboard.
 
 -  OpenMesh radio - Static ``10.TE.AM.1`` programmed by Kiosk
 -  roboRIO - Static ``10.TE.AM.2`` would be a reasonable choice, subnet mask
    of ``255.255.255.0`` (default)
 -  Driver Station - Static ``10.TE.AM.5`` would be a reasonable choice,
-   subnet mask **must** be ``255.0.0.0``
+   subnet mask **must** be ``255.0.0.0`` to enable the DS to reach both the robot and FMS Server
```
AustinShalit, line 48: Let's add a "without additional configuration" so folks don't get confused why DHCP works.

AustinShalit (changes requested):

source/docs/networking/networking-introduction/ip-configurations.rst
marshallmassengill: Agreed and if we're going to do this then I'd suggest being up front about the alternative, which is to configure the default gateway.

source/docs/networking/networking-introduction/ip-configurations.rst
marshallmassengill: Agreed.  That would help add additional context.

source/docs/networking/networking-introduction/ip-configurations.rst
AustinShalit: Works for me

AustinShalit (approved):

Daltz333 (dismissed):

Daltz333 (approved):

source/docs/networking/networking-introduction/ip-configurations.rst
```diff
@@ -26,7 +26,8 @@ This section describes networking when connected to the Field Network for match
 On the Field DHCP Configuration
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-The Field Network runs a DHCP server with pools for each team that will hand our addresses in the range of 10.TE.AM.20 and up with subnet masks
+The Field Network runs a DHCP server with pools for each team that will hand out addresses in the range of 10.TE.AM.20 to 10.TE.AM.199 with a subnet mask of 255.255.255.0, and a default gateway of 10.TE.AM.4.
```
Daltz333, line 29: While we are here 
```suggestion
The Field Network runs a DHCP server with pools for each team that will hand out addresses in the range of ``10.TE.AM.20`` to ``10.TE.AM.199`` with a subnet mask of ``255.255.255.0``, and a default gateway of ``10.TE.AM.4``.
```

Daltz333 (changes requested):

Daltz333 (approved):

PR #1419, opened 2021-09-11
Update PDP for wpilibsuite/allwpilib#3466`

sciencewhiz:


TheTripleV:
I think power-distribution-panel.rst should be renamed to power-distribution-module.rst to match the other changes.

Daltz333 (approved):

PR #1420, opened 2021-09-13
Open external links in new tab

Kevin-OConnor:
Adds a small JS file to add the target=_blank attribute to external links so they open in a new tab. Many of our external links are to downloads/installers and the user wants the documentation to stay open. This is especially useful for embedding the documentation like in a course currently under construction where the GitHub links will not work in the embed without this.


Daltz333:
:+1: on this change

Daltz333 (approved):

PR #1434, opened 2021-09-28
Italicize first

Kevin-OConnor:
Adds JS to italicize FIRST in places we can't just do it in the RST (e.g. copywrite notice)

### DIFF
--- a/source/_static/js/italicize_first.js
+++ b/source/_static/js/italicize_first.js
@@ -1,0 +1,10 @@
+$(document).ready(function () {
+    // get all body content as a string
+   var str = document.body.innerHTML;
+
+   // find all magic words and wrap them in a <em> tag
+   var result = str.replace(/(FIRST)/g, "<em>$1</em>");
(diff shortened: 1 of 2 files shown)


Daltz333:
Sphinx does have an open [issue](https://github.com/sphinx-doc/sphinx/issues/8068) for nesting markup and placing markup in titles.

Unfortunately, we cannot embed `<i>` or `<em>` directly into the `copyright` and `name` strings in `conf.py` because it doesn't work on some strings in the site, and additionally modifies the metadata and description. This is bad SEO.

I investigated adding a post-sphinx hook that goes through the HTML (similar to `inner.html` javascript done here), but it's not worth the effort than this 1kb JavaScript does. I dislike how wildcard-ey it is, but I don't see a better solution for now. I doubt we use or will use any instances of an all caps `FIRST` outside of the context of the organization.

Daltz333 (approved):

PR #1451, opened 2021-10-09
Update Game Tools and Imaging for 2022

sciencewhiz:


Daltz333:
We probably want to add a small section where it's like

```
Imaging the RoboRIO 2
=====================

The RoboRIO 2 can be imaged the same as the original RoboRIO. However, imaging the microSD card directly with :doc:`Etcher <link to rpi imaging docs here>` is recommended instead. Simply drag the RoboRIO 2 zip file located at (`~/XXXX/XXXX/XXXX`) into Etcher and press the :guilabel:`Image` button.

```

sciencewhiz:
Per our discussion there's more that needs to be understood and documented about imaging the roboRIO2 SD card, so that should be done later.

ThadHouse:
When we add the docs on how to image with the SD card, we also need to document how to then set the team number by itself in the imaging tool. You do it by setting the team number, clicking the Edit Startup Settings radio button, then hitting apply.

Daltz333:
This PR is borked.

PR #1481, opened 2021-10-21
Remove/Update 2021 specific information

sciencewhiz:
Remove relicensing LabVIEW and 2020 game data


Daltz333:
I wouldn't remove the game data until we know it's useless this year. People could think removing the game data is a leak that there is no game data, when we really don't know.

Daltz333:
I'm fine if Kevin 👍

PR #1488, opened 2021-10-24
Adds PCM to the list of devices disabled on brownout

### LINKED ISSUE #1479: Brownout documentation neglects pneumatic devices
The documentation says that `CAN-based motor controllers are sent an explicit disable command`, but this also applies to the CTRE PCM / REV PH.

jasondaming:
Closes #1479

Do you think we should have all of the Control system acronyms in the glossary and use a :term: here instead?


Daltz333:
I like glossary terms. We should follow wikipedias example and only link a term on the first reference in an article.

PR #1507, opened 2021-10-31
Show game manual and qa links on mobile

TheTripleV:


Daltz333:
I marked this as Needs Design Decision as I'm not fully in agreement about removing from the top navigation panel. Perhaps we can replace the links up there with something else?

Daltz333:
Per discussion in Slack, we shouldn't remove the top navigation panel and it's pretty pointless to have it with no content.

As a workaround, we can still add this to the TOC, but just have it `hidden` on desktop viewport.

PR #1528, opened 2021-11-20
Add unofficial resources

### LINKED ISSUE #1522: Update Support Resources to Include Other Support resources
Here is the current [page](https://docs.wpilib.org/en/latest/docs/software/support/support-resources.html)

Add rev documentation
Add Limelight docs / support
Add PhotonVision docs / support
Add unofficial support: CD & FRC Discord

Daltz333:
Close #1522 

I opted to not include documentation links, as that would be better fitted for an example page. Vendors are welcome to PR to this page and provide their own links.

### DIFF
--- a/source/docs/software/support/support-resources.rst
+++ b/source/docs/software/support/support-resources.rst
@@ -42,7 +42,7 @@ There are useful forms of support provided by the community through various foru
 - `Chief Delphi <https://chiefdelphi.com>`__
 - `FRC Discord <https://discord.gg/frc>`__
 - `PhotonVision Support (Discord) <https://discord.com/invite/wYxTwym>`__
-- `Limelight Support <https://docs.limelightvision.io/en/latest/>`__
+- `Limelight Support <https://limelightvision.io/pages/contact-us>`__
 
 Bug Reporting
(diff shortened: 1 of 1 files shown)


sciencewhiz:
Why would limelight be under unofficial support?

Daltz333:
Why would limelight not?

sciencewhiz:
What's the definition of unofficial vs official? Limelight is a company like rev and ctre.

Daltz333:
I would say FIRST would have to make that distinction. I just went off of approved motor hardware vendors.

Kevin-OConnor:
I agree with Joe that calling Limelight's official support "unofficial" seems a bit awkward. Perhaps an additional category for "Other Vendors" or similar? Limelight's official support and CD are pretty different types of resources.

sciencewhiz:
Other vendors support:
https://www.kauailabs.com/support/navx-mxp/
https://www.playingwithfusion.com/contactus.php
copperforge is just an email address, but maybe we could link this: https://copperforge.cc/docs/software/libcu/

sciencewhiz (approved):

PR #1536, opened 2021-12-03
Fix shuffleboard code example missing return statement

carelesshippo:
Added a return statement to a function that should return a list but didn't.


carelesshippo:
Sorry I don't know which branch I should merge to for this. What should I do?

sciencewhiz:
Thanks for the fix! PRs should be targeted to the main branch rather then the stable branch as described here: https://docs.wpilib.org/en/stable/docs/contributing/frc-docs/contribution-guidelines.html#creating-a-pr

carelesshippo:
> Thanks for the fix! PRs should be targeted to the main branch rather then the stable branch as described here: https://docs.wpilib.org/en/stable/docs/contributing/frc-docs/contribution-guidelines.html#creating-a-pr

Ok thanks, I opened #1537 that hopefully is targeting the correct branch.

Issue #515, opened 2020-01-25
Common Issues and Solutions


Daltz333:
Meant to compile a list of common issues and solutions related to building and deploying to the robot.

Please share

Starlight220:
> Meant to compile a list of common issues and solutions related to building and deploying to the robot.

Is this only about build/deploy or also about code problems?
Do CAN problems (i.e `CAN Recieve Timeout`) count?

Daltz333:
I think code problems would also be acceptable. CAN is an icky situation as there is no CAN controller in the KOP

Starlight220:
More with PCM/PDP issues, 99% of the times I encountered a `CAN Recieve Timeout` was because of PCM problems...

Starlight220:
A troubleshooting section for simulation should also be made...

Daltz333:
There is an issue with extending simulation docs already. I usually get CAN Receive Timeout with TalonSRXs when there is an issue with the wiring.

Starlight220:
> CAN is an icky situation as there is no CAN controller in the KOP

What motor controllers are in the KOP? And I think these docs should have troubleshooting for everything except maybe extreme vendor stuff. And if the blink codes for the TalonSRX, etc. were added, why shouldn't basic troubleshooting be added too?

AustinShalit:
This past year the KOP motor controller was the Victor SPX. The reason why we do not include troubleshooting for vendor devices is because that was part of the arrangement we came to when we decided to split out vendor hardware from the library. We make exceptions for specific pieces of documentation. For example, the status light page is great because it is a single document that contains all of the status light codes.

Starlight220:
> This past year the KOP motor controller was the Victor SPX

And the VictorSPX isn't CAN?
I understand the decision that troubleshooting vendor products should not be in the main docs, basic CAN troubleshooting should be in the main docs as that is the direction FRC is going in, and it applies to products of most vendors and KOP equipment (PCM,PDP), if the latter isn't a reason itself.

Daltz333:
To be clear. We can include CAN devices. However, if we do anything with vendors, we have to mention them all.

Starlight220:
Would we want a separate troubleshooting file for each "topic", or a single file with sections?

Daltz333:
I was thinking a section for each topic would be cleaner. It already exists in some topics. In specific, this issue relates to creating a FAQ of common deployment issues (network connection, null pointer exception, etc)

[...]

PR #1547, opened 2021-12-12
Add quote regarding asking for help

Daltz333:


sciencewhiz:
While true, I'm not sure that this is helpful (in its current form). I think someone being lazy is a very uncommon reason why there's limited detail posted. I think not understanding what is relevant is much more common. Something like an FRC version of ESR's how to ask questions the smart way is what's really needed.

Daltz333:
Is this the document you refer to? https://d.cxcore.net/Eric%20S%20Raymond/How%20To%20Ask%20Questions%20The%20Smart%20Way.pdf

sciencewhiz:
yes

jasondaming:
I also don't think "How to read Stacktraces" is the right document for this.  If you want to give advice on general seeking help I think that would be ideal in a [FAQ](https://github.com/wpilibsuite/frc-docs/issues/515) type document.  This already says "full text" and I can't even count how often we get only partial traces usually with the wrong info.

PR #1566, opened 2021-12-20
Getting to Know Romi documentation

agasser:
We found the Romi documentation a bit difficult to understand until we got deeper into using the robot. Since this it the "Getting Started" documentation, I have a few proposals that I think make the documentation more clear for new users:

1. The channel documentation was confusing, and one of the tables wasn't formatted correctly so it didn't render. I put the channel documentation under each hardware heading to clearly associate the channels with the hardware.
2. There was inconsistent use of the terms GPIO "pins" and "channels". The Web UI calls the EXT ports "pins" so I used this same terminology. There is still inconsistency with how WPILib classes call them IO "channels" where the Web UI calls them "ports", but this is not a documentation issue. I recommend using "pin" to describe the physical pins and "channel" to describe the configured IO channel.
3. There were some references to the 32U4 pin numbers in the channel documentation, but these numbers were not explained or labeled in a picture. They assume knowledge that was not included in the Getting Started docs, so I removed them and referred to the pins only as EXT0-4.
4. I added a photo with the onboard buttons and LEDs circled to make this section a little more clear. I did not immediately understand that this hardware was physically on the 32U4, or where the components were located.
5. I added two notes that the buzzer is currently not supported by WPILib.


Daltz333:
Needs a rebase

source/docs/romi-robot/getting-to-know-romi.rst
```diff
@@ -23,7 +23,9 @@ The Romi has the following built-in hardware/peripherals:
 - 1x Inertial Measurement Unit (IMU)
 - 3x LEDs (green, yellow, red)
 - 3x pushbuttons (marked A, B, and C)
-- 5x configurable GPIO channels
+- 5x configurable GPIO channels (EXT)
+
+.. note:: The built-in Buzzer is currently not supported.
```
sciencewhiz, line 28: Isn't the buzzer used by the romi firmware? Perhaps rather then saying not supported say it's not user programmable.

agasser: From what I could find, the buzzer is not supported by WPILib. Yes, it is physically on the control board and supported with Arduino, so I see how this isn't completely clear. I will restate it.

agasser:
> Needs a rebase

Done

source/docs/romi-robot/getting-to-know-romi.rst
```diff
@@ -34,6 +37,32 @@ The wheels have a diameter of 70mm (2.75"). They have a trackwidth of 141mm (5.5
...
+
++-------------+--------------------------+
+| Channel     | Romi Hardware Component  |
++=============+==========================+
+| PWM 0       | Left Motor               |
++-------------+--------------------------+
+| PWM 1       | Right Motor              |
++-------------+--------------------------+
+
+.. note:: The right motor is hardwired to spin in a backward direction when positive output is applied. Thus, the corresponding motor controller needs to be inverted in robot code.
```
Daltz333, line 50: hardwired is technically the wrong term here. The right motor is actually wired inverted so they both move forward with the same voltage in the *hardware*. The romi firmware inverts the right motor in code to match standard FRC robot behavior. So perhaps replace it with "configured"

agasser: The text said "hardwired" before my change, this was not my wording. The old wording left out the important fact that it doesn't matter to the user how it's _wired_, it matters how to _program_ it with WPILib. But, my proposed wording is not precise if the firmware is inverting it.

Daltz333 (approved):

PR #1572, opened 2021-12-21
Correct right motor inversion statement

### LINKED ISSUE #1571: Romi docs say right motor is inverted, it is not
The [Getting to know your Romi page](https://docs.wpilib.org/en/latest/docs/romi-robot/getting-to-know-romi.html) says:

> The right motor is hardwired to spin in a forward direction when forward stick movement is applied on a joystick. Thus there is no need to invert the corresponding motor controller in robot code.

This does not appear to be true. Before 2022 the `DifferentialDrive` class inverted the right motor controller automatically, so maybe this was a misunderstanding. With 2022 beta 4 I have found that the right motor controller needs to be inverted.

agasser:
Resolves #1571

### DIFF
--- a/source/docs/romi-robot/getting-to-know-romi.rst
+++ b/source/docs/romi-robot/getting-to-know-romi.rst
@@ -121,7 +121,7 @@ The table below shows the default pins. These are configurable (with the excepti
 | PWM 3       | Pin 22 / A4             |
 +-------------+-------------------------+
 
-.. note:: The right motor is hardwired to spin in a forward direction when forward stick movement is applied on a joystick. Thus there is no need to invert the corresponding motor controller in robot code.
+.. note:: The right motor is hardwired to spin in a backward direction when positive output is applied. Thus, the corresponding motor controller needs to be inverted in robot code.
 
 GPIO Channels
 ^^^^^^^^^^^^^


sciencewhiz:
This is subtly different then just inverting the previous statement, since forward on a joystick is negative.

Since I don't have a Romi, does sending the same value to both left and right motors cause it to spin in circles? Or is this about making sure forward and back are correct.

agasser:
Sending the same output to both motors makes the Romi spin in circles. This statement is about motor inversion, but it does reference the joystick. In my PR I removed the reference to the joystick to avoid that additional confusion.

There is a separate statement about joystick inversion on the [Programming the Romi](https://docs.wpilib.org/en/latest/docs/romi-robot/programming-romi.html#running-a-romi-program) page.

sciencewhiz:
The fact that the right side isn't inverted appears to be a bug in the Romi firmware. https://github.com/wpilibsuite/wpilib-ws-robot-romi/issues/115

agasser:
> The fact that the right side isn't inverted appears to be a bug in the Romi firmware. [wpilibsuite/wpilib-ws-robot-romi#115](https://github.com/wpilibsuite/wpilib-ws-robot-romi/issues/115)

It is unexpected that one motor would be hardcoded to be inverted, but it's up to you maintainers to decide whether to change the docs or change the firmware.

Daltz333 (approved):

Daltz333:
It has been discussed and determined that when both motors are given the same inputs, they should turn in opposing directions.

PR #1578, opened 2021-12-24
Fix NT Desktop Example

### LINKED ISSUE #1575: NetworkTables Listening For Value Changes Incorrect Variable Names
1. The two uses of variables named "table" are different and must be called different names.
```Java
table.addEntryListener("X", (table, key, entry, value, flags) -> {
               System.out.println("X changed value: " + value.getValue());
```
2. The return variable named "event" is subsequently incorrectly referenced as "value" which should have been called "event". (Oh the tyranny of copy/paste!)
```Java
yEntry.addListener(event -> {
               System.out.println("Y changed value: " + value.getValue());
```

Daltz333:
Close #1575

### DIFF
--- a/source/docs/software/networktables/listening-for-change.rst
+++ b/source/docs/software/networktables/listening-for-change.rst
@@ -28,22 +28,22 @@ One of the more common cases for NetworkTables is waiting for a value to change 
             NetworkTableInstance inst = NetworkTableInstance.getDefault();
 
             //get a reference to the subtable called "datatable"
-            NetworkTable table = inst.getTable("datatable");
+            NetworkTable datatable = inst.getTable("datatable");
 
             //get a reference to key in "datatable" called "Y"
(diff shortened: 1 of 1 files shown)


jasondaming:
I think in the first paragraph of this article each mention of "Driver Station" (3 of them) should be "Driver Station computer".  I think this will help reduce confusion between the computer and a different program on it and the actual FRC Driver Station program.

Daltz333:
That can be done in a separate PR.

jasondaming:
I didn't understand this as being urgent.  I don't think it is big enough to warrant a separate PR.  It seems like our policy is when a document is changed we change all of it.

jasondaming (approved):

Daltz333:
PRs are cheap. Refactoring a whole document has never been a policy when a change is done. Not since I had to fix all the garbage when I ran an autoconveter on the original markdown files.

PR #1579, opened 2021-12-28
Add REV Control System Overview

Daltz333:


sciencewhiz:
Since ZTR doesn't cover the Rev components, should the diagram wait to be added until it does?

Daltz333:
I do not see any harm in adding it now. I do not have a strong opinion and can remove it from the how-to-wire-a-robot page if you'd like, but it still can have a home in the control system hardware components page.

sciencewhiz:
Yeah, that's what I was suggesting just remove it from ZTR not remove it entirely.

Daltz333:
Sounds good

sciencewhiz (approved):

PR #1584, opened 2021-12-30
Fix publish workflow

Daltz333:
Some syntax errors that were caused on the workflow_dispatch update awhile ago.

@AustinShalit **please confirm that PUBLISH_STABLE is False**


AustinShalit:
I've set it to `false`

AustinShalit (approved):

PR #1599, opened 2022-01-04
Port IR@Home Guides

Kevin-OConnor:
Port of the guides from 2021 IR@Home guides. Still have to look at the vision one, not sure if that is going to make sense as it's own new doc or a revision/merger with "What is Vision?" and/or "Strategies for Vision Programming". The other two guides about selecting drivers and practice don't seem to fit FRC Docs and will find another home.


jasondaming:
We should probably pull the "Simple Autonomous Example" out of step 4 if we have this.

Kevin-OConnor:
> We should probably pull the "Simple Autonomous Example" out of step 4 if we have this.

Not sure I agree with that. That section is a short description of code that is in the example project and helps them understand what it does.

This article provides a more (very) high level overview of autonomous movement in general and how to program your own.

jasondaming:
I think the nested bullets don't format well here and would use (sub)headings.  Under those I think bullets will be fine but IMO multiple layers of bullets aren't very readable.

I would split up the external sensor between cameras and other.  Similar to internal.

Seems like the bullets under choose an approach could just be integrated into their sections.  I would at least remove the word approach on both of the approaches.

Use tabs to separate LV vs text based.  It is hard to combine them.  Not a big fan or sequence structures in LV.  Teams use them as a crutch and abuse them.  Just use dataflow, adds another reason to connect error wires.

jasondaming:
Additional things I think it would be helpful to discuss:
- Placing the robot on the field.  Many teams don't realize how important this is for Auto repeatability.  Placing in relation to field elements to add repeatability.  A note about using a measuring device (need to fight so many thinking this is illegal, assuming the rules hold) in relation to nearby field elements to make placement quicker.  Use feet or some some other known quantity to measure if a tape measure becomes illegal again (please don't do this).  

- Auto Strategy - Encourage teams to think about where their partners will want to place their robot and to consider having a backup routine for when their ideal spot is taken.  Can show/link Sendable Chooser and the LV auto selector here.

- Setting your robot up to play the rest of the match- I am a big fan of teams trying to do something to try and get in a better position for the rest of the match.  It can save precious seconds.  This could possibly go into other potential improvements.

Daltz333 (changes requested):
Can we change the smartquotes to regular quotes? They mess things up somewhere (I don't remember why).

Daltz333:
@Kevin-OConnor should this be closed?

Kevin-OConnor:
I don't think so, I think we still want this content, not sure why I never took it off Draft. Looks like it probably needs an update or two now (I saw at least Ramsete at a quick glance)

PR #1601, opened 2022-01-05
Fix coloring on index block

Daltz333:
Not sure when this got removed, or if it was ever there. It's supposed to be there.

### DIFF
--- a/source/index.rst
+++ b/source/index.rst
@@ -72,7 +72,7 @@ Community translations can be found in a variety of languages in the bottom-left
       :classes: btn-outline-primary btn-block
 
    ---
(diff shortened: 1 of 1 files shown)


sciencewhiz:
Still doesn't look quite right on iOS. 
2037188A-4024-401C-A0BD-E547E7DF8A73

sciencewhiz:
Returning teams also has the wrong color.

sciencewhiz:
70B90FEA-96CC-4FF0-876A-79140048FFC2

sciencewhiz (approved):

PR #1638, opened 2022-01-10
Add another solution to m1 mac problem using SDKMAN!

carelesshippo:
SDKMAN! can be helpful for solving this issue with the local version install feature.


Daltz333:
I am not sure we want to even suggest a workaround besides the forementioned workarounds. @ThadHouse?

ThadHouse:
No, I don't want any other suggestion then to use the wpilib installer. That is the supported and tested solution for M1

carelesshippo:
> No, I don't want any other suggestion then to use the wpilib installer. That is the supported and tested solution for M1

This uses the JDK installed by the WPILib installer, just making it much easier to access and switch between (command `sdk use wpilib`).

PR #1641, opened 2022-01-10
Fixes to Git data article

ctorek:
Previously in this article, the way to make the static file deploy task depend on the Git file generation tasks seems to be incorrect after more testing. In the article:

```groovy
deploy.targets.roborio.artifacts.frcStaticFileDeploy.deployTask.dependsOn(writeBranchName)
```

I believe that it should be:

```groovy
deploy.targets.roborio.artifacts.frcStaticFileDeploy.dependsOn(writeBranchName)
```

I could be inaccurate in this, but the top one does not seem to work for me. I have tested the bottom one and it does work, but I'm not aware if it is correct according to deploy-utils internals, so this may need more review.

Also fixed a random capitalization error at the bottom of the article.

### DIFF
--- a/source/docs/software/advanced-gradlerio/deploy-git-data.rst
+++ b/source/docs/software/advanced-gradlerio/deploy-git-data.rst
@@ -49,7 +49,7 @@ The next step is to make the deploy task depend on the task you created, so that
 
 .. code-block:: groovy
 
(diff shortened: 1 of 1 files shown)


Daltz333:
@ThadHouse ?

ThadHouse:
I'll have to double check. I didn't think the root artifact had a dependsOn, but I could be wrong.

ThadHouse:
Ah, yes you are correct. Good spot

PR #1653, opened 2022-01-14
Adding the vendor URLs

jasondaming:
Without the update feature adding vendordeps can be painful.  Many vendors also bury the URL in their documentation.

### DIFF
--- a/source/docs/software/vscode-overview/3rd-party-libraries.rst
+++ b/source/docs/software/vscode-overview/3rd-party-libraries.rst
@@ -68,7 +68,7 @@ To remove a library dependency from a project, select :guilabel:`Manage Current 
 Libraries
 ---------
 
-.. warning:: Many vendors, such as CTRE & REV, recommend using their offline installers over using the online vendor json URL.  This recommendation is because the offline installer is bundled with the additional programs that are extremely useful when working with their devices.
+.. warning:: Many vendors, such as CTRE & REV, recommend using their offline installers over using the online vendor json URL.  This recommendation is because the offline installer is bundled with additional programs that are extremely useful when working with their devices.
 
 Click these links to visit the vendor site to see whether they offer online installers, offline installers, or both.  URLs below are to plug in to the :guilabel:`VS Code` -> :guilabel:`Install New Libraries (online)` feature.  The vendor json will break if a team has not ran an "online" build after 30 days, since the cache will clear.
 
--- a/source/docs/software/vscode-overview/3rd-party-libraries.rst
+++ b/source/docs/software/vscode-overview/3rd-party-libraries.rst
@@ -68,7 +68,7 @@ To remove a library dependency from a project, select :guilabel:`Manage Current 
 Libraries
 ---------
 
-.. warning:: Many vendors (like CTRE & REV) with additional programs to aid in the use of their libraries prefer the offline installer.
+.. warning:: Many vendors, such as CTRE & REV, recommend using their offline installers over using the online vendor json URL.
 
 Click these links to visit the vendor site to see whether they offer online installers, offline installers, or both.  URLs below are to plug in to the :guilabel:`VS Code` -> :guilabel:`Install New Libraries (online)` feature.  The vendor json will break if a team has not ran an "online" build after 30 days, since the cache will clear.
 
--- a/source/docs/software/vscode-overview/3rd-party-libraries.rst
+++ b/source/docs/software/vscode-overview/3rd-party-libraries.rst
@@ -70,7 +70,7 @@ Libraries
 
 .. warning:: Many vendors, such as CTRE & REV, recommend using their offline installers over using the online vendor json URL.  This recommendation is because the offline installer is bundled with additional programs that are extremely useful when working with their devices.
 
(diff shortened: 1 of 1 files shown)


Daltz333:
I think it's important that we mention somewhere/somehow that it's recommended to view the documentation and that not all vendors recommend using the vendor json. The vendor json will break if a team has not ran an "online" build after 30 days, since the cache will clear.

jasondaming:
I do say that some vendors prefer offline in the warning.  View the documentation seems redundant.  I do like the note about the online build.

source/docs/software/vscode-overview/3rd-party-libraries.rst
```diff
@@ -68,24 +68,33 @@ To remove a library dependency from a project, select :guilabel:`Manage Current
 Libraries
 ---------
 
-.. warning:: These are **not** links to directly plug in to the :guilabel:`VS Code` -> :guilabel:`Install New Libraries (online)` feature. Click these links to visit the vendor site to see whether they offer online installers, offline installers, or both.
+.. warning:: Many vendors (like CTRE & REV) with additional programs to aid in the use of thier libraries prefer the offline installer.  Click these links to visit the vendor site to see whether they offer online installers, offline installers, or both.
```
sciencewhiz, line 71: I think the first sentence is the warning. The second sentence is more instructions and should probably be removed from the warning so it doesn't get lost.

source/docs/software/vscode-overview/3rd-party-libraries.rst
```diff
@@ -68,24 +68,33 @@ To remove a library dependency from a project, select :guilabel:`Manage Current
 Libraries
 ---------
 
-.. warning:: These are **not** links to directly plug in to the :guilabel:`VS Code` -> :guilabel:`Install New Libraries (online)` feature. Click these links to visit the vendor site to see whether they offer online installers, offline installers, or both.
+.. warning:: Many vendors (like CTRE & REV) with additional programs to aid in the use of their libraries prefer the offline installer.
```
sciencewhiz, line 71: I think there's some missing punctuation, but I'm no English major

jasondaming: @Daltz333 summoning the English major!

Daltz333: If you take away the (), you are left with a broken initial sentence. I also think we should replace `prefer` with `recommend`.

jasondaming:
I dislike that then new warning wording doesn't have reasoning as to why they prefer the offline but I think it should be fine.

sciencewhiz:
Is there a reason you removed the reasoning?

jasondaming:
That is what Dalton suggested I change it to

Daltz333:
Oh, I just meant the first sentence. My apologies.

source/docs/software/vscode-overview/3rd-party-libraries.rst
```diff
@@ -68,7 +68,7 @@ To remove a library dependency from a project, select :guilabel:`Manage Current
 Libraries
 ---------
 
-.. warning:: Many vendors, such as CTRE & REV, recommend using their offline installers over using the online vendor json URL.
+.. warning:: Many vendors, such as CTRE & REV, recommend using their offline installers over using the online vendor json URL.  This recommendation is because the offline installer is bundled with the additional programs that are extremely useful when working with their devices.
```
Daltz333, line 71: GitHub UI is being weird. But this sentence "This recommendation is because the offline installer is bundled with the additional programs that are extremely useful when working with their devices." should be changed to "This recommendation is because the offline installer is bundled with additional programs that are extremely useful when working with their devices."

Daltz333 (dismissed):
I'm happy. If @PeterJohnson can make sure this is the acceptable way we want to present this, we can merge!

source/docs/software/vscode-overview/3rd-party-libraries.rst
```diff
@@ -68,24 +68,33 @@ To remove a library dependency from a project, select :guilabel:`Manage Current
 Libraries
 ---------
 
-.. warning:: These are **not** links to directly plug in to the :guilabel:`VS Code` -> :guilabel:`Install New Libraries (online)` feature. Click these links to visit the vendor site to see whether they offer online installers, offline installers, or both.
+.. warning:: Many vendors, such as CTRE & REV, recommend using their offline installers over using the online vendor json URL.  This recommendation is because the offline installer is bundled with additional programs that are extremely useful when working with their devices.
+
+Click these links to visit the vendor site to see whether they offer online installers, offline installers, or both.  URLs below are to plug in to the :guilabel:`VS Code` -> :guilabel:`Install New Libraries (online)` feature.  The vendor json will break if a team has not ran an "online" build after 30 days, since the cache will clear.
```
PeterJohnson, line 73: "The vendor json will break if a team has not ran an "online" build after 30 days, since the cache will clear."

It would be good to state that this is not the case if the vendor's offline installer is used.

PeterJohnson: Yes that should work.  It's technically a bit more subtle than that (if they install online, and later install the offline...) but that's likely to be more confusing than useful.

Daltz333 (approved):

PR #1658, opened 2022-01-15
Remove Axis Camera

sciencewhiz:
NI removed Axis Camera tool and cameras had been discontinued for years


Daltz333:
I don't think we should remove Axis Camera without removing or deprecating the corresponding WPILib APIs. Marked as Design Decision and can review next WPILib meeting.

sciencewhiz:
The other option is to go back to the manual configuration via webserver that teams had to do before the NI tool existed. But I doubt anyone has the ability to test that.

Daltz333 (dismissed):

Daltz333 (approved):

PR #1659, opened 2022-01-15
Fix Capitalization of SetInverted

davidorser:


sciencewhiz:
setInverted is correct for java, which is the vast majority of users.

Also, PRs should be targeted to main: https://docs.wpilib.org/en/stable/docs/contributing/frc-docs/contribution-guidelines.html#creating-a-pr

Daltz333:
We default to Java for many of the functions throughout the docs, as that is where a vast majority of our users are.

PR #1664, opened 2022-01-16
Update known issues for 2022.2.1

sciencewhiz:


Daltz333 (approved):

Daltz333 (commented):
Should we just wait until 2022.2.1 is published and change the workaround to that. I still want to keep this in as its so critical.

Daltz333 (approved):

PR #1672, opened 2022-01-19
Update hidden dependencies & bump to Python 3.9

Daltz333:
Pillow requires Python >3.7, I went ahead and bumped up to 3.8 since that's what we build with on everything.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -9,7 +9,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-python@v2
       with:
-        python-version: 3.8
+        python-version: 3.9
     - name: Install Python Dependencies
       run: |
         pip install -r source/requirements.txt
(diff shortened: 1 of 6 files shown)


sciencewhiz:
There's several things in CI that are run with 3.7, should those get bumped to 3.8 also?

TheTripleV:
Pillow dropped 3.6 but still supports 3.7. We could keep 3.7 until it EOLs next June.

We should also bump explicit dependencies (Sphinx mainly). The docs can't be build on 3.10 right now.

Daltz333:
I don't like bumping the explicit dependencies mid season, besides just a minor version. The 3.10 bump I believe comes with a major version bump.

Daltz333:
Okay, Python 3.10 does require a major Sphinx bump (4.3.0). I will otherwise bump to 3.9 (that's what comes with the Python in the Windows store) and bump the CI.

Daltz333:
@sciencewhiz I have updated to Python 3.9 (not 3.10 as Sphinx is incompatible without updating to >=4.3.0). Additionally, the CI scripts and build instructions have been updated.

TheTripleV (approved):

PR #1683, opened 2022-01-24
Update REV CS Image to use 10amp fuse for PH

Daltz333:


TheTripleV:
Are the PDH, PH, and Radio Power Module meant to get darker?

Daltz333:
Yeah, we were on the original version of the image and this is 2 versions ahead

TheTripleV (approved):

NoahAndrews:
Why was this changed? Teams are supposed to use a 15A fuse there.

Daltz333:
See discussion in WPILib slack. There was a messup in communications resulting from a mixture of bad information.

PR #1698, opened 2022-01-30
Add radio troubleshooting step to upgrade npcap

sciencewhiz:


Daltz333:
Pending investigation to determine this actually is an issue and also resolves the issue.

Daltz333 (approved):
Going to go ahead and merge, since it won't hurt anything

PR #1701, opened 2022-01-30
Rename "robot characterization" to "system identification."

Oblarg:


Oblarg:
Does `redirects` accept globs?

Also, this doesn't seem like it actually belongs in the pathplanning section.

TheTripleV:
No, redirects does not accept globs.

Daltz333:
You can run ./make rediraffewritediff to have it automatically do it for you.

It was moved to the path following section as that is what the vast majority of teams use it for. It helps streamline the sequence of information you need for path following without having to "jump" from various "reference manuals". So while it is *technically* correct, it was deemed an appropriate trade-off for navigations sake.

sciencewhiz:
The tool also walks the redirect path. It's not necessary to change intermediate redirects, only add new ones.

Daltz333 (approved):

PR #1704, opened 2022-02-02
Implement Offline PWA Support

Daltz333:
Warning-Banner

## What is PWA?

Progressive Web Applications (PWAs) are the fancy technology sponsored by Mozilla and Google several years ago that encouraged web based native style applications. This allows websites to be turned into fully offline capable applications with notification support.

## Why?

Franky, it's super cool and enables a couple of benefits for us:
- Offline mobile support. This will be extremely useful at events
- Notifications for important FIRST/WPILib critical updates

## Important Usage Information

As of currently, consider any development on this early alpha and extremely unstable. Expect that this application will drain massive amounts of your device's battery life. Expect this application to download resources infinitely and destroying your data cap. This is confusing stuff.

## TODO

- [x] Break into it's own extension
- [x] Add configuration for manifest from conf.py
   - Partially implemented
- [x] Cache the main webpage and then cache in background, currently app will stall until resources are loaded
   - ~~This is a biggy, the offline feature doesn't fully work because its now required to have a fetch caching strategy in-place~~
   - ~~This currently is working but needs to replace `localStorage` with `indexDB` sql style database. Safari only supports >50MB via indexedDB. This is going to drastically extend the database logic.~~ done
- [x] Add splash screen
   - Chrome automatically does this, Firefox has no support, and Safari requires a splash image for every single device resolution, so we are skipping iOS.
- [x] Miscellaneous edge cases 

We also have a couple offloaded scripts that we could store locally, I think it's mathjax and jquery.


sciencewhiz:
I think you ran an old version of black

Daltz333:
Currently I think there is a mismatch with the artifacts stores in the install worker, and artifacts being searched and stored in the fetch worker.

Daltz333:
Further development has moved to https://github.com/ItayZiv/sphinxext-pwa. This PR will be updated to point at the extension.

PR #1739, opened 2022-02-25
Fix null pointer exception in RB Tank drive article

sciencewhiz:
Can't get Robot Container joysticks during command static initialization
or constructor since commands are instantiated before Robot Container
constructor finishes. Move getting joysticks to command execution


Daltz333 (approved):

SpencerGreene:
I just encountered this same issue in C++; the initialization is in the wrong order, causing a crash. Need to update the C++ example code as well.

PR #1747, opened 2022-02-28
datalog: Use guilabel in a few more places

PeterJohnson:
I didn't use it for window titles... should those be changed as well?

Also fix image reference for CSV export.


TheTripleV:
For buttons, there's mixed use of `:guilabel:text button` and `:guilabel:text`. I think the word "button" should be removed from next to guilabels.

TheTripleV (approved):

sciencewhiz (approved):

PR #1750, opened 2022-03-01
Switch from linkcheckdiff to linkcheck

TheTripleV:
Use linkcheck while the diff one is broken


Daltz333 (approved):

sciencewhiz:
We had issues before with sites blocking our link check that got much better when we changed to only checking diffs.

Daltz333:
I agree, but I'd rather rerun failed checks and false negatives then no checks and positives get through.

PR #1776, opened 2022-03-30
Implement Missing Header Feature

Daltz333:

PR #1777, opened 2022-03-31
Correction of "Implement Missing Header Feature (#1776)"

thenetworkgrinch:
## April Fools Update
* Date corrected for April not March.
Reverts wpilibsuite/frc-docs#1776


TheTripleV:
getMonth is 0 indexed.

PR #1785, opened 2022-04-08
Adding short safety warnings to the Path-Based Tutorial

clockworkthought:
The WPILib RamseteCommand have a lot of very powerful tools now that teams are going back to a genuine autonomous period, but the current Path-Based tutorial is somewhere between 'written for a higher level of familiarity than most teams needing it will be at' and 'mildly suicidal'.   In particular, there is a significantly higher than zero chance that a team copying the code directly will either find their robot's first RamseteCommand to result in their robot zipping backward at maximum power, zipping forwards at 20x their intended speed and distance, or spinning in circles at maximum power.

While I think there's some large benefit to be made, separately, be rewriting the tutorial to target a less experienced team, there are some more immediate benefits to simply not needing to fetch as much drywall spackle.  In particular, some non-obvious problems that this highlights:

- Warned that voltage, speed, and acceleration limits are only applied to trajectory calculation, and do not clamp the power settings to the drive subsystem.
- Highlighted importance of the accuracy of wheel odometry, and encouraged new teams to run the wheel odometry on its own before running a RamseteCommand.
- Encouraged teams developing new code to clamp the motor voltages before ``setVoltage()`` calls in ``tankDriveVolts()``, at least during testing.


Daltz333:
Admonition should be 1-3 sentences. Otherwise they will become trivialized by the reader and ignored. Please break the paragraphs outside of an admonition or shorten significantly.

Daltz333 (dismissed):
Looks good to me. Pinging Tyler for a double check.

source/docs/software/pathplanning/trajectory-tutorial/creating-drive-subsystem.rst
```diff
@@ -250,3 +252,5 @@ Finally, we must include one additional method - a method that allows us to set
       :lineno-start: 43
 
 It is very important to use the ``setVoltage()`` method rather than the ordinary ``set()`` method, as this will automatically compensate for battery "voltage sag" during operation.  Since our feedforward voltages are physically-meaningful (as they are based on measured identification data), this is essential to ensuring their accuracy.
+
+.. warning:: The ``RamseteCommand`` itself does not internally enforce any speed or acceleration limits before providing motor power settings to this method.  During initial code development, teams are strongly encouraged to apply both maximum and minimum bounds on the input variables before passing these values to ``setVoltage()``.
```
calcmogul, line 256: They aren't motor power settings. Power is current times voltage. They are voltage settings.

source/docs/software/pathplanning/trajectory-tutorial/entering-constants.rst
```diff
@@ -60,6 +60,8 @@ Max Trajectory Velocity/Acceleration
 
 We must also decide on a nominal max acceleration and max velocity for the robot during path-following.  The maximum velocity value should be set somewhat below the nominal free-speed of the robot.  Due to the later use of the ``DifferentialDriveVoltageConstraint``, the maximum acceleration value is not extremely crucial.
 
+.. warning:: Max velocity and acceleration, as defined here, are applied only during trajectory generation planning.  They do not limit the ``RamseteCommand`` itself, which may give values to the ``DriveSubsystem`` that can cause the robot to greatly exceed these velocities and accelerations.
```
calcmogul, line 63:
```suggestion
.. warning:: Max velocity and acceleration, as defined here, are applied only during trajectory generation.  They do not limit the ``RamseteCommand`` itself, which may give values to the ``DriveSubsystem`` that can cause the robot to greatly exceed these velocities and accelerations.
```

calcmogul (changes requested):

source/docs/software/pathplanning/trajectory-tutorial/creating-drive-subsystem.rst
```diff
@@ -250,3 +252,5 @@ Finally, we must include one additional method - a method that allows us to set
       :lineno-start: 43
 
 It is very important to use the ``setVoltage()`` method rather than the ordinary ``set()`` method, as this will automatically compensate for battery "voltage sag" during operation.  Since our feedforward voltages are physically-meaningful (as they are based on measured identification data), this is essential to ensuring their accuracy.
+
+.. warning:: ``RamseteCommand`` itself does not internally enforce any speed or acceleration limits before providing motor voltage parameters to this method.  During initial code development, teams are strongly encouraged to apply both maximum and minimum bounds on the input variables before passing these values to ``setVoltage()``.
```
sciencewhiz, line 256: While this is a good technique, to ensure everything is in the right phase, it can also make it so that the robot doesn't seem to follow the path, if the velocities/accelerations used to generate the trajectory are much larger then the achievable velocity/acceleration when limited.

Daltz333: Would you recommend that there is an additional recommendation to ensure that the input velocity and accelerations for generation are also within the bounds of the constraint?

Daltz333: @sciencewhiz

sciencewhiz: yes

sciencewhiz (approved):

PR #1792, opened 2022-04-15
velocity / accel starting values

### LINKED ISSUE #1778: pathweaver: how to compute max velocity/acceleration?
The current docs have the following two pieces:

> Max Velocity: The max speed of the robot for trajectory tracking. This does not need to be the maximum attainable speed of the robot, but just the max speed that you want to run trajectories at.
> 
> Max Acceleration: The max acceleration of the robot for trajectory tracking. This does not need to be the maximum attainable acceleration of the robot, but just the max acceleration that you want to run trajectories at.

Without further research, it's not very clear how one determines those numbers, or even what a sane WAG is. It would be useful to give one or more of the following:

* Examples of typical velocities/accelerations of robots
  * "The typical kitbot robot can expect velocity of X and acceleration of Y" 
* How to compute it yourself, a link to a page on how to compute it, or a link to a calculator for figuring out a theoretical value
* If sysid can tell you, mention that; or if you can compute it from sysid parameters, do that instead

jasondaming:
Closes #1778

### DIFF
--- a/source/docs/software/pathplanning/pathweaver/creating-pathweaver-project.rst
+++ b/source/docs/software/pathplanning/pathweaver/creating-pathweaver-project.rst
@@ -34,9 +34,9 @@ To create a PathWeaver project, click on "Create project" and then fill out the 
 
 **Export Unit:** The units to be used when exporting the paths and waypoints. If you are planning to use WPILib Trajectories, then you should choose :guilabel:`Always Meters`. Choosing :guilabel:`Same as Project` will export in the same units as :guilabel:`Length Unit` above.
 
-**Max Velocity:** The max speed of the robot for trajectory tracking. This does not need to be the maximum attainable speed of the robot, but just the max speed that you want to run trajectories at.
+**Max Velocity:** The max speed of the robot for trajectory tracking.  The kitbot runs at ~10 :math:`ft/sec` which is ~3 :math:`m/sec`.
 
-**Max Acceleration:** The max acceleration of the robot for trajectory tracking. This does not need to be the maximum attainable acceleration of the robot, but just the max acceleration that you want to run trajectories at.
+**Max Acceleration:** The max acceleration of the robot for trajectory tracking. Using the same value for acceleration (3 :math:`m/sec^2`) is a good place to start depending if you don't know your drivetains characteristics.
 
 **Wheel Base:** The distance between the left and right wheels of your robot. This is used to ensure that no wheel on a differential drive will go over the specified max velocity around turns.
 
--- a/source/docs/software/pathplanning/pathweaver/creating-pathweaver-project.rst
+++ b/source/docs/software/pathplanning/pathweaver/creating-pathweaver-project.rst
@@ -36,7 +36,7 @@ To create a PathWeaver project, click on "Create project" and then fill out the 
 
 **Max Velocity:** The max speed of the robot for trajectory tracking.  The kitbot runs at ~10 :math:`ft/sec` which is ~3 :math:`m/sec`.
 
-**Max Acceleration:** The max acceleration of the robot for trajectory tracking. Using the same value for acceleration (3 :math:`m/sec^2`) is a good place to start depending if you don't know your drivetains characteristics.
+**Max Acceleration:** The max acceleration of the robot for trajectory tracking. Using the same value for acceleration (3 :math:`m/sec^2`) is a good place to start if you don't know your drivetain's characteristics.
 
 **Wheel Base:** The distance between the left and right wheels of your robot. This is used to ensure that no wheel on a differential drive will go over the specified max velocity around turns.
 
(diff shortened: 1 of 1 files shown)


source/docs/software/pathplanning/pathweaver/creating-pathweaver-project.rst
```diff
@@ -34,9 +34,9 @@ To create a PathWeaver project, click on "Create project" and then fill out the
 
 **Export Unit:** The units to be used when exporting the paths and waypoints. If you are planning to use WPILib Trajectories, then you should choose :guilabel:`Always Meters`. Choosing :guilabel:`Same as Project` will export in the same units as :guilabel:`Length Unit` above.
 
-**Max Velocity:** The max speed of the robot for trajectory tracking. This does not need to be the maximum attainable speed of the robot, but just the max speed that you want to run trajectories at.
+**Max Velocity:** The max speed of the robot for trajectory tracking.  The kitbot runs at ~10 :math:`ft/sec` which is ~3 :math:`m/sec`.
 
-**Max Acceleration:** The max acceleration of the robot for trajectory tracking. This does not need to be the maximum attainable acceleration of the robot, but just the max acceleration that you want to run trajectories at.
+**Max Acceleration:** The max acceleration of the robot for trajectory tracking. Using the same value for acceleration (3 :math:`m/sec^2`) is a good place to start depending if you don't know your drivetains characteristics.
```
clockworkthought, line 39: Probably want ``drivetrain's``, rather than ``drivetains`` here.  May read more smoothly if you remove 'depending' , as well.

The 'right' numbers here are pretty dependent on total robot weight, height, and center of gravity, and I can definitely say that they're a little too high for taller robots.  The Trajectory Drive tutorial uses the same values, though, without much caveat, so they may just be something the frc-docs group is comfortable with.

jasondaming: We are the frc-docs group and no we don't have a certain "comfort" with these values we just want to give people a number to start with.  We can't make a number that works for every robot.

Thanks for the grammar suggestions I agree.

source/docs/software/pathplanning/pathweaver/creating-pathweaver-project.rst
```diff
@@ -34,9 +34,9 @@ To create a PathWeaver project, click on "Create project" and then fill out the
 
 **Export Unit:** The units to be used when exporting the paths and waypoints. If you are planning to use WPILib Trajectories, then you should choose :guilabel:`Always Meters`. Choosing :guilabel:`Same as Project` will export in the same units as :guilabel:`Length Unit` above.
 
-**Max Velocity:** The max speed of the robot for trajectory tracking. This does not need to be the maximum attainable speed of the robot, but just the max speed that you want to run trajectories at.
+**Max Velocity:** The max speed of the robot for trajectory tracking.  The kitbot runs at ~10 :math:`ft/sec` which is ~3 :math:`m/sec`.
```
sciencewhiz, line 37: Why remove the part about it being for the trajectory not the max speed of the robot. Since you normally want some headroom, I'd suggest 80% of robot max speed.

source/docs/software/pathplanning/pathweaver/creating-pathweaver-project.rst
```diff
@@ -34,9 +34,9 @@ To create a PathWeaver project, click on "Create project" and then fill out the
...
 
-**Max Acceleration:** The max acceleration of the robot for trajectory tracking. This does not need to be the maximum attainable acceleration of the robot, but just the max acceleration that you want to run trajectories at.
+**Max Acceleration:** The max acceleration of the robot for trajectory tracking. Using the same value for acceleration (3 :math:`m/sec^2`) is a good place to start if you don't know your drivetain's characteristics.
```
prateekma, line 39: In my experience, this is way too fast, especially for taller robots. I suggest using somewhere between 1 and 2 m/s/s as a starting point and increase from there if necessary.

prateekma (changes requested):

source/docs/software/pathplanning/pathweaver/creating-pathweaver-project.rst
jasondaming: @calcmogul ?  As I was understanding what he was saying this value should be the maximum and teams should add constraints later.  Either way all of this is kindof bogus until we decide to go all in on either constraints in the program or constraints on RIO (and thus no accurate preview)

calcmogul: > As I was understanding what he was saying this value should be the maximum and teams should add constraints later.

The prevailing wisdom for trajectory optimization solvers is to not provide a WAG for an initial velocity profile. You provide an initial guess at a path, then the rest is by the solver via the user-provided dynamics constraints (in general, dx/dt = f(x) + Bu).

WPILib splines use an initial velocity profile to provide a starting point for the time parameterization. Relying on the profile behavior vs using constraints is a trade off between commonly accepted general practice in motion planning and the quickest thing that gets a robot to move. While I see value in the latter since you don't need a model to use with the dynamics constraints, you need a model anyway for good feedforward. DifferentialDriveVoltageConstraint takes a feedforward object. The feedforward gains can be increased to provide headroom.

> Either way all of this is kindof bogus until we decide to go all in on either constraints in the program or constraints on RIO (and thus no accurate preview)

PathWeaver currently only provides a path preview, and constraints don't change the path shape. A time domain preview needs constraints though.

prateekma (approved):

PR #1799, opened 2022-04-16
Add warning about powering off while deploying

RobinsonZ:
This adds a short warning about powering off the robot mid-deployment, since I've seen and heard reports of that causing somewhat difficult-to-diagnose issues at events. "Deploying Robot Code" seemed like the natural place to put it, but if there's somewhere else it should be added then let me know.

### DIFF
--- a/source/docs/software/vscode-overview/deploying-robot-code.rst
+++ b/source/docs/software/vscode-overview/deploying-robot-code.rst
@@ -3,6 +3,8 @@ Building and Deploying Robot Code
 
 Robot projects must be compiled ("built") and deployed in order to run on the roboRIO.  Since the code is not compiled natively on the robot controller, this is known as "cross-compilation."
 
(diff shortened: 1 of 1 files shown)


source/docs/software/vscode-overview/deploying-robot-code.rst
```diff
@@ -3,6 +3,8 @@ Building and Deploying Robot Code
 
 Robot projects must be compiled ("built") and deployed in order to run on the roboRIO.  Since the code is not compiled natively on the robot controller, this is known as "cross-compilation."
 
+.. warning:: Avoid powering off the robot while deploying robot code. Interrupting the deployment process can corrupt the roboRIO filesystem and prevent your code from working until the roboRIO is :doc:`re-imaged </docs/zero-to-robot/step-3/imaging-your-roborio>`.
```
Daltz333, line 6: I feel like this warning can be a bit further down in the article; right above the point where the deploy actually occurs.

RobinsonZ:
@Daltz333 Is this better? Splitting up the paragraph is a bit suboptimal, I suppose, but the images are quite large and putting it below the last image makes it too easy to ignore in my view.

Daltz333 (dismissed):

Daltz333 (approved):

PR #1807, opened 2022-04-28
Add some python examples

virtuald:
For discussion.

Some questions:

* Should python examples include `import wpilib` statements or omit those? (my vote: omit)
* Should class names include their package name? (my vote: include full package name)
* Should this turn into a giant PR that adds python examples to everything, or is piecemeal ok?
  * Currently, there are a number of examples that don't exist for RobotPy at the moment, so can't use `remoteliteralinclude` yet  (opened issue at https://github.com/robotpy/examples/issues/49)


TheTripleV:
I'd say a single tracking issue with piecemeal PRs.

Daltz333:
> The Java and C++ example snippets don't include the import/includes, so let's elide them for consistency.

We do try to include the includes for C++, as intellisense is particularly problematic there. Assuming VS Code can pick up the right modules, we should be fine.

virtuald:
Thinking about it a bit more, I think that it would be good for a python tab to exist everywhere, but with a common stub even if an example hasn't been implemented yet. Doing it this way will also make future contributions of docs easier -- the submitter would only be obligated to submit Java/C++ docs as per usual, and they could just add this stub for python.

The stub could be something like:

```
.. note:: a python example has not been submitted yet. See [link] to contribute an example
```

TheTripleV (approved):

TheTripleV:
part of #1818

PR #1812, opened 2022-04-29
Add librsvg2-bin as apt dep

Daltz333:
ReadTheDocs no longer installs this for us. Required for PDF builds.

Pending: Waiting for build to verify functionality


Daltz333:
Confirmed that this does fix PDF build https://daltz-private.readthedocs.io/_/downloads/en/fix-pdf/pdf/

TheTripleV (approved):

PR #1821, opened 2022-05-05
Advanced Controls Revamp

gerth2:
Since there's a decent amount of tearup here, starting a draft PR early to gather/document comments.

# Core Goals

 * Revise advanced controls article to introduce concepts in a zero-to-advanced progression.
 * Clearly illustrate the "minimum viable product" techniques, why they do or don't work, and what the "next-step" should be.
 * Emphasize intuition of why certain things should and shouldn't work, feathering in more and more math as students progress toward more advanced techniques.
   * Secondairly, this intuition should help students improve their debugging abilities - knowing what symptoms are likely based in bad controller design or tune, or what's just a software bug or  mechanical/electrical issue.
 * Present feed-forward first, rather than the traditional pedagogy of PID/feedback first.
 * Pictures tell a thousand words

# Technical Changes

Add sufficient js infrastructure to support interactive visualizations of sample mechanisms:
* Flywheel with single ball injected in, plus friction effects and sensor delay
  * Good for teaching bang-bang, feedback-only, and showing how they get close but have downsides
  * Good for teaching kS/kV/kA model for motors, which accounts for defficencies in feedback-only
  * Has potential to show motion profiling basics? Not sure if this is worthwhile though
* Vertical arm, impacted by gravity
  * Good to show how bang bang is completely insufficent for the application
  * Good to show how feedback can help, but will still have limitations
  * Great to show how a non-linear feed-forward can correct for gravity
  * Great for showing motion profiling

Source for the sim for [the flywheel is here](https://trickingrockstothink.com/blog_posts/2019/10/19/tuning_pid.html), and [for the arm is here](https://trickingrockstothink.com/blog_posts/2019/10/26/controls_supp_arm.html)

To support this, we'd need:
* One or more new custom .js files, included into the sphinx build.
  * Tentative - source aggregation and minimization at page build time?
  * Tentative - only include the file in the pages that need it, rather than everywhere?
* New dependency to [Highcharts](https://www.highcharts.com/) 
  * I do have a contact here who blessed our team usage of it for FRC. They'd probably like it if we wrote them another blog post after we get this out, if we do indeed use their product
  * Per their webpage, usage for open-source projects should be free:
    * image
  * For now I just downloaded their latest release and dropped the .js file into the frc-docs repo.

# Structural Changes

* Add additional verbiage to the intro doc to explain the "burning platform" why teams might want to look into controls
* Add links and "signpost" verbiage into each page to make it clear the docs are designed to be a progression from basic to advanced
* Re-organize content to not introduce concepts before they are needed
* Make it clear what each technique is capable of doing, and where the limits are
* Ensure students are encouraged to pattern-match what they see in the docs to what they see on their particular robot
* Add auto-play mp4's with clips from robots that have controls-oriented mechanisms, for maximum inspiration.
* Add walkthrough tuning exercises to encourage students to play with different plant models and controllers, to see how they're _expected_ to behave.

# Status

~~May 5 2022 - Still very much in flight. Content is representative of the direction I want things to take, but far from complete or set in stone. Will update this PR when things are ready for detail review. In the mean time, comment away with concerns - it'll be easier to adjust course now, rather than in 5 months.~~

See comments below for latest status.


Daltz333:
> Pictures tell a thousand words

Absolutely. The more pictures and visualizations, the better.

> One or more new custom .js files, included into the sphinx build.
Tentative - source aggregation and minimization at page build time?
Tentative - only include the file in the pages that need it, rather than everywhere?

I'll look into only deploying the JS during build time, and I believe we have a script that minimizes at build time already. If not, it can be easily arranged. How large is the JavaScript?

Another consider I have is mobile usage. Over half of our viewers are using mobile devices and we want to keep that in mind. 

Lastly, this is an excellent and must-needed contribution. I do want to make the [glossary](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/glossary.html) feature aware, and that it should be used quite liberally. We have hover tips enabled on glossary entries and it makes tracing the vocabulary so much easier.

source/docs/software/advanced-controls/introduction/approaches_to_ctrl_sys_design.rst
```diff
@@ -0,0 +1,81 @@
...
+FeedForward vs. FeedBack
+------------------------
+
+There are two broad categories of strategies in control system design:
+
+**Feed-Forward** refers to the class of algorithms which incorporate knowledge of how the mechanism under control is *expected* to operate. Using this "model" of operation, the control input is chosen to make the mechanism get close to where it should be.
+
+**Feed-Back** refers to the class of algorithms which use sensors to *measure* what a mechanism is doing, and issue corrective commands to move a mechnaism from where it actually is, to where you want it to be.
+
+They are not mutually exclusive - the most advanced techniques will incorporate both of these concepts simultaneously.
```
Daltz333, line 19: > They are not mutually exclusive - the most advanced techniques will incorporate both of these concepts simultaneously.

Perhaps reword this to not showcase "the most advanced" but that this is fairly standard to combine the two.

source/docs/software/advanced-controls/introduction/approaches_to_ctrl_sys_design.rst
```diff
@@ -0,0 +1,81 @@
...
+Because most FRC mechanisms can be manufactured to be have well-defined behavior and can be easily modeled, it's generally encouraged to look into feed-forward techniques first. 
+
+First-Principles - How do you expect your system to behave?
+```````````````````````````````````````````````````````````
+
+The first step in designing a control system is writing down information on how you expect your mechanism to behave.
+
+This step is done by combining one or more concepts you may be familiar with from Physics: drawing free body diagrams of the forces that act on the mechanism, taking measurements of mass and moment of inertia from your CAD models, applying standard models of how DC motors or pneumatic cylinders convert energy into mechanical force and motion, etc.
+
+Models do not have to be extremely accurate to be useful. For example, a very simple model of a motor is that its speed is proportional to the voltage applied. A motor which spins at 3000 RPM at 12v will spin near 1500RPM at 6V. If you know you want your motor to be spinning at 1254 RPM, you can use the proportionality to calculate the amount of voltage to send to the motor.
```
Daltz333, line 33: Proportionality is a scary word. Can you include a brief definition?

source/docs/software/advanced-controls/introduction/approaches_to_ctrl_sys_design.rst
```diff
@@ -0,0 +1,81 @@
...
+
+`ReCalc is an online calculator <https://www.reca.lc/>` which helps take care of some of these physics model calcualtions for common FRC mechanisms.
```
Daltz333, line 35:
```suggestion
`ReCalc is an online calculator <https://www.reca.lc/>`__ which helps take care of some of these physics model calcualtions for common FRC mechanisms.

```

source/docs/software/advanced-controls/introduction/approaches_to_ctrl_sys_design.rst
```diff
@@ -0,0 +1,81 @@
...
+
+System Identification
+`````````````````````
+
+A key way to improve the accuracy of a First-Principles physics model is to perform experiments on it, record data, and use the data to *derive* the constants associated with different parts of the model.
```
Daltz333, line 40: I'm not sure how I feel about the "First-Principles" term.

source/docs/software/advanced-controls/introduction/approaches_to_ctrl_sys_design.rst
```diff
@@ -0,0 +1,81 @@
...
+
+This is useful if CAD models are not perfectly accurate, or parts of the model are not easily predicted (ex: friction in a gearbox).
+
+`FRC's system identification tool (sysid)<link me>` supports some common FRC mechanisms, including drivetrains. It can be configured to load custom code onto a robot, exercise the mechanism and record data, and derive relevant constants to build a more accurate model of the system behavior.
```
Daltz333, line 44:
```suggestion
:doc:`FRC's system identification tool </docs/software/pathplanning/system-identification/introduction>` supports some common FRC mechanisms, including drivetrains. It can be configured to load custom code onto a robot, exercise the mechanism and record data, and derive relevant constants to build a more accurate model of the system behavior.

```

source/docs/software/advanced-controls/introduction/tuning-bang-bang-controller.rst
```diff
@@ -0,0 +1,43 @@
...
+         flywheel_bb = new FlywheelBangBang("flywheel_bb");
+         flywheel_bb.runSim();
+      </script>
+    </div> 
+
+
+Mechanism Walkthrough - Vertical Arm
+````````````````````````````````````
+
+
```
Daltz333, line 34: I'm not a fan of section on top of section without any content. Can we have some sort of sentence that gives a brief explanation on what the topic of this section is about?

Daltz333 (changes requested):
To keep things standard and consistent, we have the following style constraints.

1. Articles with spaces in the name should be in the format `this-is-my-article`. Instead of `this_is_my_article`
2. You should only have one blank line between paragraphs and sections. Remember that source readability is just as important as output. 
3. Section underlines go in the order of `=, -, ^, ~`. If this you are using more than this, then the article should need to be restructured.

gerth2:
Thanks for all the early stuff - will work on incorporating it tonight and continuing to flesh out the actual wording on a lot of sections.

From the technical side:

> How large is the JavaScript?

highcharts.js - ~300kb, already minified.
the new custom stuff - I'm expecting about 30kB when all is said and done. Currently ~1000 lines of js in one file, which really should get split up for comprehensibility.

source/docs/software/advanced-controls/introduction/approaches_to_ctrl_sys_design.rst
gerth2: Updated. I've always heard it to imply "drew a free body diagram and applied F=ma" which presumably lots of high schoolers could do. But probably not the most common implication. I can use other words.

source/docs/software/advanced-controls/introduction/approaches_to_ctrl_sys_design.rst
gerth2: "ratio". Updated.

source/docs/software/advanced-controls/introduction/approaches_to_ctrl_sys_design.rst
gerth2: got em

source/docs/software/advanced-controls/introduction/tuning-bang-bang-controller.rst
gerth2: got it too.

gerth2:
Ok poking the group for a design review on the .js side before I get too far. 

55aecc05 should have most of the functional content required. It's close to being split out how I would like, with much less code duplication, and better class names. Not perfect yet but directionally where I want to go.

Additionally, conf.py has been updated to minify and concatenate all pid-tuning related javascript into a single file. However, as of that node, the files aren't getting concatenated in the right order, meaning the final output isn't actually valid javascript.

As far as getting include order right, the options I have researched:

1. [Custom RST Directive](https://docutils.sourceforge.io/docs/ref/rst/directives.html) which handles some of the boilerplate HTML and imports the specified top-level js as a [module](https://stackoverflow.com/questions/950087/how-do-i-include-a-javascript-file-in-another-javascript-file), which in turn allows the JS itself to use include directives and self-define the class hierarchy. I like this because it feels the most-right, but seems the most intrusive to the whole frc-docs project in general.
2. Implement some custom python classes to describe the javascript structure better and use that to get the order right. I don't like it because it seems like we're re-inventing the wheel with 1, and it causes the JS class hierarchy to get split between two spots (the js itself, and now some new python config).
3. Prefix the js files with 0_, 1_, 2_ or whatever, then alphabetize the list prior to minimization and concatenization. Simple and non-intrusive, and keeps the class hierarchy all in the JS space. But smells hacky as all getout.


Thoughts?

gerth2:
FWIW - `bb60a12ec3e` implements option 3. It's functional, but hacky.

gerth2:
Ok so as of `80921fb`, I'm ready to solicit review on the functionality, code structure, and core content that's listed. I haven't spent time on getting the lint and spell checks passing yet, will fix as soon as we know, directionally, we've got something we want merged.

Thoughts? What do you like, what do you hate, what needs more soak time.... anything and everything welcome!

source/docs/software/advanced-controls/introduction/control-system-basics.rst
```diff
@@ -33,6 +44,15 @@ The figure below shows a system with a hypothetical input and output. Since the
 .. image:: images/control-system-basics-whatisgain.png
    :alt: A system diagram with hypothetical input and output
 
+What is a Model?
+----------------
+
+A *model* of your mechanism is a set of math equations which describe its expected behavior.
+
+They can be built up from first-principles physics equations, and refined by doing experiments on the mechanism.
```
Starlight220, line 52: > first-principles physics equations

I'm not sure I'm familiar with the term?

source/docs/software/advanced-controls/introduction/tuning-pid-controller.rst
```diff
@@ -1,32 +1,229 @@
...
+To counteract this, we introduce a feed-forward term which is also proportional to the cosine of the angle. 
+
+.. math::
+   V_{ff} = K_{cosFF} * cos(\theta_{arm})
+
+:math:`K_{cosFF}` could be calculated if all the mechanical and physical properties of the system are known. However, since a lot of these are hard to model accurately, we will determine it experimentally.
+
+Perform the following:
+
+1. Set :math:`K_p`, :math:`K_i`, :math:`K_d`, and :math:`K_{cosFF}` to zero.
```
Starlight220, line 179:
```suggestion
1. Set :math:`K_p`, :math:`K_i`, :math:`K_d`, and :math:`K_g` to zero.
```

source/docs/software/advanced-controls/introduction/tuning-pid-controller.rst
```diff
@@ -1,32 +1,229 @@
...
+2. Increase and decrease :math:`K_{cosFF}` until the arm can hold its position with as little movement as possible. In this simulation, you'll want to go out to at least four decimal points.
+3. Increase :math:`K_p` until the :term:`output` starts approaches the :term:`setpoint`.
+4. Increase :math:`K_d` as much as possible without introducing jittering in the :term:`system response`. It should help reduce some of the if present.
```
Starlight220, line 182: > It should help reduce some of the if present.

I suspect something is reduced and not present in this sentence.

source/docs/software/advanced-controls/introduction/tuning-pid-controller.rst
```diff
@@ -1,32 +1,229 @@
...
+
+Adjust the setpoint up and down. Now, the arm should exhibit good behavior - quickly and precisely approaching the :term:`setpoint`.
+
+.. raw:: html
+
+   <details>
+     <summary>Tuning Solution</summary>
+
+
+In this particular example, reasonable values for the constants are :math:`K_{cosFF} = 5.92465`, :math:`K_p = 6.0`, and :math:`K_d = 2.0`. These should produce good results at all setpoints.
```
Starlight220, line 192:
```suggestion
In this particular example, reasonable values for the constants are :math:`K_g = 5.92465`, :math:`K_p = 6.0`, and :math:`K_d = 2.0`. These should produce good results at all setpoints.
```

source/docs/software/advanced-controls/introduction/tuning-pid-controller.rst
```diff
@@ -1,32 +1,229 @@
...
-.. note:: The :ref:`SysId toolsuite <docs/software/pathplanning/system-identification/index:System Identification>` can be used to model your system and give accurate Proportional and Derivative values. This is preferred over tuning the controller yourself.
+There are a few ways to mitigate this:
+
+1. Decrease the value of :math:`K_i`, down to zero if possible.
+2. Add logic to reset the integrator term to zero if the :term:`output` is too far from the :term:`setpoint`. Some smart motor controllers implement this with a ``setIZone()`` method.
+3. Cap the integrator at some maximum value. WPILib's ``PIDController`` implements this with the ``setIntegratorRange()`` method.
 
-Actuator Saturation
--------------------
+Actuator Saturation
```
Starlight220, line 217:
```suggestion
Actuator Saturation
```

source/docs/software/advanced-controls/introduction/tuning-pid-controller.rst
```diff
@@ -1,32 +1,229 @@
...
 
 .. math::
    u_{max} &< u \\
    k_{max}(r - x) &< k(r - x) \\
    k_{max} &< k
 
-For the inequality to hold, :math:`k_{max}` must be less than the original value for :math:`k`. This reduced gain is evident in a :term:`system response` when there is a linear change in state instead of an exponential one as it approaches the :term:`reference`. This is due to the :term:`control effort` no longer following a decaying exponential plot. Once the :term:`system` is closer to the :term:`reference`, the controller will stop saturating and produce realistic controller values again.
+For the inequality to hold, :math:`k_{max}` must be less than the original value for :math:`k`. This reduced gain is evident in a :term:`system response` when there is a linear change in state instead of an exponential one as it approaches the :term:`reference`. This is due to the :term:`control effort` no longer following a decaying exponential plot. Once the :term:`system` is closer to the :term:`reference`, the controller will stop saturating and produce realistic controller values again.
```
Starlight220, line 229:
```suggestion
For the inequality to hold, :math:`k_{max}` must be less than the original value for :math:`k`. This reduced gain is evident in a :term:`system response` when there is a linear change in state instead of an exponential one as it approaches the :term:`reference`. This is due to the :term:`control effort` no longer following a decaying exponential plot. Once the :term:`system` is closer to the :term:`reference`, the controller will stop saturating and produce realistic controller values again.
```
assuming there's a newline at EOF

source/docs/software/frc-glossary.rst
```diff
@@ -59,6 +59,9 @@ FRC Glossary
    NetworkTables
       A way to communicate key / value pairs, of data, between programs.
 
+   non-linear
+      In a control system: some system which is not well-modeled by a set of polynomials and their derivatives. For example, in a mechanism where friction forces jump instantaneously between two values (due to changing static/kinetic coefficients of friction), that mechanism can be said to act in a "non-linear" way.
```
Starlight220, line 63: I'm a bit surprised that the term is used in one context, and then explained in the glossary with a totally different example that isn't mentioned anywhere else

source/index.rst
```diff
@@ -13,7 +13,7 @@ FIRST Robotics Competition Control System
 
 Welcome to the *FIRST*\ |reg| Robotics Competition Control System Documentation! This site contains everything you need to know for programming a competition robot!
 
-Community translations can be found in a variety of languages in the bottom-left menu.
+Community translations can be found in a variety of languages in the bottom-left menu.
```
Starlight220, line 16:
```suggestion
Community translations can be found in a variety of languages in the bottom-left menu.
```
I don't think there should be any changes to this file?

Starlight220 (changes requested):

source/docs/software/advanced-controls/introduction/approaches-to-ctrl-sys-design.rst
```diff
@@ -0,0 +1,77 @@
...
+How do you expect your system to behave?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The first step in designing a control system is writing down information on how you expect your mechanism to behave.
+
+This step is done by combining one or more concepts you may be familiar with from Physics: drawing free body diagrams of the forces that act on the mechanism, taking measurements of mass and moment of inertia from your CAD models, applying standard models of how DC motors or pneumatic cylinders convert energy into mechanical force and motion, etc.
+
+Models do not have to be extremely accurate to be useful. For example, a very simple model of a motor is that its speed is proportional to the voltage applied. A motor which spins at 3000 RPM at 12v will spin near 1500RPM at 6V. If you know you want your motor to be spinning at 1254 RPM, you can use the voltage/speed ratio to calculate the amount of voltage to send to the motor.
+
+`ReCalc is an online calculator <https://www.reca.lc/>`__ which helps take care of some of these physics model calcualtions for common FRC mechanisms.
```
Oblarg, line 35: This comes a bit out of nowhere - we need a section here that discusses/derives the common system models that teams will encounter (and that ReCalc/SysId support).

source/docs/software/advanced-controls/introduction/approaches-to-ctrl-sys-design.rst
```diff
@@ -0,0 +1,77 @@
...
+
+
```
Oblarg, line 37: Add a section here on "feedforward and vendor implementations" that discusses the definition and motivation for `kF`, and guides users through accurately converting from SI units to whatever the specific vendor library requires (we will need at least one example per vendor).

source/docs/software/advanced-controls/introduction/approaches-to-ctrl-sys-design.rst
```diff
@@ -0,0 +1,77 @@
...
+System Identification
+^^^^^^^^^^^^^^^^^^^^^
+
+A key way to improve the accuracy of a simple physics model is to perform experiments on the real mechanism, record data, and use the data to *derive* the constants associated with different parts of the model.
+
+This is useful if CAD models are not perfectly accurate, or parts of the model are not easily predicted (ex: friction in a gearbox).
```
Oblarg, line 43: I'd combine this sentence with the previous paragraph, and be specific with reference to equations if possible (if the feedforward equations are introduced in the previous section, you can reference the simple motor equation with and without friction here, and talk about how it's nearly impossible to *predict* `kS` from system parameters

source/docs/software/advanced-controls/introduction/approaches-to-ctrl-sys-design.rst
```diff
@@ -0,0 +1,77 @@
...
+
+:doc:`FRC's system identification tool </docs/software/pathplanning/system-identification/introduction>` supports some common FRC mechanisms, including drivetrains. It can be configured to load custom code onto a robot, exercise the mechanism and record data, and derive relevant constants to build a more accurate model of the system behavior.
+
+
+Combining Feedback and FeedForward
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Even with lots of experiments and study, it is not possible to know every force that will be exerted on a robot's mechanism with great detail. For example, in a flywheel shooter, the timing and exact forces associated with a ball being put through the mechanism are not easy to define accurately. For another example, consider the fact that gearboxes can be greased to reduce friction, but throw off grease over time and get more friction. This is a complex process to model well.
+
+In cases where not everything can be known, incorporating sensors and *feedback* control is useful to account for inaccuracies in the model of system behavior.
```
Oblarg, line 53: Be a bit more forceful than "this is useful for."  Feedback is *necessary* to account for inaccuracies in the system model - it just might not be necessary to do that in the context of any given control task.

source/docs/software/advanced-controls/introduction/approaches-to-ctrl-sys-design.rst
```diff
@@ -0,0 +1,77 @@
...
+
+
+Incorporating FeedBack
+----------------------
+
+Let's say we are controlling a DC brushed motor. With just a mathematical model and knowledge of all the current states of the system (i.e., angular velocity), we can predict all future states given the future voltage inputs. Why then do we need feedback control? If the system is disturbed in any way that isn't modeled by our equations, like a load was applied, or voltage sag in the rest of the circuit caused the commanded voltage to not match the applied voltage, the angular velocity of the motor will deviate from the model over time.
```
Oblarg, line 59: refer to system equations here

source/docs/software/advanced-controls/introduction/control-system-basics.rst
```diff
@@ -1,10 +1,21 @@
...
 
-.. note:: This article is taken out of `Controls Engineering in FRC <https://file.tavsys.net/control/controls-engineering-in-frc.pdf>`__ by Tyler Veness with permission.
+.. note:: This article includes sections of `Controls Engineering in FRC <https://file.tavsys.net/control/controls-engineering-in-frc.pdf>`__ by Tyler Veness with permission.
+
+The Need for Control Systems
+----------------------------
 
 Control systems are all around us and we interact with them daily. A small list of ones you may have seen includes heaters and air conditioners with thermostats, cruise control and the anti-lock braking system (ABS) on automobiles, and fan speed modulation on modern laptops. Control systems monitor or control the behavior of systems like these and may consist of humans controlling them directly (manual control), or of only machines (automatic control).
 
+The thing that is common to all of these things: They all have a mechanism which does useful work, but cannot be *directly* commanded to the state that is desired.
```
Oblarg, line 11: Remove "the thing that is."  Reads better as just "Common to all of these things: a mechanism which does useful work, but cannot be *directly* commanded to the state that is desired.

source/docs/software/advanced-controls/introduction/control-system-basics.rst
Oblarg: "first principles" is another term for a set of axioms - that is, in the sense of a single set of starting principles that is sufficient to derive the whole theory.

source/docs/software/advanced-controls/introduction/control-system-basics.rst
```diff
@@ -33,6 +44,15 @@ The figure below shows a system with a hypothetical input and output. Since the
 .. image:: images/control-system-basics-whatisgain.png
    :alt: A system diagram with hypothetical input and output
 
+What is a Model?
+----------------
+
+A *model* of your mechanism is a set of math equations which describe its expected behavior.
```
Oblarg, line 50: I'd say "a mathematical description of its behavior."  You can go more in depth, and talk about how what this really means is some sort of law relating its *inputs* and *outputs.*

source/docs/software/advanced-controls/introduction/introduction-to-pid.rst
```diff
@@ -3,6 +3,11 @@
 Introduction to PID
 ===================
 
+This page provides a math-focused approach to explaining a PID controller. In parallel, in :ref:`a video from WPI is available <docs/software/advanced-controls/introduction/pid-video:PID Introduction Video by WPI>` if you prefer videos over reading.
+
+What is a PID Controller?
+-------------------------
+
 The PID controller is a commonly used feedback controller consisting of proportional, integral, and derivative terms, hence the name. This article will build up the definition of a PID controller term by term while trying to provide some intuition for how each of them behaves.
```
Oblarg, line 11: "feedback controller" should be a hyperlink to the earlier section explaining the motivation behind feedback control

source/docs/software/advanced-controls/introduction/tuning-bang-bang-controller.rst
```diff
@@ -0,0 +1,52 @@
...
+A "Bang-Bang" controller is a simple controller which applies a binary (present/not-present) force to a mechanism to try to get it closer to a setpoint. 
+
+:ref:`WPILib provides an implementation <docs/software/advanced-controls/controllers/bang-bang:Bang-Bang Control with BangBangController>` for bang-bang control as well. This page includes a more in-depth description.
+
+
+General Tuning Techniques
+-------------------------
+
+Bang-Bang controllers do not have any parameters to tune. They either push the mechanism toward the setpoint from the "too-low" side, or allow it to drift back toward the setpoint from the "too-high" side.
```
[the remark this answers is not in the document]
Oblarg: yeah, we can distinguish between asymmetric bang-bang and symmetric bang-bang here.  Asymmetric is way more usable because systems that want meaningful control effort in both directions generally do not want an ugly response discontinuity at 0.

source/docs/software/advanced-controls/introduction/tuning-bang-bang-controller.rst
```diff
@@ -0,0 +1,52 @@
...
+         flywheel_bb = new FlywheelBangBang("flywheel_bb");
+         flywheel_bb.runSim();
+      </script>
+    </div> 
+
+A Bang-Bang controller is most well known for its simplicity - there is nothing to tune!
+
+Try adjusting the setpoint up and down. You should see that for almost all values, the output converges to be somewhat near the setpoint.
+
+Note it's not perfect: Sensor delay in the system prevents the control effort from turning on and off at exactly the right time. In turn, this causes a cycle of "overshoot" and "undershoot", as the output repeatedly goes above and below the setpoint.
```
Oblarg, line 39: Not necessarily just sensor delay - *any* loop delay will cause the same problem.

source/docs/software/advanced-controls/introduction/tuning-bang-bang-controller.rst
```diff
@@ -0,0 +1,52 @@
...
+
+Common Issues
+-------------
+
+The imprecision observed is one common issue with bang-bang control.
```
Oblarg, line 44: How about "The guarantee of an oscillation at steady-state," since the dynamics causing that were just explained above?

source/docs/software/advanced-controls/introduction/tuning-bang-bang-controller.rst
```diff
@@ -0,0 +1,52 @@
...
+
+Additionally, the fact the control effort turns on and off quickly, zooming from zero to full, can cause problems. The cycles of rapidly applying and removing forces can loosen bolts and joints, and put a lot of stress on gearboxes.
+
+The abrupt changes in control effort can cause abrupt changes in current draw. Some motor controllers may not last for an extended period of time.
```
Oblarg, line 48: "This may stress motor control hardware, and cause eventual damage or failure."

source/docs/software/advanced-controls/introduction/tuning-bang-bang-controller.rst
```diff
@@ -0,0 +1,52 @@
...
+
+Finally, this technique only works for mechanisms that accelerate relatively slowly, in comparison to the amount of control effort applied. For mechanisms like flywheels, or drivetrain control for simple autonomous, it may be sufficient.
```
Oblarg, line 50: Per above conversation - this deserves more explicit discussion, and should be tied in to the notion of an asymmetric controller (which is what WPILib supports)

source/docs/software/advanced-controls/introduction/tuning-pid-controller.rst
```diff
@@ -1,32 +1,229 @@
...
+         flywheel_pid = new FlywheelPIDF("flywheel_pid");  
+         flywheel_pid.runSim(); 
+      </script> 
+    </div>
+
+Mechanism Description
+~~~~~~~~~~~~~~~~~~~~~
+The "Flywheel" is nothing more than a circular mass, attached to a gearbox, and driven by a motor. They're commonly used to propel game pieces through the air, into a target. 
+
+In general: the more voltage that is applied to the motor, the faster the flywheel will spin. Once voltage is removed, friction slowly decreases the spinning until the flywheel stops.
```
Oblarg, line 40: link to section with system equations - may not be appropriate to pull them in here verbatim but they should be available for reference

source/docs/software/advanced-controls/introduction/tuning-pid-controller.rst
```diff
@@ -1,32 +1,229 @@
...
+
+To consistently launch a gamepiece, a good first step is to make sure it is spinning at a particular speed before putting a gamepiece into it. 
+
+This design drives the controls goal we will use in this example: Put the correct amount of voltage into the motor to get the flywheel to a certain speed, and then keep it there.
+
+As a test, a gamepiece is injected into the flywheel about halfway through the simulation.
```
Oblarg, line 46: this needs a footnote explaining the actual loading simulation for curious students

source/docs/software/advanced-controls/introduction/tuning-pid-controller.rst
```diff
@@ -1,32 +1,229 @@
...
+.. raw:: html
+
+   </details> <br>
+
+This is a case where feedback control alone is insufficient to achieve good behavior with the system.
+
+Step 2: Feed-Forward, then FeedBack
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The core reason for this is that gravity pulls on the arm in a :term:`non-linear` fashion. That is to say, the amount of :term:`torque` that gravity exerts on our arm is proportional to the *cosine* of the current angle.
```
Oblarg, line 168: This isn't quite right - even if gravity pulled on the arm in a linear fashion (e.g. an elevator) you still can't really tune with feedback only.  The reason is that we're missing the system model entirely.  Nonlinearity is incidental here.

source/docs/software/advanced-controls/introduction/tuning-pid-controller.rst
```diff
@@ -1,32 +1,229 @@
...
+
+To counteract this, we introduce a feed-forward term which is also proportional to the cosine of the angle. 
+
+.. math::
+   V_{ff} = K_{cosFF} * cos(\theta_{arm})
+
+:math:`K_{cosFF}` could be calculated if all the mechanical and physical properties of the system are known. However, since a lot of these are hard to model accurately, we will determine it experimentally.
```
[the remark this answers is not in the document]
Oblarg: Link to earlier page that describes these constants and SysId/Recalc

source/docs/software/advanced-controls/introduction/tuning-pid-controller.rst
```diff
@@ -1,32 +1,229 @@
...
+
+Integral Term Windup
+^^^^^^^^^^^^^^^^^^^^
 
 Beware that if :math:`K_i` is too large, integral windup can occur. Following a large change in :term:`setpoint`, the integral term can accumulate an error larger than the maximal :term:`control input`. As a result, the system overshoots and continues to increase until this accumulated error is unwound.
 
-.. note:: The :ref:`SysId toolsuite <docs/software/pathplanning/system-identification/index:System Identification>` can be used to model your system and give accurate Proportional and Derivative values. This is preferred over tuning the controller yourself.
+There are a few ways to mitigate this:
+
+1. Decrease the value of :math:`K_i`, down to zero if possible.
```
Oblarg, line 213: Add: "Most mechanisms in FRC do not require any integral control, and systems that seem to require integral control to respond well probably have an inaccurate feedforward model."

source/docs/software/frc-glossary.rst
Oblarg: We should probably remove this.  Nonlinearity is pretty complicated and probably not within-scope; we should keep all of this implicitly within the realm of linear control theory.  We can talk about linearization in a separate article maybe, but that seems to be getting towards the realm of Tyler's book so that might be out of scope

Oblarg: It's worth noting that the meaning of the term "linear" in the context of controls is "algebraically homomorphic," which is probably not something that *most* students are quite ready to deal with in high school.  The term is unfortunately quite overloaded.

source/docs/software/advanced-controls/introduction/control-system-basics.rst
gerth2: It's definitely loaded and, while I"ve heard it used a lot in the context to describe "basic hs physics equations", it probably doesn't mean the same to everyone. Dalton had asked it removed earlier and I pulled out all the ones I could find, but missed this one.

source/docs/software/frc-glossary.rst
gerth2: Definitely ok with removing it. "Nonlinearity" here was brought up as a proxy for saying 'This is purposefully and example where PID by itself performs poorly". Rather than getting technical, I'll reword to just express that notion, and maybe point to a section of the textbook so a student who likes depth-first knowledge searches can get a taste as to "why" it would perform poorly.

source/docs/software/advanced-controls/introduction/tuning-pid-controller.rst
```diff
@@ -1,32 +1,229 @@
 Tuning a PID Controller
 =======================
 
-These steps apply to position PID controllers. Velocity PID controllers typically don't need :math:`K_d`.
+As seen in :ref:`the introduction to PID controller <docs/software/advanced-controls/introduction/introduction-to-pid:Introduction to PID>`, a PID controller has three *tuned* constants. The numeric values for these constants must be picked carefully for the specific mechanism under control. These constants will generally have different values on different robots. 
 
-1. Set :math:`K_p`, :math:`K_i`, and :math:`K_d` to zero.
+There are multiple methods for determining the values for these constants on your particular mechanism.
+
+The :ref:`SysId toolsuite <docs/software/pathplanning/system-identification/index:System Identification>` can be used to model your system and give accurate Proportional and Derivative values. This is preferred for supported mechanism types.
```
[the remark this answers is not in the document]
gerth2: I'm fine with either. the $K_p$ style was copy-pasted from the other Introduction to PID article.

Alternatively, I could drop a note near the top that explains K_p == kP, one is software, the other is fancy schmancy math notation.

source/docs/software/advanced-controls/introduction/tuning-pid-controller.rst
```diff
@@ -1,32 +1,229 @@
...
+We will first attempt to tune the flywheel using only the feedback terms :math:`K_p`, :math:`K_i`, and :math:`K_d`. 
+
+Perform the following:
+
+1. Set :math:`K_p`, :math:`K_i`, :math:`K_d`, and :math:`K_v` to zero.
 2. Increase :math:`K_p` until the :term:`output` starts to oscillate around the :term:`setpoint`.
 3. Increase :math:`K_d` as much as possible without introducing jittering in the :term:`system response`.
+4. *In some cases*, increase :math:`K_i` if :term:`output` gets "stuck" before converging to the :term:`setpoint`.
+
+.. important:: Adding an integral gain to the :term:`controller` is often a sub-optimal way to eliminate :term:`steady-state error`. A better approach would be to tune it with an integrator added to the :term:`plant`, but this requires a :term:`model`.
```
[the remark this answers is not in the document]
gerth2: Definitely needs rephrased. Communication intent is "In FRC, a better approach is to calculate a feed-forward term from assumptions about how the system is supposed to behave."

source/docs/software/advanced-controls/introduction/tuning-pid-controller.rst
calcmogul: Gravity for an elevator is actually affine, not linear.

source/docs/software/advanced-controls/introduction/control-system-basics.rst
```diff
@@ -33,6 +44,15 @@ The figure below shows a system with a hypothetical input and output. Since the
...
+
+They can be built up from first-principles physics equations, and refined by doing experiments on the mechanism.
+
+Models are often simple algebra equations. They can also include some linear algebra, matricies, and differential equations. WPILib provides a number of classes to help simplify the more complex math.
```
Daltz333, line 54: These three sentences could use some transitional vocabulary. They are sort of "shoved" together if that makes any sense.

source/docs/software/advanced-controls/introduction/tuning-bang-bang-controller.rst
```diff
@@ -0,0 +1,52 @@
...
+Try adjusting the setpoint up and down. You should see that for almost all values, the output converges to be somewhat near the setpoint.
+
+Note it's not perfect: Sensor delay in the system prevents the control effort from turning on and off at exactly the right time. In turn, this causes a cycle of "overshoot" and "undershoot", as the output repeatedly goes above and below the setpoint.
+
+Common Issues
+-------------
+
+The imprecision observed is one common issue with bang-bang control.
+
+Additionally, the fact the control effort turns on and off quickly, zooming from zero to full, can cause problems. The cycles of rapidly applying and removing forces can loosen bolts and joints, and put a lot of stress on gearboxes.
```
Daltz333, line 46: > Additionally, the fact the control effort turns on and off quickly, zooming from zero to full, can cause problems.

Can this sentence be rephrased, there's a lot of fragments here delimited by a comma.

source/docs/software/advanced-controls/introduction/tuning-pid-controller.rst
```diff
@@ -1,28 +1,225 @@
...
+      </div>
+      <script>
+         flywheel_pid = new FlywheelPIDF("flywheel_pid");  
+         flywheel_pid.runSim(); 
+      </script> 
+    </div>
+
+Mechanism Description
+~~~~~~~~~~~~~~~~~~~~~
+The "Flywheel" is nothing more than a circular mass, attached to a gearbox, and driven by a motor. They're commonly used to propel game pieces through the air, into a target.
```
Daltz333, line 38: Should we include an FRC picture example of a flywheel?

source/docs/software/advanced-controls/introduction/tuning-pid-controller.rst
Daltz333: Yep, the above suggestion should probably be an ``.. important::`` admonition.

source/docs/software/frc-glossary.rst
```diff
@@ -68,5 +71,8 @@ FRC Glossary
    teleop
       The second phase of each match is called the Teleoperated Period (teleop) and consists of drivers controlling their robots.
 
+   torque
```
Daltz333, line 74: Are we actually linking to this anywhere in the docs?

source/docs/software/advanced-controls/introduction/tuning-bang-bang-controller.rst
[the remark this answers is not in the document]
gerth2: For the sake of introduction.... maybe just ignore the difference and only present the Asymetric version.... and ignore the fact that the command voltage can be tuned?

A lot of this page is just to show "look, you can get kinda close by barely thinking at all, but it might not work because of XYZ. If it doesn't work, keep going".

source/docs/software/advanced-controls/introduction/tuning-pid-controller.rst
```diff
@@ -1,32 +1,229 @@
...
+      </div>
+      <script>
+         arm_pidf = new VerticalArmPIDF("arm_pid"); 
+         arm_pidf.runSim(); 
+      </script> 
+    </div>
+
+Mechanism Description
+~~~~~~~~~~~~~~~~~~~~~
+The "Vertical Arm" is a mass on a stick, moved up and down by a gearbox, and driven by a motor. They're commonly used to lift gamepieces from the ground, and up higher to place and score them.
```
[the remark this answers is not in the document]
gerth2: "Up from the ground to a scoring position"?

Lots of ideas built into one sentence for sure. The rough goal is just to say something generic enough to never be wrong, but specific enough to get the reader to connect the situation in the walkthrough to some unknown, future game where they use this mechanism.

source/docs/software/advanced-controls/introduction/tuning-pid-controller.rst
gerth2: I made it more hand wavey so it should be more meaningful to the lay audience.

source/docs/software/advanced-controls/introduction/tuning-bang-bang-controller.rst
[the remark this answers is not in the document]
Oblarg: That's fair.  Just word it carefully so that the article doesn't lie (except by omission, and with links where things are omitted).

source/docs/software/frc-glossary.rst
gerth2: Yes, we are now

gerth2:
Folks, thank you all for the comments - all were valuable, and I'm in the process of chunking them out. Will also work to get CI to pass as much as possible as soon as content updates are in plcae.

@Daltz333 there was a comment on discord about "beta testing" these updates through a CD post. Thoughts on merging this PR to latest, opening a second issue for "round-2" updates, collecting the feedback, and executing on it?

source/docs/software/advanced-controls/introduction/tuning-bang-bang-controller.rst
gerth2: Added additional examples of delay sources

source/docs/software/advanced-controls/introduction/tuning-bang-bang-controller.rst
gerth2: like

source/docs/software/advanced-controls/introduction/tuning-pid-controller.rst
gerth2: Option 1:

gerth2: Option 2:

https://trickingrockstothink.com/assets/img/shooter_wheel_1640.png

source/docs/software/advanced-controls/introduction/approaches-to-ctrl-sys-design.rst
gerth2: Rather than including it here.... can I punch in an issue to do this separately? I like the idea and think it's needed, but also wanna keep it kinda independent (both from a pedagogy perspective, and I assume vendor implementations will indeed evolve over the years)

gerth2:
Taking some notes on updates for myself:

1) Add (somwhere) [this](https://discord.com/channels/176186766946992128/368993897495527424/976884015397228614) diagram with notes to explain terminology.
2) Tackle all meaningful comments from [the CD thread](https://www.chiefdelphi.com/t/wpi-docs-interactive-pid-tuning-example/411813/18).

Daltz333 (commented):
Could we include examples of mechanisms from actual FRC robots? I can see about reaching out on discord to see if we have permission to relicense and give them a shout out in a caption.

.github/workflows/CI.yml
```diff
@@ -69,6 +69,9 @@ jobs:
     - uses: actions/setup-python@v2
       with:
         python-version: 3.9
+    - name: Install Dependencies
+      run: |
+        pip install -r source/requirements.txt
```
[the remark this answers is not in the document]
gerth2: I did do it intentionally. `jsmin` was showing up as not-found in CI without it.

source/docs/software/advanced-controls/introduction/tuning-flywheel.rst
```diff
@@ -0,0 +1,238 @@
...
+Finally, this technique only works for mechanisms that accelerate relatively slowly. A more in-depth discussion of the details :ref:`can be found here <docs/software/advanced-controls/controllers/bang-bang:Bang-Bang Control with BangBangController>`.
+
+PID Control
+~~~~~~~~~~~
+
+As seen in :ref:`the introduction to PID <docs/software/advanced-controls/introduction/introduction-to-pid:Introduction to PID>`, a PID controller has three *tuned* constants. The numeric values for these constants must be picked carefully for the specific mechanism under control. These constants will generally have different values on different robots.
+
+There are multiple methods for determining the values for these constants on your particular mechanism.
+
+The :ref:`SysId toolsuite <docs/software/pathplanning/system-identification/index:System Identification>` can be used to model your system and give accurate Proportional and Derivative values. This is preferred for supported mechanism types.
```
Oblarg, line 102: Mention recalc, too

source/docs/software/advanced-controls/introduction/tuning-flywheel.rst
```diff
@@ -0,0 +1,238 @@
...
+
+.. raw:: html
+
+   </details> <br>
+
+
+Flywheel Tuning Step 2: Feedforward, then Feedback
+--------------------------------------------------
+
+Tuning with only feedback can produce reasonable results in many cases. However, there is an easier way. Rather than starting with feedback, start by calibrating an appropriate feedforward value.
```
Oblarg, line 173: Per above, "many cases" should be more specific.  It works well in systems where no control signal is required to maintain steady state at the setpoint, and poorly elsewhere.

source/_static/js/pid-tune/1-base-sim.js
```diff
@@ -0,0 +1,180 @@
...
+
+        this.voltsChart.addSeries({name: "Control Effort", color: '#00BB00'});
+        this.valsChart.addSeries({name: "Output", color: '#FF0000'});
+        this.valsChart.addSeries({name: "Setpoint", color: '#0000FF'});
+
+        this.voltsChart.xAxis[0].addPlotLine({color: '#BBBB00',width: 2, value: 0.0, id:"curTime"})
+        this.valsChart.xAxis[0].addPlotLine({color: '#BBBB00',width: 2,value: 0.0, id:"curTime"})
+
+        this.voltsChart.yAxis[0].setTitle({ text : "Volts"});
+        this.valsChart.yAxis[0].setTitle({ text :stateUnits });
```
Oblarg, line 21: Plot y-axes should have a fixed scale depending on the mechanism being simulated, rather than being inferred from the calculated output values.

source/docs/software/advanced-controls/introduction/tuning-flywheel.rst
```diff
@@ -0,0 +1,238 @@
...
+
+.. note:: When "increasing" a value, multiply it by two until the expected effect is observed. Similarly, when "decreasing" a value, divide by two. Once you find the point where the expected effect starts or stops, switch to "bumping" the value up and down by ~10% until the behavior is good enough.
+
+.. raw:: html
+
+   <details>
+     <summary>Tuning Solution</summary><br>
+
+
+In this particular example, for a setpoint of 1000, values of :math:`K_p = 0.13`, :math:`K_i = 0.0`, and :math:`K_d = 0.002` will produce somewhat reasonable results. It will get better or worse as you change the setpoint.
```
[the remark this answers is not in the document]
gerth2: > (other than, potentially, integral gain) 

Yup. The interpretation here is that the I term is dynamically "learning" about gravity, which really isn't the right way to go about this.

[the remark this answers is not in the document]
Oblarg: In the case of a flywheel it's even worse - the integral term is learning the EMF of the motor (which varies with the setpoint).  This murders your settling times.

source/docs/software/advanced-controls/introduction/tuning-flywheel.rst
Starlight220: Maybe give examples (drive/turret position vs elevators and velocity systems)

Oblarg:
Per discussion on discord, I think this needs one substantial expansion to serve as a complete intro: a turret position controller.

Right now we have: a complicated system that requires both feedforward and feedback for effective control (arm under gravity) and a simple system that works well with feedforward-only but not with feedback-only (flywheel velocity).  We are missing a simple system that works well with feedback-only, for which feedforward is unnecessary (turret position).

This is an important case to cover, because it matches the case presented in many other PID tutorials and thus can serve as a clean point of translation between prior team knowledge and the "new" approach we're suggesting here.  It also (given suitable expansion of the other articles) gives rise to a pleasing symmetry in structure:

Approach 1: Feedforward only

* Works well for flywheel
* Works poorly for turret
* Works poorly for arm

Approach 2: Feedback only

* Works poorly for flywheel
* Works well for turret
* Works poorly for arm

Approach 3: Feedforward and feedback

* Works well for all three systems

This also potentially addresses concerns raised in the CD thread about the sequencing of the control strategies (we may still want to reorder them, but we can cross that bridge when we get there).

sciencewhiz:
I get high CPU usage on the pages with the simulations. Looks like it's pegging one core.

gerth2:
> I get high CPU usage on the pages with the simulations. Looks like it's pegging one core.

Interesting. I know mine goes up too, but not quite to max. OS and Browser?

The animation is probably what is causing it, but we are using the "request animation frame" mechanism, so the browser should be able to scale up and down based on what it thinks is appropriate. 

I am out of town currently but should be able to get back to this in a week or two, I could compare execution traces from my setup to see where the bottleneck is

sciencewhiz:
It happened in both firefox 100 and chrome 102 on Windows 10. Here's a picture from task manager showing a noticeable jump on 2 cores, and less jumps on others.

image

gerth2:
@sciencewhiz I'm having issues replicating the cpu usage issues.

image


image

On my setup I'm seeing only 20% utilization, which I think is about reasonable based on the `requestAnimationFrame()` mechanism. Actually, I'm not sure anything can be done about it, as the browser is fully in control of how often to invoke the animation routine (which in turn should enable it to tradeoff between CPU load and framerate).

sciencewhiz:
I tried it in a Windows 10 VM and couldn't reproduce it, and also couldn't reproduce it in Ubuntu. Not sure why it affects multiple browsers on this computer, though.

source/docs/software/advanced-controls/introduction/approaches-to-ctrl-sys-design.rst
```diff
@@ -0,0 +1,70 @@
...
+**Feedforward** refers to the class of algorithms which incorporate knowledge of how the mechanism under control is *expected* to operate. Using this "model" of operation, the control input is chosen to make the mechanism get close to where it should be.
+
+**Feedback** refers to the class of algorithms which use sensors to *measure* what a mechanism is doing, and issue corrective commands to move a mechanism from where it actually is, to where you want it to be.
+
+They are not mutually exclusive - many techniques will use these concepts simultaneously.
+
+Start with Feedforward
+-----------------------
+
+Because most FRC mechanisms can be manufactured to be have well-defined behavior and can be easily modeled, it's generally encouraged to look into feedforward techniques first.
```
sciencewhiz, line 24: At the risk of starting a holy war, I'm not sure I agree with this. A pure feedforward control scheme won't work with unmodeled disturbances, which there are many. If I was going to suggest either feedforward or feedback control, I'd have to say feedback control would be better for the vast majority of teams. Obviously feedforward + feedback is the best.

source/docs/software/advanced-controls/introduction/approaches-to-ctrl-sys-design.rst
```diff
@@ -0,0 +1,70 @@
...
+How do you expect your system to behave?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The first step in designing a control system is writing down information on how you expect your mechanism to behave.
+
+This step is done by combining one or more concepts you may be familiar with from physics: drawing free body diagrams of the forces that act on the mechanism, taking measurements of mass and moment of inertia from your CAD models, applying standard equations of how DC motors or pneumatic cylinders convert energy into mechanical force and motion, etc.
+
+The act of creating these mathematical descriptions is called *modeling* your system's behavior. The resulting set of equations are the *model* of how you expect the system to behave.
+
+Note that models do not have to be perfectly accurate to be useful. As we will see in later tuning exercises, even using a simple model of a mechanism can make the tuning effort much simpler.
```
sciencewhiz, line 35: This implies the necessity of feedforward + feedback, which seems to contradict with above

source/docs/software/advanced-controls/introduction/approaches-to-ctrl-sys-design.rst
```diff
@@ -0,0 +1,70 @@
...
+
+To combat this undesired effect, we can take measurements of the system and the environment to detect this deviation and account for it. For example, we could measure the current position and estimate an angular velocity from it. We can then give the motor corrective commands as well as steer our model back to reality. This feedback allows us to account for uncertainty and be robust to it.
+
+Feedback-Only Techniques
+------------------------
+
+In many controls textbooks, you may see a set of techniques which rely on feedback control only. These are very common in industry, and works well in many cases, especially when the underlying system behavior is not easy to model.
+
+For most FRC mechanisms, these feedback-only techniques may produce good results in some cases. However, there will be limits to this success. For most FRC use cases, it is recommended to include feedforward.
```
sciencewhiz, line 70: With the exception of flywheels and one complicated multi-jointed arm, feedback only has been sufficient in my experience. This seems to undersell it's usefulness.

source/docs/software/advanced-controls/introduction/approaches-to-ctrl-sys-design.rst
Oblarg: It depends critically on what type of system you are trying to control.  See my post above for a more thorough discussion.

source/docs/software/advanced-controls/introduction/approaches-to-ctrl-sys-design.rst
Oblarg: Even when using just feedback, modeling the system is a crucial step if you want to understand what approximate gains you will need and why.  Tuning without *some* sense for how the system behaves is effectively impossible.

source/docs/software/advanced-controls/introduction/approaches-to-ctrl-sys-design.rst
Oblarg: It works if you throw enough motor power at it, but it's really not that hard to end up in a situation where the mechanism you have does not work without feedforward.

calcmogul: However, NEOs and Falcons make a lot of subsystems overpowered nowadays.

Oblarg: The turret demo is currently awfully overpowered.  This proves the point well for "feedback only sometimes works," though we might want to rescale the control effort plot...

Oblarg:
This seems good enough to merge now, to my eyes; it might need a couple of typographical changes per the Chief Delphi thread, but I don't think there are any major errors.  I'm gonna go ahead and approve.

Oblarg (approved):

source/docs/software/advanced-controls/introduction/common-control-issues.rst
```diff
@@ -0,0 +1,35 @@
...
+3. Cap the integrator at some maximum value. WPILib's ``PIDController`` implements this with the ``setIntegratorRange()`` method.
+
+.. important:: Most mechanisms in FRC do not require any integral control, and systems that seem to require integral control to respond well probably have an inaccurate feedforward model.
+
+Voltage Sag
+-----------
+
+When we operate mechanisms on our robot, we draw current from its battery.  This causes the available "bus voltage" that all the robot mechanisms operate off of to drop.  This means that the performance of our mechanisms will vary depending on the loading and action of the robot - this is not ideal.
+
+To fix this, most voltage controllers offer a "voltage compensation" setting for their internal control loops that keep the output voltage of the control loops constant despite changes in the bus voltage.  The WPILib ``SpeedController`` class offers a ``setVoltage`` method can do the same thing if the control loops are being run on the RIO (provided you call it every robot loop iteration).
```
Daltz333, line 24:
```suggestion
To fix this, most voltage controllers offer a "voltage compensation" setting for their internal control loops that keep the output voltage of the control loops constant despite changes in the bus voltage.  The WPILib ``MotorController`` class offers a ``setVoltage`` method can do the same thing if the control loops are being run on the RIO (provided you call it every robot loop iteration).
```

source/docs/software/advanced-controls/introduction/control-system-basics.rst
```diff
@@ -33,6 +44,15 @@ The figure below shows a system with a hypothetical input and output. Since the
...
    :alt: A system diagram with hypothetical input and output
 
+What is a Model?
+----------------
+
+A *model* of your mechanism is a mathematical description of its behavior. Specifically, this mathematical description must define the mechanism's inputs and outputs, and how the output values change over time as a function of its input values.
+
+The mathematical description is often just simple algebra equations. It can also include some linear algebra, matrices, and differential equations. WPILib provides a number of classes to help simplify the more complex math.
+
+`Classical Mechanics <https://en.wikipedia.org/wiki/Classical_mechanics>` defines many of the equations used to build up models of system behavior. Many of the values inside those equations can be determined by doing experiments on the mechanism.
```
Daltz333, line 54:
```suggestion
`Classical Mechanics <https://en.wikipedia.org/wiki/Classical_mechanics>`__ defines many of the equations used to build up models of system behavior. Many of the values inside those equations can be determined by doing experiments on the mechanism.
```

Daltz333 (changes requested):
Additionally, I'd like to see most of the Wikipedia pages links/definitions recreated in the glossary section (for the tooltips), with the glossary definition linking to Wikipedia instead.

Daltz333 (changes requested):
I think I'm seeing unrelated changes to the `conf.py` and GitHub actions workflow.

Daltz333 (approved):

TheTripleV:
To only include the js on pages that need, this code could be wrapped in a directive that does it when it's processed.

PR #1823, opened 2022-05-08
Update physics sim descriptions

sciencewhiz:
clarify how gravity is treated differently between elevator and arm


sciencewhiz:
broken links are unrelated and fixed in #1824

Daltz333 (approved):

PR #1808, opened 2022-04-28
[command] Split "convenience features" article into separate articles about builtins and decorators

Starlight220:
The docs haven't been updated since the 2020 command rewrite, and the docs should be updated to recommend the modern way of using command-based.

Still a lot of progress needed; we also need to decide on what the idioms and recommended structure should be.

PR #1828, opened 2022-05-15
Add discussion of command organization

RobinsonZ:
Definitely WIP, but throwing this up so people can take a look at it because I might be entirely off base with the general vibe of the article.

I plan to write a more prescriptive tutorial that can replace "Structuring A Command-Based Robot Project" to go with this more in-depth discussion.

Also, this needs C++ examples throughout, I can probably take a crack at it but I don't have any experience with WPILibC so I'm leaving them as TODOs for now.

This also intersects with #1808, but I'm doing it as a separate PR for now in order to not cause conflicts. Happy to move things around if that's desired.

### DIFF
--- a/source/docs/software/commandbased/organizing-command-based.rst
+++ b/source/docs/software/commandbased/organizing-command-based.rst
@@ -1,0 +1,259 @@
+Organizing Command-Based Robot Projects
+=======================================
+
+As your robot code becomes more complicated, navigating, understanding, and maintaining the code takes up more and more time and energy. Making changes to the code often becomes more difficult, sometimes for reasons that have very little to do with the actual complexity of the underlying logic. For a simplified example: putting the logic for many unrelated robot functions into a single 1000-line file makes it difficult to find a specific piece of code within that file, particularly under stress at a competition. But spreading out closely related logic across dozens of tiny files is often just as difficult to navigate.
+
+This is not a problem unique to FRC, and in fact, good organization only becomes more and more critical as software projects become bigger and bigger. The "best" organization system is a perennial topic of debate, much like the "best" programming language, but in the end, the choice (in both cases) comes down to the specific task at hand and the programmer (or programmers) implementing said task. Even in the relatively small space of FRC robot programming, there is no right answer. The best choice for a given team will depend on the nature of the specific robot code, team structure, and pure personal preference.
+
+This article discusses various facets of command-based robot program design that advanced FRC programmers may want to be aware of when writing code. It is not a prescriptive tutorial, though it presents some recommended best practices. If this level of choice seems daunting, however, many teams have been highly successful while sticking closely to WPILib's example code and guidelines. However, this discussion may be of interest to intermediate and advanced programmers who want to make their code not only effective, but flexible, easily changeable, and sometimes even beautiful.
+
+Why Care About Organization?
+----------------------------
+Good code organization will rarely make or break a team's competitive ability—but it does mean easier debugging, faster modifications, nicer-looking code, and happier programmers. But while it's impossible to define "good" organization by way of what the code looks like from the inside, it's easier to define in terms of what the robot's software looks like from the outside.
+
+What Good Organization Looks Like
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+When code is well-designed and well-organized, the code's internal structure is intuitive and easily comprehensible. Cumbersome boilerplate is minimized, meaning that new robot functionality can often be added with just a few lines of code. When a constant value (such as the speed of the robot's intake) needs to be changed, it only needs to change in one place. If multiple programmers are working together, they can easily understand each others' work. Bugs are rare, since it is difficult to accidentally introduce unintended behavior (such as creating a command that does not require necessary subsystems). Implementing more advanced functions like unit tests is easier, since the code is abstracted away from the physical hardware. Programmers are happy (most of the time).
+
+What Bad Organization Looks Like
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Poorly organized code often has internal structure that makes little to no sense, even to whoever wrote it. When functionality has to be added or changed, it often breaks unrelated parts of the robot: adding automatic shooter control might introduce a bug in the climbing sequence for unclear reasons. Alternatively, the organizational framework might be so strict that it's impossible to implement necessary behavior, requiring nasty hacks or workarounds. Many lines of boilerplate code are needed for simple robot logic. Collaboration among multiple programmers is difficult or impossible.
+
+Defining Commands
+-----------------
+
+In larger robot codebases, multiple copies of the same command need to be used in many different places. For instance, a command that runs a robot's intake might be used in teleop, bound to a certain button; as part of a complicated command group for an autonomous routine; and as part of a self-test sequence.
+
+As an example, let's look at some ways to define a simple command that simply runs the robot's intake forward at full power until cancelled.
+
+Inline Commands
+^^^^^^^^^^^^^^^
+
+The easiest and most expressive way to do this is with a ``StartEndCommand``:
+
+.. tabs::
+
+  .. code-tab:: java
+
+    Command runIntake = new StartEndCommand(() -> intake.set(1), () -> intake.set(0), intake);
+
+  .. code-tab:: c++
+
+    // TODO
+
+This is sufficient for commands that are only used once. However, for a command like this that might get used in many different autonomous routines and button bindings, inline commands everywhere means a lot of repetitive code:
+
+.. tabs::
+
+  .. code-tab:: java
+
+    // RobotContainer.java
+    intakeButton.whileHeld(new StartEndCommand(() -> intake.set(1), () -> intake.set(0), intake));
+
+
+    Command intakeAndShoot = new StartEndCommand(() -> intake.set(1), () -> intake.set(0), intake)
+        .alongWith(new RunShooter(shooter));
+
+    Command autonomousCommand = new ParallelCommandGroup(
+        new StartEndCommand(() -> intake.set(1), () -> intake.set(0), intake).withTimeout(5),
+        new WaitCommand(3),
+        new StartEndCommand(() -> intake.set(1), () -> intake.set(0), intake).withTimeout(5)
+    )
+
+  .. code-tab:: c++
+
+    // TODO
+
+(Creating one ``StartEndCommand`` instance and putting it in a variable won't work here, since once an instance of a command is added to a command group it is effectively "owned" by that command group and cannot be used in any other context.)
+
+Subclassing CommandBase
+^^^^^^^^^^^^^^^^^^^^^^^
+
+One possible solution to this problem is to create a new subclass of ``CommandBase`` that implements the necessary ``initialize`` and ``end`` routines.
+
+.. tabs::
+
+  .. code-tab:: java
+
+    public class RunIntakeCommand extends CommandBase {
+        private Intake m_intake;
+
+        public RunIntakeCommand(Intake intake) {
+            this.m_intake = intake;
+            addRequirements(intake);
+        }
+
+        @Override
+        public void initialize() {
+            m_intake.setPercent(1);
+        }
+
+        @Override
+        public void end(boolean interrupted) {
+            m_intake.setPercent(0);
+        }
+
+        @Override
+        public boolean isFinished() {
+            return false;
+        }
+    }
+
+  .. code-tab:: c++
+
+    // TODO
+
+This, however, is just as cumbersome. The only two lines that really matter in this entire file are the two calls to ``intake.setPercent``, yet there are over 20 lines of boilerplate code! Not to mention, doing this for a lot of robot actions quickly clutters up a robot project with dozens of small files. Nevertheless, this might feel more "natural," particularly for programmers who don't feel confident with Java and want to stick closely to an object-oriented model.
+
+Factory Methods
+^^^^^^^^^^^^^^^
+
+A useful middle ground between these two extremes is using a `factory method <https://www.tutorialspoint.com/design_pattern/factory_pattern.htm>`__. A factory method is a method that, each time it is called, returns a new object according to some specification.
+
+A command like the intake-running command is conceptually related to exactly one subsystem: the ``Intake``. As such, it makes sense to put a ``commandRun`` method as an instance method of the ``Intake`` class:
+
+.. tabs::
+
+  .. code-tab:: java
+
+    public class Intake {
+        // [code for motor controllers, configuration, etc.]
+        // ...
+
+        public Command commandRun() {
+            return new StartEndCommand(() -> this.set(1), () -> this.set(0), this);
+        }
+    }
+
+  .. code-tab:: c++
+
+    // TODO
+
+Notice how since we are in the ``Intake`` class, we no longer refer to ``intake``; instead, we use the ``this`` keyword to refer to the current instance.
+
+.. warning:: To preserve encapsulation, avoid referring to ``private`` variables and methods of the ``Intake`` class from within the ``commandRun`` method, even though they are technically accessible. (For instance, the ``commandRun`` method should use the subsystem's public ``set`` method instead of interfacing directly with the motor controllers.)
+
+Using this new factory method in command groups and button bindings is highly expressive:
+
+
+.. tabs::
+
+  .. code-tab:: java
+
+    intakeButton.whileHeld(intake.commandRun());
+
+    Command intakeAndShoot = intake.commandRun().alongWith(new RunShooter());
+
+    Command autonomousCommand = new SequentialCommandGroup(
+        intake.commandRun().withTimeout(5),
+        new WaitCommand(3),
+        intake.commandRun().withTimeout(5)
+    );
+
+  .. code-tab:: c++
+
+    // TODO
+
+Adding a parameter to the ``commandRun`` method to provide the exact percentage to run the intake is easy and allows for even more flexibility.
+
+.. tabs::
+
+  .. code-tab:: java
+
+    public Command commandRun(double percent) {
+        return new StartEndCommand(() -> this.set(percent), () -> this.set(0), this);
+    }
+
+  .. code-tab:: c++
+
+    // TODO
+
+For instance, this code creates a command group that runs the intake forwards for two seconds, waits for two seconds, and then runs the intake backwards for five seconds.
+
+.. tabs::
+
+  .. code-tab:: java
+
+    Command intakeRunSequence = intake.commandRun(1).withTimeout(2)
+        .andThen(new WaitCommand(2))
+        .andThen(intake.commandRun(-1).withTimeout(5));
+
+  .. code-tab:: c++
+
+    // TODO
+
+Command Groups
+^^^^^^^^^^^^^^
+
+Command groups have slightly different organizational concerns, but many of the same principles apply. Certain command groups are used in just one place, and so it's more sensible to define a command group entirely using inline decorators shortly before it is used (such as the ``intakeRunSequence`` command immediately above). However, command groups that are reused often, or large command groups such as autonomous routines, are still better when split into separate files. We'll consider command group using the same example from the previous section:
+
+.. tabs::
+
+  .. code-tab:: java
+
+    intake.commandRun(1).withTimeout(2)
+        .andThen(new WaitCommand(2))
+        .andThen(intake.commandRun(-1).withTimeout(5));
+
+  .. code-tab:: c++
+
+    // TODO
+
+Subclassing
+~~~~~~~~~~~
+
+When splitting a command group into a separate file, subclassing the desired type is a natural solution. In this design, the command group is configured in the constructor.
+
+.. tabs::
+
+  .. code-tab:: java
+
+    public class IntakeThenOuttake extends SequentialCommandGroup {
+        public IntakeThenOuttake(Intake intake) {
+            super(
+                intake.commandRun(1).withTimeout(2),
+                new WaitCommand(2),
+                intake.commandRun(-1).withTimeout(5)
+            );
+        }
+    }
+  .. code-tab:: c++
+
+    // TODO
+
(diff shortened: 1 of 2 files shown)


source/docs/software/commandbased/organizing-command-based.rst
```diff
@@ -0,0 +1,259 @@
...
+  .. code-tab:: c++
+
+    // TODO
+
+(Creating one ``StartEndCommand`` instance and putting it in a variable won't work here, since once an instance of a command is added to a command group it is effectively "owned" by that command group and cannot be used in any other context.)
+
+Subclassing CommandBase
+^^^^^^^^^^^^^^^^^^^^^^^
+
+One possible solution to this problem is to create a new subclass of ``CommandBase`` that implements the necessary ``initialize`` and ``end`` routines.
```
Starlight220, line 74: I think it's not completely clear what the "problem" here is and how subclassing solves it.
Maybe something like "An alternative architecture is ...", and then establishing pros/cons or something.

source/docs/software/commandbased/organizing-command-based.rst
```diff
@@ -0,0 +1,259 @@
...
+        new StartEndCommand(() -> intake.set(1), () -> intake.set(0), intake).withTimeout(5),
+        new WaitCommand(3),
+        new StartEndCommand(() -> intake.set(1), () -> intake.set(0), intake).withTimeout(5)
+    )
+
+  .. code-tab:: c++
+
+    // TODO
+
+(Creating one ``StartEndCommand`` instance and putting it in a variable won't work here, since once an instance of a command is added to a command group it is effectively "owned" by that command group and cannot be used in any other context.)
```
Starlight220, line 69: I'd have this as a regular paragraph or in an admonition instead of in parentheses.

source/docs/software/commandbased/organizing-command-based.rst
```diff
@@ -0,0 +1,259 @@
...
+  .. code-tab:: c++
+
+    // TODO
+
+This, however, is just as cumbersome. The only two lines that really matter in this entire file are the two calls to ``intake.setPercent``, yet there are over 20 lines of boilerplate code! Not to mention, doing this for a lot of robot actions quickly clutters up a robot project with dozens of small files. Nevertheless, this might feel more "natural," particularly for programmers who don't feel confident with Java and want to stick closely to an object-oriented model.
+
+Factory Methods
+^^^^^^^^^^^^^^^
+
+A useful middle ground between these two extremes is using a `factory method <https://www.tutorialspoint.com/design_pattern/factory_pattern.htm>`__. A factory method is a method that, each time it is called, returns a new object according to some specification.
```
Starlight220, line 113: I'm not sure about the link here; it seems to refer more to factory objects, which I'm not sure is fully relevant.

source/docs/software/commandbased/organizing-command-based.rst
```diff
@@ -0,0 +1,259 @@
...
+        }
+    }
+
+  .. code-tab:: c++
+
+    // TODO
+
+Notice how since we are in the ``Intake`` class, we no longer refer to ``intake``; instead, we use the ``this`` keyword to refer to the current instance.
+
+.. warning:: To preserve encapsulation, avoid referring to ``private`` variables and methods of the ``Intake`` class from within the ``commandRun`` method, even though they are technically accessible. (For instance, the ``commandRun`` method should use the subsystem's public ``set`` method instead of interfacing directly with the motor controllers.)
```
Starlight220, line 136: I disagree with this. If already preserving encapsulation, I'd advise *not* having public `set` methods and exposing only command factories (and getters, when they are needed from outside the subsystem). See [6738's 2022 code for example](https://github.com/ExcaliburFRC/RapidReact2022/blob/main/src/main/java/io/excaliburfrc/robot/subsystems/Intake.java).
It also mostly solves the namespace collision issue.

source/docs/software/commandbased/organizing-command-based.rst
```diff
@@ -0,0 +1,259 @@
...
+        intake.commandRun().withTimeout(5),
+        new WaitCommand(3),
+        intake.commandRun().withTimeout(5)
+    );
+
+  .. code-tab:: c++
+
+    // TODO
+
+Adding a parameter to the ``commandRun`` method to provide the exact percentage to run the intake is easy and allows for even more flexibility.
```
Starlight220, line 159: Beware of too much flexibility. In most cases it's better to accept an enum or multiple factories for a number of preset options instead of giving absolute control to whoever is calling the command.

source/docs/software/commandbased/organizing-command-based.rst
```diff
@@ -0,0 +1,259 @@
...
+        .andThen(intake.commandRun(-1).withTimeout(5));
+
+  .. code-tab:: c++
+
+    // TODO
+
+Command Groups
+^^^^^^^^^^^^^^
+
+Command groups have slightly different organizational concerns, but many of the same principles apply. Certain command groups are used in just one place, and so it's more sensible to define a command group entirely using inline decorators shortly before it is used (such as the ``intakeRunSequence`` command immediately above). However, command groups that are reused often, or large command groups such as autonomous routines, are still better when split into separate files. We'll consider command group using the same example from the previous section:
```
Starlight220, line 190: I'm not sure about this split. I think that the line between command and group is very blurry, and often times the "simple" factories in the subsystem are groups as well--that isn't necessarily clear from the way this is currently written.

I think the decision of whether the command should be in a separate file should be based on number of requirements rather than "how much it's used".

source/docs/software/commandbased/organizing-command-based.rst
```diff
@@ -0,0 +1,259 @@
...
+  .. code-tab:: c++
+
+    // TODO
+
+This is relatively short and minimizes boilerplate. It is also comfortable to use in a purely object-oriented paradigm and may be more acceptable to novice programmers. However, it has some downsides. For one, it is not immediately clear exactly what type of command group this is from the constructor definition: it is better to define this in a more inline and expressive way, particularly when nested command groups start showing up. Additionally, it requires a new file for every single command group, even when the groups are conceptually related.
+
+Static Factory Methods
+~~~~~~~~~~~~~~~~~~~~~~
+
+Instead, it's better to split command groups into *static* factory methods, and provide their subsystem dependencies via dependency injection.
```
Starlight220, line 231: see above re "An alternative architecture".

Starlight220 (changes requested):
This is really good, and I really like how it shows the different options!

Maybe it could better indicate that each structure isn't all-or-nothing and some commands can be done in one way and others in another?

Expanding the command vs group topic from one of the review comments: maybe it would be better to present the four options of (factories in subsystem, static factories, subclassing CommandBase, subclassing CommandGroupBase) without splitting into groups vs not groups?

source/docs/software/commandbased/organizing-command-based.rst
RobinsonZ: In real-world code I'd be inclined to agree, but this is just as an example of how command factories can be parameterized. Defining an enum and explaining the logic behind it would distract from the main lesson IMO.

source/docs/software/commandbased/organizing-command-based.rst
RobinsonZ: Maybe this prose is unclear—by "separate file" I mean defined right when it's used, for example in `RobotContainer` as button bindings are being set up. I definitely agree that the line between commands and command groups is a bit awkward though.

I'm trying to preserve a sort of narrative, starting with very simple intake commands and moving on to command groups in separate static classes. If this were a spec sheet of information it's probably better to use the four different categories, but I'm not sure if that's best here.

source/docs/software/commandbased/organizing-command-based.rst
Starlight220: I think it would be better in the instance factory method to use the motor object directly, and then add a comment somewhere that in some cases it might be better to extract logic to non-command methods (and then teams would make them public if needed).

source/docs/software/commandbased/organizing-command-based.rst
Starlight220: Most of the articles (especially the command-based ones) are spreads of information, and the narrative/progression is only at the article level (ie articles about simpler topics are before articles about advanced topics).

Each of the structures has a use case that it is good for (and the example code should show that), so sorting the sections by the complexity of the example/use case is a good idea.
I'd say something like this:
1. instance factory methods -- simple intake commands
2. subclassing CommandBase  -- single commands that have somewhat custom logic (using the same example as 1 to show how it can get verbose is a good idea, well done!)
3. static factories -- "big" commands such as auto routines
4. subclassing command groups (with only a `super()` call; no overriding methods) -- "big" commands such as auto routines

source/docs/software/commandbased/organizing-command-based.rst
```diff
@@ -0,0 +1,299 @@
...
+
+    Command intakeRunSequence = IntakeCommands.run(intake, 1).withTimeout(2)
+        .andThen(new WaitCommand(2))
+        .andThen(IntakeCommands.run(intake, 1).withTimeout(5));
+
+  .. code-tab:: c++
+
+    // TODO
+
+The trade-offs between static and instance factory methods for single subsystems make the choice between them entirely a matter of personal opinion, and shouldn't be the subject of too much concern. It's only recommended to pick one way and stick with it throughout the entire robot project. Having a prescribed system makes code easy and fluid to navigate and reduces confusion between multiple contributing team members.
```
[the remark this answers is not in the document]
RobinsonZ: Maybe this paragraph needs more clarification. The point I want to make is that for most of the choices presented, there are distinct tradeoffs that you need to carefully weigh in accordance with your situation, and while there's no one right answer there are some wrong ones. (e.g. doing [this](https://github.com/wpilibsuite/frc-docs/pull/1828/files#diff-2dabb56aecfa45c4147245f695e93b673ac6b6d1b640942812d230ba86c9172cR53) is just wrong and bad, period, and using factory methods might be a bad idea if most of your programming team doesn't even fluently understand object-oriented programming yet.) By contrast, the choice between static factory and instance factory for single-subsystem commands is almost entirely a matter of vibes (see @Oblarg on [Discord](https://discord.com/channels/176186766946992128/368993897495527424/975657701923246091))—it's basically whether you subjectively care more about making the `Intake` class look nice or making your call sites look nice.

source/docs/software/commandbased/organizing-command-based.rst
RobinsonZ: > Most of the articles (especially the command-based ones) are spreads of information, and the narrative/progression is only at the article level (ie articles about simpler topics are before articles about advanced topics).

Yes, but I'm explicitly not following that model here—I plan to write a more spec-sheet-y companion piece that's just "here's a good structure for intermediate robot code" (that might need some design discussion). Maybe I should put a bigger notice up at the top, but this is designed for programmers who want to do some legitimate thinking about what works best for them.

I'm not sure I agree with your categorization of which command group is best for which use case (which sort of indicates how this works better as a design _discussion_ rather than a prescriptive guide). 

For example, for single commands that have somewhat complicated logic, I personally would usually avoid subclassing `CommandBase` and instead use a `FunctionalCommand` in a static or non-static factory method. (I'd also try to just generally avoid complicated logic in commands, and try to achieve the functionality via command composition instead.) But again, this is a matter of personal taste—figuring out command composition is often more difficult for inexperienced teams then building a somewhat chunky command subclass with its own little state machine inside. 

I may take another crack at revising things, though, so watch this space.

source/docs/software/commandbased/organizing-command-based.rst
RobinsonZ: I strongly dislike giving any `Command` code direct knowledge of its subsystem's hardware setup (consider the scenario where a motor controller gets changed, or motors are added or removed).

I think I'm going to try to have another discussion of subsystem encapsulation here as well (it's looking like maybe this should be split into multiple documents), so maybe we have a link here to that discussion but for now we should just take a position and stick to it while we talk about other things (namely where commands live). Sort of the same deal as the enum thing.

source/docs/software/commandbased/organizing-command-based.rst
```diff
@@ -0,0 +1,299 @@
...
+        .andThen(intake.commandRun(-1).withTimeout(5));
+
+  .. code-tab:: c++
+
+    // TODO
+
+Static Command Factories
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+The factory methods presented above do have some negatives. They clutter up the `Intake` class (and the namespace of `Intake` methods) with command-related methods. With that pattern, it's also possible to break the intended encapsulation of the subsystem (as the warning box alludes to), since code in these factory methods can access `private` members of the `Intake` class. To avoid these pitfalls, it's possible to define intake-related methods as static members of a separate `IntakeCommands` class:
```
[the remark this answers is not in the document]
RobinsonZ: What do you see duplicated?

> Regardless, I see static factories more for multi-subsystem commands than for single-subsystem ones.

I personally agree, but I can see some teams being more comfortable with "all commands are static factories" because (one possible reason) that's just fewer bits of information they have to drill into the heads of a bunch of novices. (Maybe I should add that to the "pros" list for static single-subsystem factories.)

source/docs/software/commandbased/organizing-command-based.rst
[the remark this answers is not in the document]
Starlight220: All four are mostly semantic differences. I'm not sure what would be gained by highlighting such a contrast, and why wouldn't mentioning the "wrong answers" as cons accomplish that?
To give an example of what I mean, how about something like the following in the section for instance factory methods:
> While very convenient for commands that require only a single subsystem, this approach introduces multiple code smells and does not work well for commands requiring multiple subsystems.

Is there something here lacking for pointing away from the "wrong answer" of an instance factory method for multi-subsystem commands?

--- 
<details>
<summary>off-topic</summary>

> (e.g. doing [this](https://github.com/wpilibsuite/frc-docs/pull/1828/files#diff-2dabb56aecfa45c4147245f695e93b673ac6b6d1b640942812d230ba86c9172cR53) is just wrong and bad, period, and using factory methods might be a bad idea if most of your programming team doesn't even fluently understand object-oriented programming yet.)

I'm going to disagree with you again about that.
But this is going off-topic, if you want we can talk about it on Discord.
</details>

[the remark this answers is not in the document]
RobinsonZ: I didn't consider the case of commands (not command groups) that require two subsystems at once. I'll add a note about that. I also am rephrasing this paragraph somewhat, so see how well you like the revision that should go up in a few minutes.

RobinsonZ:
It seems like there's an unrelated failure in the link checking task because https://www.analog.com/en/landing-pages/001/first.html has some sort of blocking going on (at least, I can access it from my machine, but the link checker [isn't happy about it](https://github.com/wpilibsuite/frc-docs/runs/6448802217?check_suite_focus=true#step:6:1441)). Not sure what to do about that.

source/docs/software/commandbased/organizing-command-based.rst
Starlight220: > I personally would usually avoid subclassing CommandBase and instead use a FunctionalCommand in a static or non-static factory method. 

Since you're implementing all four command lifecycle methods anyway, there's little difference in verbosity; writing a class allows you to more easily share state between the lifecycle methods. Having said all that, I do agree that it's a matter of personal preference.
As for the rest of that paragraph, I 100% agree.

---

> Yes, but I'm explicitly not following that model here—I plan to write a more spec-sheet-y companion piece that's just "here's a good structure for intermediate robot code" (that might need some design discussion). Maybe I should put a bigger notice up at the top, but this is designed for programmers who want to do some legitimate thinking about what works best for them.

I'm not sure what we're disagreeing here about. What I was thinking of for this doc is an explanation of each option with pros and cons, and leave it as an exercise to the reader to figure out what's appropriate and works best for their use case. I was not intending for this to be a "The *One* True Way".
However, concrete examples massively help to understand these abstract topics, and examples need to be based on experience--which is why it's open for discussion. I gave those examples because that's what matches my experience, I'm open to others' experience.

source/docs/software/commandbased/organizing-command-based.rst
Starlight220: Swapping motor controllers shouldn't change anything, that's what polymorphism exists for. Motors being added/removed would either need to modify commands anyway or it wouldn't at all.
And encapsulation isn't a problem here; nothing escapes outside.
What caused my stance on this is seeing three levels of indirection with an `arcadeDriveCommand` factory method that called an `arcadeDrive` method that called `DifferentialDrive.arcadeDrive()`: the middle level is more than redundant.

source/docs/software/commandbased/organizing-command-based.rst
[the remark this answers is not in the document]
Starlight220: > What do you see duplicated?

There's a static command factory section and a static command group factory section.

> (Maybe I should add that to the "pros" list for static single-subsystem factories.)

I'd be very careful with wording that pro, if at all. I'm not sure knowing to only do one thing is good.

source/docs/software/commandbased/organizing-command-based.rst
```diff
@@ -0,0 +1,313 @@
...
+        .andThen(new WaitCommand(2))
+        .andThen(IntakeCommands.run(intake, 1).withTimeout(5));
+
+  .. code-tab:: c++
+
+    // TODO
+
+The trade-offs between static and instance factory methods for single subsystems make the choice between them entirely a matter of personal opinion, and shouldn't be the subject of too much concern. Some programming teams might prefer static methods to maintain consistency with multi-subsystem command groups, in order to define commands consistently across the entire codebase. However, other teams might want to take advantage of the more fluid syntax offered by instance methods. For the command group discussion below, this article uses instance method calls.
+
+.. note:: Much like command groups, commands that require multiple subsystems should not be defined in instance methods of subsystems, since they are not conceptually related to a single subsystem. These commands should be either defined inline at their usage location, constructed using a static factory method, or made a subclass of ``CommandBase`` (depending on complexity of the command and programmer preference).
```
Starlight220, line 225:
```suggestion
.. note:: Commands that require multiple subsystems should not be defined in instance methods of subsystems, since they are not conceptually related to a single subsystem. These commands should be either defined inline at their usage location (if used in only one location) or constructed using one of the other options detailed in this article, depending on the complexity of the command and programmer preference).
```
No reason to split commands vs groups.

This should be moved to the section about instance factory methods.

Starlight220:
What do you think about this sort of structure for the article?:

- introduction
  - (what you have now is mostly good)
- inline commands
  - problem (=lack of reusability)
- method factories
  - explanation
  - example code
  - (dis)advantages
- subclassing `CommandBase`
  - (same 3 subitems as method factories)
- static factories
  - (same 3 subitems as method factories)
- subclassing `CommandGroupBase`
  - (same 3 subitems as method factories)
- (maybe somewhat of a summary here? like a table comparing pros and cons or something)

(maybe flip the order between static factories and extending `CommandGroupBase`)

---
(side note: the lack of linearity in the discussion here is really confusing)

Starlight220:
seeing as this hasn't been touched in a few months, I want to ask:
@RobinsonZ do you intend to continue working on this? if not, do you mind if I take over?

RobinsonZ:
@Starlight220 Yeah, unfortunately I've had considerably less free time than expected this summer with internships and all. Feel free to take over, hopefully you find some of this existing copy useful.

Starlight220:
@RobinsonZ what do you think? since this is originally your PR, I'd be happy for your review as well.

Daltz333:
I think it would be a good idea to have a visual representation (graph/tree) of project structure.

Daltz333 (approved):

source/docs/software/commandbased/organizing-command-based.rst
```diff
@@ -0,0 +1,318 @@
...
+    Command intakeRunSequence = intake.runIntakeCommand(1.0).withTimeout(2.0)
+        .andThen(new WaitCommand(2.0))
+        .andThen(intake.runIntakeCommand(-1.0).withTimeout(5.0));
+
+  .. code-tab:: c++
+
+    // TODO
+
+
+This approach is recommended for commands that are conceptually related to only a single subsystem, and is very consise. However, it doesn't fare well with commands related to more than one subsystem: passing in other subsystem objects is unintuitive and can cause race conditions and circular dependencies, and thus should be avoided. Therefore, this approach is best suited for single-subsystem commands, and should be used only for those cases.
```
shueja-personal, line 150: “Concise”, not “consise”

source/docs/software/commandbased/organizing-command-based.rst
```diff
@@ -0,0 +1,318 @@
...
+  .. code-tab:: c++
+
+    // TODO
+
+Creating one ``StartEndCommand`` instance and putting it in a variable won't work here, since once an instance of a command is added to a command group it is effectively "owned" by that command group and cannot be used in any other context.
+
+Instance Command Factory Methods
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+One way to solve this quandry is using the "factory method" design pattern: a function that returns a new object every invocation, according to some specification. Using command composition and :ref:`decorators<docs/software/commandbased/convenience-features:Command Decorator Methods>`, a factory method can construct a complex command object with merely a few lines of code.
```
shueja-personal, line 74: “quandary”, not “quandry”

source/docs/software/commandbased/organizing-command-based.rst
```diff
@@ -0,0 +1,318 @@
...
+                intake.runIntakeCommand(-1).withTimeout(2.0)
+            );
+        }
+
+    }
+  .. code-tab:: c++
+
+    // TODO
+
+Also, note the use of static factories to construct sequential and parallel command groups: this is equivalent to the ``andThen`` and ``alongWith`` decorators, but can be more expressive in some cases. Their use is a matter of personal preference
```
shueja-personal, line 254: Missing period.

shueja-personal (commented):
A few editing nitpicks, but the content looks wonderful.

Starlight220 (approved):

PR #1842, opened 2022-06-20
Fix and Improve Internal Links Style Guide

JonathanLindsey:
This fixes and improves the Page References section of the Style Guide.

If there's a better way to format these examples rather than strung together in a paragraph, that would be great.


Daltz333:
Sorry for the late response.  This looks great and better represents what we currently follow.

Daltz333 (approved):

PR #1849, opened 2022-07-09
Fixing some typos and run-on sentences

SpencerGreene:


Daltz333 (dismissed):

sciencewhiz:
I'm not sure I like the periods immediately after commands, seems like there could be some misunderstandings.

SpencerGreene:
> I'm not sure I like the periods immediately after commands, seems like there could be some misunderstandings.

Good point on clarity - I checked and this is in two places, https://docs.wpilib.org/en/stable/docs/contributing/frc-docs/build-instructions.html?highlight=poetry#installing-poetry and https://docs.wpilib.org/en/stable/docs/contributing/frc-docs/build-instructions.html?highlight=poetry#installing-poetry.  On the HTML render at least, the literal command renders boxed in red, and the period is outside the box - so I think it's clear.  I'm not sure if other renderings run them together?

Eg, run the following command: ``poetry lock``.

Daltz333 (approved):

PR #1851, opened 2022-07-24
explain how RP are only given out during qual matches

michaelleong25:


sciencewhiz:
Since ranking point isn't used in any article, I think I'd prefer to remove it from the glossary, especially since sometimes it's definition changes based on a particular year's rules

Kevin-OConnor:
I agree with Joe. Game Manual terms like RP probably shouldn't be in this glossary to avoid having the same thing defined multiple places and forgetting to update one.

PR #1861, opened 2022-07-28
Drivetrain Simulation note: write code in steps 1-2 in a separate subsystem 

michaelleong25:
It is not clear where to write the code in steps 1-2; for this reason, I added a note indicating that if the user is using the command-based framework, they should write the following code in a separate subsystem.


source/docs/software/wpilib-tools/robot-simulation/drivesim-tutorial/simulation-instance.rst
```diff
@@ -2,6 +2,8 @@ Step 1: Creating Simulated Instances of Hardware
 ================================================
 The WPILib simulation framework contains several ``XXXSim`` classes, where ``XXX`` represents physical hardware such as encoders or gyroscopes. These simulation classes can be used to set positions and velocities (for encoders) and angles (for gyroscopes) from a model of your drivetrain. See :ref:`the Device Simulation article<docs/software/wpilib-tools/robot-simulation/device-sim:Device Simulation>` for more info about these simulation hardware classes and simulation of vendor devices.
 
+.. note:: If you are using the command-based framework, it is recommended that you write the code in steps 1-2 in a separate subsystem named ``Drivetrain``.
```
prateekma, line 5: Not a big fan of the wording here. Maybe we should just mention that simulation objects associated with a particular subsystem should live in that subsystem and then point to a command-based example instead. Currently it sounds like simulation objects should be placed in a **separate** subsystem that is different from the main subsystem.

prateekma (changes requested):

Daltz333 (approved):

PR #1862, opened 2022-07-28
Make state-space Python scripts work with latest version of frccontrol

calcmogul:
Also give execute permissions to Python scripts. Images weren't
regenerated because the output looks identical.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -8,7 +8,7 @@ python-versions = "*"
 
 [[package]]
 name = "babel"
-version = "2.10.1"
+version = "2.10.3"
 description = "Internationalization utilities"
 category = "main"
 optional = false
@@ -53,7 +53,7 @@ cffi = ">=1.0.0"
 
 [[package]]
 name = "certifi"
-version = "2022.5.18.1"
+version = "2022.6.15"
 description = "Python package for providing Mozilla's CA Bundle."
 category = "main"
 optional = false
@@ -61,7 +61,7 @@ python-versions = ">=3.6"
 
 [[package]]
 name = "cffi"
-version = "1.15.0"
+version = "1.15.1"
 description = "Foreign Function Interface for Python calling C code."
 category = "main"
 optional = false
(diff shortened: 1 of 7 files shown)


TheTripleV:
The versions of these dependencies should also be locked in vcs

calcmogul:
They don't appear to be installed at all in CI. The image generation isn't automated.

TheTripleV:
I think automating image generation is the way to go. Even without that, these should be locked so that the scripts won't magically stop working.

calcmogul:
Locking them wouldn't do anything because currently, the scripts have to be run manually using whatever packages are installed locally.

TheTripleV:
Locally, you're already supposed to use the deps specified via requirements.txt or poetry. So it would use the locked version.

Daltz333:
Some of the bumped dependencies require Python 3.8. Can you double check that the contributing instructions have that listed as a minimum version, and if not, bump it?

sciencewhiz:
Is the latest version of numpy actually required or could an older version be used to avoid bumping the python version?

calcmogul:
It would take me a while to figure out what earlier versions of everything in the dependency chain are valid. For what it's worth, Python 3.8 had its first release almost three years ago, and Python 3.7 will be end-of-life in 11 months.

https://devguide.python.org/versions/#supported-versions

calcmogul:
> Can you double check that the contributing instructions have that listed as a minimum version, and if not, bump it?

The minimum version in build-instructions.rst says the required version is Python 3.9 (and that 3.10 is incompatible).
```
frc-docs/build-instructions.rst
16:- `Python 3.9 <https://www.python.org/downloads/>`__ (Python >= 3.10 is incompatible)
```

TheTripleV:
In pyproject.toml, the min python version is marked as 3.9 so poetry picks 3.9+ compatible dependencies. The behavior is correct

TheTripleV (approved):

Daltz333 (approved):

PR #1864, opened 2022-08-02
fix VS Code styling 

michaelleong25:


sciencewhiz:
Since VS Code is an abbreviation, the first time it's used on the page it should be spelled out Visual Studio Code

Daltz333 (approved):

PR #1874, opened 2022-08-16
Force scipy and lower maximum python version

Daltz333:
Required for `pip install -r source/requirements.txt` on Windows after adding `frccontrol`. This is due to `scipy` being classified by poetry as a restricted dependency, so it prefers their maximum Python version over our own.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -29,8 +29,8 @@ python-versions = ">=3.6.0"
 soupsieve = ">1.2"
 
 [package.extras]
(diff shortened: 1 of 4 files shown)


TheTripleV:
Does frc docs build on 3.10/11?

sciencewhiz:
The build instructions should be updated. It says 3.10 is incompatible.

TheTripleV (approved):

PR #1876, opened 2022-08-18
Improve CI

uvidyadharan:
Cuts down build times from 16+ min to sub 8 min. 

- Use apt-fast
- Cache python packages
- Disable man pages
- Multithread builds
- Split PDF and HTML builds into separate jobs 
- Fail build if artifact not found

*Note, if merged please do a squash merge

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -3,7 +3,7 @@ name: CI
 on: [pull_request, push]
 
 jobs:
-  build:
+  build-pdf:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
@@ -20,14 +20,6 @@ jobs:
       run: |
         sudo apt-fast -y update
         sudo apt-fast install -y texlive-latex-recommended texlive-fonts-recommended texlive-latex-extra latexmk texlive-lang-greek texlive-luatex texlive-xetex texlive-fonts-extra dvipng librsvg2-bin
-    - name: Build HTML
-      run: |
-        make html SPHINXOPTS="-j 3"
-    - name: Archive HTML
-      uses: actions/upload-artifact@v2
-      with:
-        name: frc-docs-html
-        path: build/html/
     - name: Build PDF
       run: |
         make latexpdf SPHINXOPTS="-j 3"
@@ -36,6 +28,31 @@ jobs:
       with:
         name: frc-docs-pdf
         path: build/latex/firstroboticscompetition.pdf
+        if-no-files-found: error
+        
+  build-html:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Run Tune
+      uses: abbbi/github-actions-tune@v1
+    - uses: actions/setup-python@v4
+      with:
+        python-version: 3.9
+        cache: 'pip'
+    - name: Install Python Dependencies
+      run: |
+        pip install -r source/requirements.txt
+    - name: Build HTML
+      run: |
+        make html SPHINXOPTS="-j 3"
+    - name: Archive HTML
+      uses: actions/upload-artifact@v2
+      with:
+        name: frc-docs-html
+        path: build/html/
+        if-no-files-found: error
+        
   check-links:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


Daltz333:
What are build times with just the PDF/HTML split changes?

uvidyadharan:
> What are build times with just the PDF/HTML split changes?

I did a quick test and I got 9 min 45 seconds.

https://github.com/uvidyadharan/frc-docs/actions/runs/2884628582

image

.github/workflows/CI.yml
```diff
@@ -3,43 +3,63 @@ name: CI
...
 
 jobs:
-  build:
+  build-pdf:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-    - uses: actions/setup-python@v2
+    - name: Run Tune
+      uses: abbbi/github-actions-tune@v1
```
prateekma, line 11: We should probably vet this to make sure it has enough users / won't break under us. Otherwise we can fork it like we do for the macOS notarization actions.

.github/workflows/CI.yml
```diff
@@ -3,43 +3,63 @@ name: CI
...
-  check-links:
+        if-no-files-found: error
+        
+  build-html:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-    - name: Checkout main
+    - name: Run Tune
+      uses: abbbi/github-actions-tune@v1
```
prateekma, line 38: Same comment as above.

.github/workflows/CI.yml
```diff
@@ -3,43 +3,63 @@ name: CI
 on: [pull_request, push]
 
 jobs:
-  build:
+  build-pdf:
```
prateekma, line 6: We'll need to update the repo settings to amend the "required" builds before merging the PR. @Daltz333

TheTripleV:
I think PRs are time bound by the rtd build, not the GitHub one.

Also, let's not do parallel builds. Last I checked, they don't work on most platforms and don't work on newer python versions and potentially contain some race conditions.

sciencewhiz:
> I think PRs are time bound by the rtd build, not the GitHub one.

For a while now, RTD builds have been faster then Github. RTD builds are 10-12 min.

uvidyadharan:
> I think PRs are time bound by the rtd build, not the GitHub one.

The build for this PR took 700 seconds (11min 40sec)
image

Daltz333:
Cool PR! For now, I think we should remove the auto-tune dependency and we can merge this PR. The autotune dependency can be brought up as an issue for discussion.

uvidyadharan:
> Cool PR! For now, I think we should remove the auto-tune dependency and we can merge this PR. The autotune dependency can be brought up as an issue for discussion.

Changes made as requested

uvidyadharan:
I also tested out multithreading just the building of the pdf (not latex) and found that it had around a 1 min drop in build times. If desired I can add that to this PR.
image

prateekma (dismissed):

prateekma:
Looks good to me. We should make sure that we rename the required build check post-merge.

.github/workflows/CI.yml
```diff
@@ -33,13 +26,38 @@ jobs:
...
         path: build/latex/firstroboticscompetition.pdf
-  check-links:
+        if-no-files-found: error
+        
+  build-html:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-    - name: Checkout main
+    - name: Run Tune
```
Daltz333, line 35: Still a reference to Tune here

Daltz333 (changes requested):

uvidyadharan:
Sorry about that, changes made as requested.

Starlight220:
Why not use a matrix instead of splitting into two jobs? The steps for building HTML and PDF are nearly identical; the only differences are the build task and artifact path (which can be solved with matrix variables; the same applies to the artifact name), and the LaTeX install step (which can be solved with a conditional).
It might remove the need to change the required checks settings.

uvidyadharan:
You could if you wanted to but I think there is enough divergence in the steps to warrant them not use a matrix. Otherwise you would use a matrix for the link check step, linter and more.

Daltz333 (approved):

PR #1892, opened 2022-08-28
Add quiet flag to sphinxbuild

calcmogul:
From https://www.sphinx-doc.org/en/master/man/sphinx-build.html:
```
-q
    Do not output anything on standard output, only write warnings and errors to standard error.
```
This makes it much easier to find errors and warnings that are normally
buried in the middle of the build log.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 #
 # You can set these variables from the command line.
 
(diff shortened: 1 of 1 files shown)


Daltz333 (commented):
Needs to be added to make.bat as well

Daltz333 (approved):

PR #1894, opened 2022-08-30
Update controls-glossary.rst

Oblarg:
Going with the "less-is-more" theme, this is fewer words but more accurate.  I think it is *crucially* important to mention the link between the gaussian distribution and the concept of averaging, because that really is the fundamental reason the gaussian distribution is "special."  I dislike pointing out things as special without providing at least some hint as to why they are worth remembering.

### DIFF
--- a/source/docs/software/advanced-controls/controls-glossary.rst
+++ b/source/docs/software/advanced-controls/controls-glossary.rst
@@ -52,7 +52,7 @@ Controls Glossary
       A scalar value that relates the magnitude of an input signal to the magnitude of an output signal. In the signal-dimensional case, gain can be thought of as the proportional term of a PID controller. A gain greater than one would amplify an input signal, while a gain less than one would dampen an input signal. A negative gain would negate the input signal.
 
    Gaussian distribution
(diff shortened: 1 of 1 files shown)


gerth2:
Probably for the PR that builds on this but to spitball a footnote for the usage of Gaussian within the context of a Kalman filter:

The choice of using a gaussian function to model the "unknowns" of the system is very deliberate, and relies on the core of what makes a gaussian function special. In a real robot, noise comes from all sorts of sources. Stray electromagnetic radiation adds extra voltages to sensor readings, vibrations and temperature variations throw off inertial measurement units, gear lash causes encoders to have inaccuracies when directions change... all sorts of things. It's important to realize that, by themselves, each of these sources of "noise" aren't guaranteed to follow any pattern. Some of them might be the "white noise" random vibrations you've probably heard on the radio. Others might be "pops" or single-loop errors. Others might be nominally zero, but strongly correlated with events on the robot.

However, it can be mathematically proven that many sources of random, non-linear, unrelated noise will add up to have a _gaussian_ distribution. Since, by definition, these sources of noise are "unknowable", the best choice of a model we can make is indeed that gaussian function.

Daltz333 (approved):

PR #1897, opened 2022-09-05
Update Docker image version since 18.04 isn't supported for 2023

sciencewhiz:


Starlight220:
Tbh, wouldn't it be better from a maitainability standpoint to set this up as an RLI?

Daltz333 (approved):

PR #1914, opened 2022-09-29
Update reference to discontinued AndyMark product

nobody5050:
The AM14U4 has been replaced with the AM14U5 on andymark’s webpage

PR #1915, opened 2022-09-29
Update reference to discontinued andymark product

nobody5050:
The AM14U4 has been replaced with the AM14U5


nobody5050:
#1914 but targeting the correct branch

sciencewhiz:
This has a change we don't want in main. You'll need to cherry-pick to main, rather then merge

sciencewhiz (approved):

PR #1918, opened 2022-10-04
Move pidtune js to extension

gerth2:
Another attempt at reworking the JS build. This time, I tested locally with the frc translations repo, and it builds in both without issue.


source/conf.py
```diff
@@ -199,7 +197,7 @@
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
-html_static_path = ["_static"]
+html_static_path = ["_static", "_static_generated"]
```
TheTripleV, line 200: This would be undone with the long comment above.

source/conf.py
```diff
@@ -66,6 +63,7 @@
     "_extensions.post_process",
     "_extensions.rtd_patch",
     "_extensions.localization",
+    "_extensions.controls_js_sim.controls_js_build",
```
TheTripleV, line 66: Rename `controls_js_build.py` to `__init__.py`?
Then, this import would change to `_extensions.controls_js_sim`.
It would make it clearer which file has the setup.

TheTripleV (approved):

TheTripleV (changes requested):
didn't mean to approve

source/_extensions/controls_js_sim/controls_js_build.py
```diff
@@ -0,0 +1,86 @@
...
+def setup(app: Sphinx) -> Dict[str, Any]:
+
+    print("Generating and adding controls javascript...")
+
+    # Generate merged/minified PID tuning source
+    mergeAndMinify("source/_static_generated")
+
+    # Add interactive PID tuning
+    app.add_js_file("pid-tune.js")
+    app.add_css_file("css/pid-tune.css")
```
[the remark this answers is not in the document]
gerth2: mostly makes sense. I get what you're saying now about the relative-to-file paths, thanks for the explanation.

I am not sure what the return value is for but, I think that's ok? I'll just add it.

source/conf.py
gerth2: concur.

Do we care much to differentiate a `_static` folder which has contents which ought to be in version control and edited, and between a `_static` folder which is generated at build time? That's the only reason I tacked on the suffix.

source/conf.py
gerth2: got it.

source/conf.py
```diff
@@ -312,16 +253,16 @@ def setup(app):
 
 latex_elements = {
     "fontpkg": r"""
-	\setmainfont{DejaVu Serif}
```
gerth2, line 315: These got hit when I tried to do a tabs->spaces cleanup. It's not directly related to this activity but seems like worthwhile cleanup?

source/conf.py
gerth2: RE: CSS file - I haven't yet moved it yet for this answer.... any thoughts on where it (and the generated js files) ought to live?

TheTripleV: The _static folder doesn't matter. Any extension can dump whatever it wants in there.

The css file should go somewhere in the extension folder.
Either, you can have a pre-populated _static folder that already contains the css file and generate the js file into the same folder.
Or, you can have a _static and _static_generated folder and add them both in the app.connect.

gerth2: Got it. I think this is finalized now. Closing this out, but feel free to re-open if I ended up off in the weeds.

TheTripleV (approved):

PR #1922, opened 2022-10-07
Fix the ever-expanding diagram issue

gerth2:


gerth2:
Should be ready to rumble. Key change was changing the `css/pid-tune.js` to `pid-tune.js`. Other changes are, I believe, valid cleanups to make things less smelly. Kept them here too.

Daltz333 (approved):

PR #1923, opened 2022-10-08
Advanced Controls Tuning Wording Streamlining

gerth2:
Streamline the repeated explanation into a single common doc.

Moved some of the notes into clumped sections to prevent interrupting flow as students are learning.

Intent is only to allow students to focus on one concept at a time, but not to remove information that was already there.

Improved tuning solution dropdowns.


Oblarg (approved):
LGTM.  Great improvements overall!

PR #1924, opened 2022-10-10
Update NetworkTables documentation for NT 4

PeterJohnson:


Daltz333:
Just a ping on this, I know you're busy :D

agasser:
I have some feedback to offer. I'm just a nobody, so you can take it or leave, no hard feelings.

Overall, consider your audience. The documentation is for a user of the API for use in FRC. Implementation details are not needed. Focus on how to use the API and why someone would use the various features.

- What is the purpose of `Entry`? For the API user, is Entry just both a publisher and subscriber? Is it recommended that I use it, or is it for backward compatibility? Do I need to know that the publisher is "weak"?
- Timestamps need more detail. The documentation tells me about the time units but it's not clear what they represent; Are they the time the server received the topic? Ultimately, I want to know how it's useful in my robot program. Can I, for example, rely on the timestamp from my robot code to know when a co-processor updated a topic about a vision target?
The documentation says there's a difference between server time and local time. Please explain this.
- I think the _NetworkTables Concepts_ section should include definitions of Table, Sub-table, Topic Key, Value, and Property. An illustration might be really helpful to explain this.
- The _NetworkTables Organization_ section should focus on NetworkTables use via the API. I suggest you remove the implementation details about the NetworkTables "namespace" being flat. It contradicts the first sentence about data being organized in a hierarchy. The API, ShuffleBoard, and Outline Viewer present a hierarchy. From an API perspective, what matters is that you can access a sub-table or topic by navigating the hierarchy via `NetworkTable`, or you can access it directly using the absolute path.
- I question if the "NetworkTables API variants" section belongs in the FRC documentation. If it's not recommended to use the handle based API, why include it here? Maybe it belongs in other NetworkTables documentation instead.
- There is mention that the `close` method must be called on a publisher. It seems it would be a best practice to use try-with-resource or try/finally to ensure `close` get called, but the example does not do this. Some guidance on what to do in a robot program would be helpful as well. For example, if I have a subsystem that publishes data in `periodic`, when would I call close?

PeterJohnson:
> The examples do call `close` since they're just inline code. That leaves me conflicted, if I was going to write that code in a function where the publisher would fall out of scope, I should use try-with-resource. In a real program I'm more likely to create the publisher/subscriber during initialization and reuse it throughout the life of my program. I don't know if the statement about close is helpful, it leaves the reader with an answered question and most likely they don't actually need to worry about it. If you really feel it's needed, maybe it could be reworded to suggest the user _should_ call it in cases where they know they are done with the publisher/subscriber.

Part of the challenge is the documentation needs to cover both robot and non-robot use cases.  I agree the examples would perhaps be better written to show publishers/subscribers as instance variables, as that's going to be the much more common use case, and we also want to discourage users from trying to create publishers in their periodic loop.  Maybe then the class could show a `close()` function that calls the publisher `close()`, with a comment saying how this may not be applicable to robot code?

jasondaming (dismissed):
Didn't get through the whole thing...too tired.  Hopefully tomorrow evening

sciencewhiz (dismissed):
Should add NT 4 spec to docs/contributing/wpilib/index.rst

source/docs/software/frc-glossary.rst
```diff
@@ -53,9 +56,18 @@ FRC Glossary
    encapsulation
       A software design pattern which uses a class to hide the implementation details of other classes. See `encapsulation <https://en.wikipedia.org/wiki/Encapsulation_(computer_programming)>`__ on Wikipedia for more info.
 
+   entry
+      A combined :term:`publisher` and :term:`subscriber`. The subscriber is always active, but the publisher is not created until a publish operation is performed (e.g. a value is "set", aka published, on the entry). This may be more convenient than maintaining a separate publisher and subscriber.
```
sciencewhiz, line 60: This probably needs something tying it to NT, as entry could be many things

source/docs/software/frc-glossary.rst
```diff
@@ -93,23 +105,38 @@ FRC Glossary
...
 
+   mutable
+      An object that can be modified after it is created.
+
    permanent-magnet DC motor
       The classification of all legal motors for the FIRST robotics competition. This type of motor takes direct current as input, and uses it to create a magnetic field. In turn, this magnetic field interacts with a physical magnet to create a force that turns the output shaft. Electrical ("brushless") or mechanical ("brushed") means are used to ensure the electrically-generated magnetic field always points in a direction that creates forces when it interacts with the physical magnet, even as the motor's shaft rotates. See `permanent-magnet motor <https://en.wikipedia.org/wiki/Brushed_DC_electric_motor#Permanent-magnet_motors>`__ on Wikipedia for more info.
 
+   persistent
+      A :term:`topic` that is saved to a file by the server and restored at startup.
```
sciencewhiz, line 123: Same comment about mentioning NT for the next few entries

source/docs/software/frc-glossary.rst
```diff
@@ -93,23 +105,38 @@ FRC Glossary
...
+   publisher
+      Defines a :term:`topic` and creates and sends timestamped data values.
+
    retro-reflection
       The property of reflecting incoming light back at the same angle it came in at, rather than an incident angle (like a mirror), absorbing it, or scattering it. Most FRC vision processing targets are retro-reflective. See `retroreflector <https://en.wikipedia.org/wiki/Retroreflector>`_ on Wikipedia for more information.
 
    recursive composition
       A type of :term:`composition` in which the composite object may contain components of the same type as itself. For example, a command group may contain one or more command groups. See `recursive composition <https://en.wikipedia.org/wiki/Object_composition#Recursive_composition>`__ on Wikipedia for more info. See also :term:`recursive composition`.
 
+   retained
```
sciencewhiz, line 137: same comment about NT

source/docs/software/networktables/client-side-program.rst
```diff
@@ -3,47 +3,132 @@
 Creating a client-side program
```
sciencewhiz, line 3: The gradle code block should be updated for additional platformas (at least macOS Arm

source/docs/software/networktables/nt4-migration-guide.rst
```diff
@@ -0,0 +1,165 @@
+Migrating from NetworkTables 3.0 to NetworkTables 4.0
+=====================================================
+
+NetworkTables 4.0 (new for 2023) has a number of significant API breaking changes from NetworkTables 3.0, the version of NetworkTables used from 2016-2022.
+
+NetworkTableEntry
+-----------------
+
+While ``NetworkTableEntry`` can still be used (for backwards compatibility), users are encouraged to migrate to use of type-specific Publisher/Subscriber/Entry classes as appropriate, or if necessary, ``GenericEntry``. It's important to note that unlike ``NetworkTableEntry``, these classes need to have appropriate lifetime management. Some functionality (e.g. persistent settings) has also moved to ``Topic`` properties.
```
sciencewhiz, line 9: recommend adding links to the API docs

source/docs/software/networktables/nt4-migration-guide.rst
```diff
@@ -0,0 +1,165 @@
...
+
+NT3 code (was):
```
sciencewhiz, line 11: @Daltz333 is there a way to put the code examples side by side on wider screens?

source/docs/software/networktables/reading-array-values-published-by-networktables.rst
```diff
@@ -1,50 +1,37 @@
...
 
-Very often the values are for one or more areas of interest such as goals or game pieces and multiple instances are returned. In
-the example below, several x, y, width, height, and areas are returned by the image processor and the robot program can sort out
-which of the returned values are interesting through further processing.
+Very often the values are for one or more areas of interest such as goals or game pieces and multiple instances are returned. In the example below, several x, y, width, height, and areas are returned by the image processor and the robot program can sort out which of the returned values are interesting through further processing.
 
 Verify the NetworkTables keys being published
 ---------------------------------------------
 .. image:: images/reading-array-values-published-by-networktables-1.png
```
sciencewhiz, line 9: need a new outline viewer picture

source/docs/software/networktables/reading-array-values-published-by-networktables.rst
```diff
@@ -1,50 +1,37 @@
...
    :alt: Image of OutlineViewer with the NetworkTables keys
 
-You can verify the names of the NetworkTables keys used for publishing the values by using the Outline Viewer application.
-It is a Java program in your user directory in the wpilib/<YEAR>/tools folder. The application is started by selecting the "WPILib" menu
-in Visual Studio Code then Start Tool then "OutlineViewer". In this example, with the image processing program running (GRIP) you can see the values being put
-into NetworkTables.
+You can verify the names of the NetworkTables keys used for publishing the values by using the Outline Viewer application. It is a Java program in your user directory in the wpilib/<YEAR>/tools folder. The application is started by selecting the "WPILib" menu in Visual Studio Code then Start Tool then "OutlineViewer". In this example, with the image processing program running (GRIP) you can see the values being put into NetworkTables.
```
sciencewhiz, line 12:
```suggestion
You can verify the names of the NetworkTables keys used for publishing the values by using the Outline Viewer application. It is a C++ program in your user directory in the wpilib/<YEAR>/tools folder. The application is started by selecting the "WPILib" menu in Visual Studio Code then Start Tool then "OutlineViewer". In this example, with the image processing program running (GRIP) you can see the values being put into NetworkTables.
```

source/docs/software/networktables/networktables-intro.rst
```diff
@@ -1,18 +1,64 @@
...
+:term:`NetworkTables` is an implementation of a `publish-subscribe messaging system <https://en.wikipedia.org/wiki/Publish%E2%80%93subscribe_pattern>`_. Values are published to named "topics" either on the robot, driver station, or potentially an attached coprocessor, and the values are automatically distributed to all subscribers to the topic. For example, a driver station laptop might receive camera images over the network, perform some vision processing algorithm, and come up with some values to sent back to the robot. The values might be an X, Y, and Distance. By writing these results to NetworkTables topics called "X", "Y", and "Distance" they can be read by the robot shortly after being written. Then the robot can act upon them. Similarly, the robot program can write sensor values to topics and those can be read and plotted in real time on a dashboard application.
```
sciencewhiz, line 4: It's probably worth adding a note about NT changing for 2023

source/docs/software/networktables/client-side-program.rst
PeterJohnson: Let's wait to do this since it looks like we're migrating to universal binaries.

source/docs/software/networktables/nt4-migration-guide.rst
PeterJohnson: Added in various other pages; added cross-links here to the new docs.

source/docs/software/networktables/listening-for-change.rst
```diff
@@ -1,57 +1,308 @@
...
+                // get the subtable called "datatable"
+                NetworkTable datatable = inst.getTable("datatable");
+
+                // subscribe to the topic in "datatable" called "Y"
+                ySub = datatable.getDoubleTopic("Y").subscribe(0.0);
+
+                // add a listener to only value changes on the Y subscriber
+                valueListenerHandle = inst.addListener(
+                    ySub,
+                    EnumSet.of(NetworkTableEvent.kValueAll),
```
gcschmit, line 191: While updating our 2022 code to 2023 wpilib beta 3, I referenced this example. I had to specify NetworkTableEvent.Kind.kValueAll since kValueAll is in the Kind enum.

TheTripleV:
There should a blurb near the migration guide explaining the differences between NT3 and NT4 and why NT4 is better (an upgrade)

source/docs/software/networktables/networktables-intro.rst
```diff
@@ -39,89 +87,16 @@ There are some default tables that are created automatically when the program st
...
+In Java, a common bug is to create a subscriber or publisher and not properly release it by calling ``close()``, as this will result in the object lingering around for an unknown period of time until it is garbage collected. This is less common of an issue in robot programs, as long as the publisher or subscriber is stored in an instance variable that persists for the life of the program.
 
-.. note:: Actually NetworkTables has a flat namespace for the keys. Having tables and subtables is an abstraction to make it easier to organize your data. So for a table called "SmartDashboard" and a key named "xValue", it is really a single key called "/SmartDashboard/xValue". The hierarchy is not actually represented in the distributed data, only keys with prefixes that are the contained table.
+In C++, publishers, subscribers, and entries are RAII, which means they are automatically destroyed when they go out of scope. ``NetworkTableInstance`` is an exception to this; it is designed to be explicitly destroyed, so it's not necessary to maintain a global instance of it.
```
rzblue, line 102: We should add a term definition for RAII

source/docs/software/networktables/nt4-migration-guide.rst
Daltz333: Use `sphinx_design` or `sphinx_panels` for a dynamic grid

TheTripleV: > @Daltz333 is there a way to put the code examples side by side on wider screens?

Like https://discord.com/channels/176186766946992128/368993897495527424/797728025113788468 ?

source/docs/software/networktables/client-side-program.rst
```diff
@@ -3,47 +3,132 @@
...
 ==============================
 
-If all you need to do is have your robot program communicate with GRIP or a dashboard running on the Driver Station laptop, then the previous examples of writing robot programs are sufficient. But if you would like to write some custom client code that would run on the drivers station or on a coprocessor then you need to know how to build :term:`NetworkTables` programs for those (non-roboRIO) platforms.
+If all you need to do is have your robot program communicate with a COTS coprocessor or a dashboard running on the Driver Station laptop, then the previous examples of writing robot programs are sufficient. But if you would like to write some custom client code that would run on the drivers station or on a coprocessor then you need to know how to build :term:`NetworkTables` programs for those (non-roboRIO) platforms.
```
Daltz333, line 6:
```suggestion
If all you need to do is have your robot program communicate with a :term:`COTS` coprocessor or a dashboard running on the Driver Station laptop, then the previous examples of writing robot programs are sufficient. But if you would like to write some custom client code that would run on the drivers station or on a coprocessor then you need to know how to build :term:`NetworkTables` programs for those (non-roboRIO) platforms.
```

source/docs/software/networktables/networktables-intro.rst
```diff
@@ -1,18 +1,66 @@
...
+:term:`NetworkTables` is an implementation of a `publish-subscribe messaging system <https://en.wikipedia.org/wiki/Publish%E2%80%93subscribe_pattern>`_. Values are published to named "topics" either on the robot, driver station, or potentially an attached coprocessor, and the values are automatically distributed to all subscribers to the topic. For example, a driver station laptop might receive camera images over the network, perform some vision processing algorithm, and come up with some values to sent back to the robot. The values might be an X, Y, and Distance. By writing these results to NetworkTables topics called "X", "Y", and "Distance" they can be read by the robot shortly after being written. Then the robot can act upon them. Similarly, the robot program can write sensor values to topics and those can be read and plotted in real time on a dashboard application.
 
-NetworkTables can be used by programs on the robot in either C++, Java or LabVIEW and is built into each version of WPILib.
+NetworkTables can be used by programs on the robot in C++, Java, or LabVIEW, and is built into each version of WPILib.
```
Daltz333, line 6:
```suggestion
NetworkTables can be used by programs on the robot in Java, C++, or LabVIEW, and is built into each version of WPILib.
```

source/docs/software/networktables/networktables-intro.rst
```diff
@@ -1,18 +1,66 @@
...
+
+.. note: NetworkTables has changed substantially in 2023. For more information on migrating pre-2023 code to use the new features, see :ref:`docs/software/networktables/nt4-migration-guide:migrating from networktables 3.0 to networktables 4.0`.
+
+NetworkTables concepts
```
Daltz333, line 10: We should default to title case for section titles when it makes sense

```suggestion
NetworkTables Concepts
```

source/docs/software/networktables/publish-and-subscribe.rst
```diff
@@ -0,0 +1,911 @@
+Publishing and Subscribing to a Topic
+=====================================
+
+Publishing to a topic
+---------------------
+
+In order to create a :term:`topic` and publish values to it, it's necessary to create a :term:`publisher`.
+
+NetworkTable publishers are represented as type-specific Publisher objects (e.g. ``BooleanPublisher``: `Java <https://github.wpilib.org/allwpilib/docs/beta/java/edu/wpi/first/networktables/BooleanPublisher.html>`__, `C++ <https://github.wpilib.org/allwpilib/docs/beta/cpp/classnt_1_1_boolean_publisher.html>`__). The publisher object must be retained for the publisher to continue publishing. In Java, the ``close()`` method needs be called to ensure publishing is stopped; in C++ this is handled by the destructor. C++ publishers are moveable and non-copyable.
```
Daltz333, line 9: This paragraph doesn't read cleanly. 

> The publisher object must be retained for the publisher to continue publishing

Can this sentence be reworded?

Daltz333 (changes requested):
After reading through this and the examples, I have a few concerns.

1. The language is very stiff. There are areas where sentences are in the same paragraph, but don't seem to have transitional vocabulary (starting a sentence with `it's`)
2. Are the examples tested? If not, they should.
3. I personally think the API for publishing and subscribing is more verbose. For example, my team uses around ~15 different names that would now be a `Topic` that we would want to subscribe or publish. This would require a 4 function chained call or storing a lot of variables we use in only one place.

PeterJohnson:
> * The language is very stiff. There are areas where sentences are in the same paragraph, but don't seem to have transitional vocabulary (starting a sentence with `it's`)

Recommendations in this area are welcome.  My tendency is to use a stiffer style, rather than more conversational (e.g. I avoid "you").

> * Are the examples tested? If not, they should.

No, they aren't.  So there could easily be typos.

> * I personally think the API for publishing and subscribing is more verbose. For example, my team uses around ~15 different names that would now be a `Topic` that we would want to subscribe or publish. This would require a 4 function chained call or storing a lot of variables we use in only one place.

Yes, it's more verbose than getEntry() was, but part of that is to capture the type information, and we have more options (publish/subscribe/entry).  So we go from `table.getEntry("blah")` to `table.getDoubleTopic("blah").publish()` (or `.getEntry()` or `.subscribe()`).

Daltz333:
The only thing besides my explicit review comments to merge this would be to ensure the examples are tested. The rest can be iteration in additional PRs

source/docs/software/networktables/publish-and-subscribe.rst
PeterJohnson: Maybe "Publishers are only active as long as the Publisher object exists. Typically you want to keep publishing longer than the local scope of a function, so it's necessary to store the Publisher object somewhere longer term, e.g. in an instance variable."?

source/docs/software/networktables/networktables-intro.rst
PeterJohnson: Is this to be consistent with other places in the docs?

Daltz333: Yeah.

rzblue:
One of the things that wasn't clear to me was what does "continue publishing" mean for the robot program? Does stopping publishing mean that the value disappears entirely (if not marked retained)? I think teams will get tripped up on this since it differs from the old behavior

PeterJohnson:
That's exactly correct.  Unless a topic is marked "retained", it disappears when there are no remaining publishers (in the network, e.g. robot program, dashboard, coprocessors).

Daltz333:
What's the reasoning for not making `retained` the default?

rzblue:
Ok- makes sense to me. I think something along those lines clarifying that would help. You can figure it out by connecting a couple different sections of the docs but it isn't immediately clear.

rzblue:
> What's the reasoning for not making `retained` the default?


Not sure Peter's reasoning, but the old behavior would retain values for the lifetime of the NT instance- values on dashboards from previous runs would propagate to the server and you could end up with "ghost values" in certain situations

PeterJohnson:
Ryan hit the nail on the head for why retained is not the default--it's super easy to get stale values.  NT3 used various tricks to try to prevent this from happening, but it's much better to just be explicit about the (relatively rare) instances you actually *want* things to be retained (e.g. configuration values like autonomous mode selection inputs).

PeterJohnson:
There's a section on the intro page about retained and persistent topics (https://frc-docs--1924.org.readthedocs.build/en/1924/docs/software/networktables/networktables-intro.html#retained-and-persistent-topics), but maybe it needs to be moved and linked from elsewhere?

source/docs/software/networktables/nt4-migration-guide.rst
Daltz333: Unfortunately, it doesn't look good due to the small (1000px) horizontal width of the website.

Daltz333 (approved):

PR #1944, opened 2022-10-31
Explain how to launch DS Log Viewer from Driver Station

bovlb:
This seems a much easier method for teams to use.


bovlb:
I don't know why this PR is pulling in someone else's changes.

Starlight220:
It pulls in the changes because the branch is created on `stable` and then PRd against `main`. The branch should be created on `main`.

bovlb:
> It pulls in the changes because the branch is created on `stable` and then PRd against `main`. The branch should be created on `main`.

Yes, I worked that out eventually, thanks.  Clicking on the live website takes you to `stable`, which is what confused me.  I was able to resubmit my changes as PR #1945.

PR #1953, opened 2022-11-03
Revert "Update Inspector to v1.2 (#1939)"

sciencewhiz:
This reverts commit cb2e3baf1b44389580ef0a3fcceaa03f7b5bc90f.


Starlight220:
Note that the version needs to be updated in the JSON.

sciencewhiz:
It needs to be done as two steps since GH actions won't run the new stuff until merged.

Starlight220:
Note that the old Inspector version will have all the deprecation warnings.

sciencewhiz:
Yes, but the things deprecated don't get removed until mid next year

PR #1954, opened 2022-11-03
Update API doc links to github

sciencewhiz:
Also switch some links from development to beta that were missed previously


TheTripleV:
The beta github url also has to go in https://github.com/wpilibsuite/frc-docs/blob/main/source/_static/js/api-docs-redirect.js.

TheTripleV (approved):

PR #1969, opened 2022-11-11
Update Trigger docs for HID factories and CommandPtr

Starlight220:


Starlight220:
Where's the bad link syntax?

source/docs/software/commandbased/binding-commands-to-triggers.rst
```diff
@@ -5,14 +5,16 @@ Apart from autonomous commands, which are scheduled at the start of the autonomo
...
+.. note:: The C++ command-based library offers two overloads of each button binding method - one that takes an `rvalue reference<http://thbecker.net/articles/rvalue_references/section_01.html>`__ (``CommandPtr&&``), and one that takes a raw pointer (``Command*``).  The rvalue overload moves ownership to the scheduler, while the raw pointer overload leaves the user responsible for the lifespan of the command object.  It is recommended that users preferentially use the rvalue reference overload unless there is a specific need to retain a handle to the command in the calling code.
```
sciencewhiz, line 13: This indicates the lack of space before the opening bracket could cause the link error. https://superuser.com/questions/1669143/cryptic-error-from-sphinx-anonymous-hyperlink-mismatch

```suggestion
.. note:: The C++ command-based library offers two overloads of each button binding method - one that takes an `rvalue reference <http://thbecker.net/articles/rvalue_references/section_01.html>`__ (``CommandPtr&&``), and one that takes a raw pointer (``Command*``).  The rvalue overload moves ownership to the scheduler, while the raw pointer overload leaves the user responsible for the lifespan of the command object.  It is recommended that users preferentially use the rvalue reference overload unless there is a specific need to retain a handle to the command in the calling code.
```

sciencewhiz (approved):

sciencewhiz (approved):

PR #1978, opened 2022-11-13
Update old command-based migration guide

Starlight220:


sciencewhiz (commented):
I'm thinking the Trigger renames should be here too, since a team updating would need to deal with those also.

Starlight220:
I doubt that the Trigger API didn't change for v2, and there isn't anything about it all in the doc.

sciencewhiz:
Not specifically Trigger, but you don't really use commands without JoystickButton

Daltz333 (approved):
We should aim to remove this for 2024

PR #1979, opened 2022-11-14
Declare static import for Commands factories

democat3457:
It is currently unclear where `parallel` and `sequence` are from, but the addition of a static import will make it clear.


Starlight220:
In a documentation example, I think it's preferable to qualify the method calls rather than adding a static star import.

democat3457:
@Starlight220 how’s that?

Starlight220:
The import is redundant,  and maybe remove the in-paragraph comment as well -- we might add factories in other places.

Starlight220 (approved):

PR #1995, opened 2022-11-20
Added WPILib installer download instructions

### LINKED ISSUE #1992: Mention the Download is in the Assets section of the linked Github page
https://docs.wpilib.org/en/latest/docs/zero-to-robot/step-2/wpilib-setup.html

https://www.chiefdelphi.com/t/download-location-of-wpilib/418276

itsmichaelyu:
Added instructions for downloading the WPILib installer to help people who might be unused to installing off of GitHub. Also fixes #1992.


daltzctr:
This needs to be duplicated on the `Prerequisites` page as well

Daltz333:
This change needs to be done to the prerequisites page as well

itsmichaelyu:
> This change needs to be done to the prerequisites page as well

What if the offline installation preparation WPILib GitHub link was changed to direct to the more [detailed guide](https://docs.wpilib.org/en/stable/docs/zero-to-robot/step-2/wpilib-setup.html)?

Daltz333:
If we do that, we should do it the other way around. With the WPILib guide redirecting to prerequisites page.

itsmichaelyu:
If the offline pages links to the WPILib install, then we could have the more detailed instructions for running the installer, which is just a note right now.

Screenshot 2022-11-20 at 2 47 16 PM

Then another note could be added here, to say something like "If you want to create future offline installation click the **Create VS Code zip to share with other computers/OSes for offline install**"

Screenshot 2022-11-20 at 2 47 05 PM

Daltz333:
I'd argue that note belongs in the WPILib setup article and not there. Since prerequisites are the details of *downloading* and each installer has their respective installation article.

itsmichaelyu:
I think it might be better to just put it in both pages, as people might be confused by being redirected to a page called "Offline Installation Preparation".

sciencewhiz (approved):

PR #1996, opened 2022-11-21
Recommending PathPlanner over PathWeaver

jasondaming:


Starlight220:
Maybe explain why PathPlanner is better than PathWeaver?

Daltz333 (approved):

PR #2012, opened 2022-11-27
Rename pose estimators article

calcmogul:


sciencewhiz (changes requested):
Needs a redirect added to source/redirects.txt

sciencewhiz (approved):

PR #2015, opened 2022-11-28
Fix minor errors in changelog doc

prateekma:


sciencewhiz:
I think there's probably other references to m1 that should be updated throughout the docs.

sciencewhiz (approved):

PR #2019, opened 2022-12-01
Add Python RLI for mecanum drive

virtuald:


sciencewhiz:
@Starlight220 should there be updates to inspector to handle other URLs?

Starlight220:
It'll need a completely separate config JSON and everything.
I wonder how it'll be best to set things up, because merging the reports won't work with the current version of Inspector.

I can try to work it into Inspector, but that would likely take more time. I don't have a preference between upgrading Inspector for it vs adding the additional JSON and everything here, your choice.

source/docs/software/hardware-apis/motors/wpi-drive-classes.rst
```diff
@@ -321,6 +321,14 @@ MecanumDrive is a method provided for the control of holonomic drivetrains with
       :linenos:
       :lineno-start: 31
 
+  .. tab:: Python
+
+    .. remoteliteralinclude:: https://raw.githubusercontent.com/robotpy/examples/main/mecanum-drive/robot.py
```
Starlight220, line 326: Regardless of Inspector, this should use a pinned version rather than `main`.

Starlight220 (changes requested):

source/docs/software/hardware-apis/motors/wpi-drive-classes.rst
virtuald: We don't have versions for our examples? But I suppose we can do that. What version do you want?

Starlight220: Like the allwpilib examples, it should be a git tag or commit hash.

sciencewhiz: The issue is that examples may be updated and if lines are added or removed the frc-docs RLI will point to junk.

Starlight220:
After a rebase, the `inspector.json` will need to be updated with another object in the `sources` array.

Daltz333:
Just a ping.

virtuald:
Added a tag.

Daltz333 (approved):

PR #2034, opened 2022-12-10
Stop photofinish in htmlzip

TheTripleV:


sciencewhiz:
Is there a way to test this locally or on RTD? When I build locally, it looks like the webp images aren't generated even before this change. It also doesn't look like the PR build on RTD has the htmlzip.

sciencewhiz:
After setting `photofinish_ci_only = False`, I was able to confirm it worked locally

PeterJohnson (approved):

PR #2041, opened 2022-12-13
Fix capitalization of method name

bovlb:


Starlight220:
It's `GetDouble` in C++

Daltz333 (approved):

Daltz333:
We should prefer Java naming practice unless we are explicitly dealing with C++.

PR #2055, opened 2022-12-26
Remove ghcontributors extension due to API limits

Daltz333:
I decided to not remove the extension dependency in this PR as my local environment is configured with Python 3.11 and poetry wants to force a minimum python version to 3.9.


Daltz333:
Closing in favor of #2056

PR #2060, opened 2022-12-27
Remove motor controller encoder example

Daltz333:


Starlight220 (commented):
Add the mention of an encoder to the MotorControl example

Starlight220 (approved):

PR #2075, opened 2023-01-03
Add article explaining functions as data

Oblarg:
A bunch of recent feedback suggests that the poor pedagogical quality of Java's official lambda expression documentation is a significant limiting factor to the ability of students to use functional callback-based API features.

This will serve as a reference for people struggling with the concept.


source/docs/software/basic-programming/functions-as-data.rst
```diff
@@ -0,0 +1,71 @@
+Functions as Data
+=================
+
+Regardless of programming language, one of the first things anyone learns to do when programming a computer is to write a function (also known as a "method" or a "subroutine").  Functions are a fundamental part of organized code - writing functions lets us avoid duplicating the same piece of code over and over again.  Instead of writing duplicated sections of code, we call a single function that contains the code we want to execute from multiple places.  If the section of code needs some additional information about its surrounding context to run, we pass those to the function as parameters.
```
calcmogul, line 4: I agree with the deduplication argument, but functions are also good for behavioral abstraction, even if they're used once. Rather than writing a giant function with comments above each block of behavior, you can move them into well-named functions so the high-level control flow is easier to discern.

Oblarg: for sure, this isn't an exhaustive list of reasons why functions are good

source/docs/software/basic-programming/functions-as-data.rst
```diff
@@ -0,0 +1,71 @@
...
+^^^^^^^^^^^^^^^^^
+
+A method reference lets us pass an already-existing function as our ``Runnable``:
+
+.. code-block:: java
+
+   // Create an InstantCommand that runs the `resetEncoders` method of the `drivetrain` object
+   Command disableCommand = runOnce(drivetrain::resetEncoders, drivetrain);
+
+The expression ``drivetrain::resetEncoders`` is a reference to the ``resetEncoders`` method of the ``drivetrain`` object.  It is not a method *call* - this line of code does not *itself* reset the encoders of the drivetrain.  Instead, it returns a ``Command`` that will do so *when it is scheduled.*
```
gcschmit, line 39: This paragraph is absolutely critical. Students are so accustomed to invoking methods, despite the different syntax, they don't appreciate that the resetEncoders method is not being invoked at this time.

source/docs/software/basic-programming/functions-as-data.rst
```diff
@@ -0,0 +1,71 @@
...
+
+Sometimes, we need to pass functions from one part of the code to another part of the code.  This might seem like a strange concept, if we're used to thinking of functions as part of a class definition rather than objects in their own right.  But at a basic level, functions are just data - in the same way we can store an ``integer`` or a ``double`` as a variable and pass it around our program, we can do the same thing with a function.  A variable whose value is a function is called a "functional interface" in Java, and a "function pointer" or "functor" in C++.
+
+Why Would We Want to Treat Functions as Data?
+---------------------------------------------
+
+Typically, we have a need to treat functions as data when the function needs to be called from somewhere other than where it was defined in the code.
```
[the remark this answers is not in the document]
gcschmit: Typically, code that calls a function is coupled (depends on) the definition of the function. While this occurs all the time, it becomes problematic when the code calling the function is developed independently (for example, WPILib) and without direct knowledge of the function definition (for example, code from an FRC team). Sometimes we solve this challenge through the use of class interfaces to keep code decoupled. However, if just a single function is involved, passing the function as data is a more convenient solution.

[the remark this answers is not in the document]
Oblarg: I like that, and will incorporate it.

source/docs/software/basic-programming/functions-as-data.rst
Oblarg: This is a problem that extends outside of computer science; people in general tend to conflate `f` and `f(x)` all the time.  Doing this can be a pretty terrible category error and lead to all sorts of confusion.

source/docs/software/basic-programming/functions-as-data.rst
```diff
@@ -0,0 +1,118 @@
...
+
+This means we can only capture primitive types (like ``int``, ``double``, and ``boolean``) if they're constants.  If we want to capture a state variable that can change, it *must be wrapped in a mutable object*.
+
+Treating Functions as Data in C++
+---------------------------------
+
+C++ has a number of ways to treat functions as data.  For the sake of this article, we'll only talk about the parts that are relevant to using WPILibC.
+
+In WPILibC, function types are represented with the ``std::function`` class <TODO: link>.  This standard library class is templated on the function's signature - that means we have to provide it a function pointer type <TODO: link> as a template parameter to specify the signature of the function (compare this to Java above, where we have a separate interface type for each kind of signature).
```
calcmogul, line 84: The `double(double)` in `std::function<double(double)>` is not a function pointer type; it's an actual function type. A function pointer type equivalent would be `double(*)(double)`.

source/docs/software/basic-programming/functions-as-data.rst
```diff
@@ -0,0 +1,118 @@
...
+
+Lambda Expressions in C++
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+To pass a function to ``runOnce``, we need to write a short inline function expression using a special syntax that resembles ordinary C++ function declarations, but varies in a few important ways:
+
+.. code-block:: cpp
+
+   // Create an InstantCommand that runs the drive forward at half speed
+   CommandPtr driveHalfSpeed = cmd::runOnce(() [this] { drivetrain.arcadeDrive(0.5, 0.0); }, {drivetrain});
```
calcmogul, line 105:
```suggestion
   CommandPtr driveHalfSpeed = cmd::RunOnce([this] { drivetrain.ArcadeDrive(0.5, 0.0); }, {drivetrain});
```

source/docs/software/basic-programming/functions-as-data.rst
```diff
@@ -0,0 +1,118 @@
...
+
+C++ calls ``() [this] { drivetrain.arcadeDrive(0.5, 0.0); }`` a "lambda expression".  It has three parts: a *parameter list* (parentheses), a *capture list* (square brackets), and a *function body* (curly brackets).  It may look a little strange, but the only real difference between a lambda expression and an ordinary function (apart from the lack of a function name) is the addition of the capture list.
```
calcmogul, line 107:
```suggestion
C++ calls ``[captures] (params) { body; }`` a "lambda expression".  It has three parts: a *capture list* (square brackets), an optional *parameter list* (parentheses), and a *function body* (curly brackets).  It may look a little strange, but the only real difference between a lambda expression and an ordinary function (apart from the lack of a function name) is the addition of the capture list.
```

source/docs/software/basic-programming/functions-as-data.rst
```diff
@@ -0,0 +1,118 @@
...
+
+Since ``runOnce`` wants a function with no parameters and no return value, our lambda expression has an empty parameter list and no return statement.  The "lambda expression" here represents a function that calls ``drivetrain.arcadeDrive`` with a specific set of parameters - note again that this does not *call* the function, but merely defines it and passes it to the ``Command`` to be run later when the ``Command`` is scheduled.
```
calcmogul, line 109:
```suggestion
Since ``RunOnce`` wants a function with no parameters and no return value, our lambda expression has an empty parameter list and no return statement.  The "lambda expression" here represents a function that calls ``drivetrain.ArcadeDrive`` with a specific set of parameters - note again that this does not *call* the function, but merely defines it and passes it to the ``Command`` to be run later when the ``Command`` is scheduled.
```

source/docs/software/basic-programming/functions-as-data.rst
```diff
@@ -0,0 +1,118 @@
...
+
+C++ has somewhat more-powerful semantics than Java.  One cost of this is that we generally need to give the C++ compiler some help to figure out *how exactly* we want it to capture state from the enclosing scope.  This is the purpose of the *capture list*.  For the purposes of using the WPILibC Command-based framework, it is usually sufficient to use a capture list of ``[this]``, which gives access to members of the enclosing class by capturing the enclosing class's ``this`` pointer by value.
+
+Method locals cannot be captured this way, and must be captured explicitly either by reference or by value by including them in the capture list.  It is typically safer to capture locals by-value, since a lambda can outlive the lifespan of an object it captures by reference.  For more details, consult the C++ standard library documentation on capture semantics <TODO: link>.
```
calcmogul, line 118: Here's the link on capture semantics: https://en.cppreference.com/w/cpp/language/lambda#Lambda_capture. The top of the page describes the various syntaxes for lambdas too.

calcmogul (changes requested):

source/docs/software/basic-programming/functions-as-data.rst
```diff
@@ -0,0 +1,118 @@
...
+This means we can only capture primitive types (like ``int``, ``double``, and ``boolean``) if they're constants.  If we want to capture a state variable that can change, it *must be wrapped in a mutable object*.
+
+Treating Functions as Data in C++
+---------------------------------
+
+C++ has a number of ways to treat functions as data.  For the sake of this article, we'll only talk about the parts that are relevant to using WPILibC.
+
+In WPILibC, function types are represented with the ``std::function`` class <TODO: link>.  This standard library class is templated on the function's signature - that means we have to provide it a function pointer type <TODO: link> as a template parameter to specify the signature of the function (compare this to Java above, where we have a separate interface type for each kind of signature).
+
+This sounds a lot more complicated than it is to use in practice.  Let's look at the call signature of ``cmd::runOnce`` (which creates an ``InstantCommand`` that, when scheduled, runs the given function once and then terminates):
```
calcmogul, line 86:
```suggestion
This sounds a lot more complicated than it is to use in practice.  Let's look at the call signature of ``cmd::RunOnce`` (which creates an ``InstantCommand`` that, when scheduled, runs the given function once and then terminates):
```

source/docs/software/basic-programming/functions-as-data.rst
```diff
@@ -0,0 +1,118 @@
...
+The expression ``drivetrain::resetEncoders`` is a reference to the ``resetEncoders`` method of the ``drivetrain`` object.  It is not a method *call* - this line of code does not *itself* reset the encoders of the drivetrain.  Instead, it returns a ``Command`` that will do so *when it is scheduled.*
+
+Remember that in order for this to work, ``resetEncoders`` must be a ``Runnable`` - that is, it must take no parameters and return no value.  So, its signature must look like this:
+
+.. code-block:: java
+
+   // void because it returns no parameters, and has an empty parameter list
+   public void resetEncoders()
+
+If the function signature does not match this, Java will not be able to interpret the method reference as a ``Runnable`` and the code will not compile.
```
shueja-personal, line 54: This is accurate-but-not-really. In this case, the method reference must be a `Runnable` because `runOnce` expects a `Runnable`.  Other common usecases like `Trigger` constructors require `BooleanSuppliers` so the method would have to be `public boolean getSpeed()` in order for `new Trigger(shooter::getSpeed)` to compile.

source/docs/software/basic-programming/functions-as-data.rst
```diff
@@ -0,0 +1,118 @@
...
+
+In Java capturing state is a fairly safe thing to do in general, with one major caveat: we can only capture state that is "effectively final".  That means it is only legal to capture a variable from the enclosing scope if that variable is never reassigned after initialization.  Note that this does not mean that the captured state cannot change: Remember that Java objects are references, so the object that the reference *points to* may change after capture - but the reference itself cannot be made to point to another object.
+
+This means we can only capture primitive types (like ``int``, ``double``, and ``boolean``) if they're constants.  If we want to capture a state variable that can change, it *must be wrapped in a mutable object*.
```
shueja-personal, line 77: Will violating this lead to compilation failure or the lambda not picking up subsequent changes to the value? I think VS Code warns about it sometimes but I don't recall if it fails compilation. Probably worth including a bit on how someone can know they messed this up.

source/docs/software/basic-programming/functions-as-data.rst
Oblarg: The case given is that of a `Runnable` in particular, and the text says as much; other functional interfaces can be discovered by readers in time.

source/docs/software/basic-programming/functions-as-data.rst
Oblarg: It'll lead to compilation failure.

source/docs/software/basic-programming/functions-as-data.rst
```diff
@@ -0,0 +1,118 @@
...
+
+Why Would We Want to Treat Functions as Data?
+---------------------------------------------
+
+Typically, code that calls a function is coupled to (depends on) the definition of the function. While this occurs all the time, it becomes problematic when the code *calling* the function (for example, WPILib) is developed independently and without direct knowledge of the code that *defines* the function (for example, code from an FRC team). Sometimes we solve this challenge through the use of class interfaces, which define collections of data and functions that are meant to be used together.  However, often we really only have a dependency on a *single function*, rather than on an *entire class*.
+
```
jwhite66, line 12: I feel like this why is a little bit too dry and something a student might have a hard time following.  How about injecting something like this:
For example, WPILib provides great facilities for handling joysticks.  It handles a lot of  the problems of tracking button state and watching for a button press and release.  But then we  want to write code that achieves something like:  'When this button is pressed, move the arm up'.  So if we have a function that 'moves the arm up' it is very powerful if we can pass that function in to the WPILib joystick methods.

Oblarg: I like that a lot, and will incorporate it.

tom131313:
The interesting thing (as in potentially slightly confusing) is in the example of `runOnce()` needing a `Runnable` argument is the argument doesn't have have to be defined as `Runnable` say using the `implements Runnable` on the `class`, if it's a method reference. The argument is implicitly defined `Runnable` because the takes no arguments and returns no value. That is stated almost but not exactly that way.

My confusion was I knew there is a `Runnable` type but where is it?

If the argument is a named (through a variable) lambda then the variable/object does have to be defined as a `Runnable`.

Adding just a few words pointing out implicit vs explicit syntax would have helped me and save a lot of time figuring it out (I'm a Java novice like the students).
```
 public void testMethod(){}
 public Runnable testLambda = () -> {};
 RunCommand aa = new RunCommand(driveSubsystem::testMethod, driveSubsystem);
 RunCommand bb = new RunCommand(driveSubsystem.testLambda, driveSubsystem);
 ```

File: source/docs/software/basic-programming/functions-as-data.rst (hunk 12 above)
Comment (ngreen-gt) [reply source/docs/software/basic-programming/functions-as-data.rst:12]: Generally, I felt like describing what a function is (the first paragraph) was something someone should understand if they are reading this page, so you could do without that whole paragraph.  Starting just with the second paragraph gets more to the point of functions as data.  It may be appropriate to make the first mention of function there have a mouse-over definition, to make clear that methods and functions are the same.

Comment (jwhite66) [reply source/docs/software/basic-programming/functions-as-data.rst:12]: I disagree.  I liked that paragraph.  The whole point of this is that WPILib is using Lambda expressions in the very first example code a student programmer might use.   If anything, I feel that this article should be studied and made sure it is comprehensible to someone with zero previous programming experience.

**Comment by Oblarg** [comment]
@tom131313 I've changed up the language to be more explicit about this - thanks for the observation!

File: source/docs/software/basic-programming/functions-as-data.rst [hunk 13]
```diff
@@ -0,0 +1,120 @@
...
+---------------------------------
+
+C++ has a number of ways to treat functions as data.  For the sake of this article, we'll only talk about the parts that are relevant to using WPILibC.
+
+In WPILibC, function types are represented with the ``std::function`` class (https://en.cppreference.com/w/cpp/utility/functional/function).  This standard library class is templated on the function's signature - that means we have to provide it a `function type <https://stackoverflow.com/questions/17446220/c-function-types>`__ as a template parameter to specify the signature of the function (compare this to Java above, where we have a separate interface type for each kind of signature).
+
+This sounds a lot more complicated than it is to use in practice.  Let's look at the call signature of ``cmd::RunOnce`` (which creates an ``InstantCommand`` that, when scheduled, runs the given function once and then terminates):
+
+.. note:: The ``requirements`` parameter is explained in the :ref:`Command-based documentation <docs/software/commandbased/commands:getRequirements>`, and will not be discussed here.
+.. code-block:: cpp
```
Daltz333, line 91: Space between these two

Daltz333 (commented):
I love the article, but we should probably have some external references to alternative resources on learning about functional arguments and higher-order functions.

Oblarg:
@Daltz333 Do you have any recommended resources in particular?

source/docs/software/basic-programming/functions-as-data.rst
```diff
@@ -0,0 +1,121 @@
...
+
+This means we can only capture primitive types (like ``int``, ``double``, and ``boolean``) if they're constants.  If we want to capture a state variable that can change, it *must be wrapped in a mutable object*.
+
+Treating Functions as Data in C++
+---------------------------------
+
+C++ has a number of ways to treat functions as data.  For the sake of this article, we'll only talk about the parts that are relevant to using WPILibC.
+
+In WPILibC, function types are represented with the ``std::function`` class (https://en.cppreference.com/w/cpp/utility/functional/function).  This standard library class is templated on the function's signature - that means we have to provide it a `function type <https://stackoverflow.com/questions/17446220/c-function-types>`__ as a template parameter to specify the signature of the function (compare this to Java above, where we have a separate interface type for each kind of signature).
```
Starlight220, line 86: Link.

source/docs/software/basic-programming/functions-as-data.rst
```diff
@@ -0,0 +1,121 @@
...
+
+.. note:: The ``requirements`` parameter is explained in the :ref:`Command-based documentation <docs/software/commandbased/commands:getRequirements>`, and will not be discussed here.
+
+.. code-block:: cpp
+
+   CommandPtr RunOnce(
+    std::function<void()> action,
+    std::initializer_list<Subsystem*> requirements);
+
+``runOnce`` expects us to give it a ``std::function<void()>`` parameter (named ``action``).  A ``std::function<void()>`` is the C++ term for a ``std::function`` that takes no parameters and returns no value (the template parameter, ``void()``, is a function type with no parameters and no return value).  When we call ``runOnce``, we need to give it a function with no parameters and no return value.  C++ lacks a clean way to refer to existing class methods in a way that can automatically be converted to a ``std::function``, so the typical way to do this is to define a new function inline with a "lambda expression".
```
Starlight220, line 98: I think it's a type rather than a term...

Starlight220 (dismissed):
Consolidate with the similar content in the index of the command-based section, and add links to here

source/docs/software/basic-programming/functions-as-data.rst
Oblarg: There's no link syntax when you're not providing an aliased name, I think?

Oblarg:
@Starlight220 done; i kept most of the summary prose intact but swapped the links to point to our article instead of the oracle/cpp docs

Starlight220:
Why do we need the duplicate summary prose?

Oblarg:
Because it's accurate and there's no reason to remove it since it's not taking up much room at the end of the page

calcmogul (approved):

Daltz333 (approved):

PR #2077, opened 2023-01-03
2023 kickoff downloads

Kevin-OConnor:
Update pages for 2023 kickoff downloads.

Waiting on making 2023 blob public to check radio link before merging


Starlight220:
In the future, on the rare occasion that specific dates are mentioned, please use a date format that won't be confusing outside the US. A team could get see these dates and get worried that tools would be released only in April or something.

PR #2096, opened 2023-01-05
Add platform detecting wpilib installer download button

TheTripleV:
Platform detection for Windows, macOS Intel, macOS arm has been implemented.
If a platform isn't detected, the download link points to the github release page.

Tested:
- [x] macOS Intel
- [x] macOS Arm
- [x] Windows


TheTripleV:
The os detection code uses an api that only works in chrome. I figured some platform detection is better than no platform detection. 

I can look into detection on safari and Firefox after this pr.

TheTripleV:
Yeah on non chrome, the download link just takes you to the same Github release page as before so there's no UX change in that case.

TheTripleV:
So on Safari, it's impossible since Safari always claims to be an Intel Mac so no platform links there.
On firefox, this PR doesn't show a platform link, and Blender's download page actually shows the wrong one. So frc-docs 1, Blender 0.

Daltz333 (approved):

PR #2101, opened 2023-01-06
Add Mac USB camera support to changelog

PeterJohnson:


ThadHouse:
We should add some notes here where there are some cases (Specifically simulating using gradle directly) where the cameras will not work.

PeterJohnson:
Do you want to draft some text to that effect?

Daltz333:
@ThadHouse

Daltz333 (approved):

PR #2103, opened 2023-01-06
CTRE Vendor Library URL Update

dawonn:
The existing URL hits a 404, updated to the latest from https://docs.ctre-phoenix.com/en/stable/ch05a_CppJava.html

### DIFF
--- a/source/docs/software/vscode-overview/3rd-party-libraries.rst
+++ b/source/docs/software/vscode-overview/3rd-party-libraries.rst
@@ -78,7 +78,7 @@ Click these links to visit the vendor site to see whether they offer online inst
    ``https://copperforge.cc/files/dev/vendordeps/LibCu-latest.json``
 
 `CTRE Phoenix Framework <https://store.ctr-electronics.com/software/>`__ - Contains CANcoder, CANifier, Pigeon IMU, Talon FX, Talon SRX, and Victor SPX Libraries and Phoenix Tuner program for configuring CTRE CAN devices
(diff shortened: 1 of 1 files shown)


JCaporuscio:
This technically does fix the 2022 URL and can be merged, but it's going to be changed again for the 2023 release very soon.

dawonn:
@JCaporuscio, I don't suppose you have a preview / beta dev link or eta on 'soon'? :)

dawonn:
I'm guessing you're already aware that the 2022 CTRE lib has linker errors with the WPILIB 2023 release?  (Which is what lead me here to try and make sure I had the latest and greatest release)

Daltz333:
I probably won't merge this in favor of a PR that updates all the vendordep for all vendors for 2023, when each vendor publicly releases their 2023 supported vendordep.

JCaporuscio:
Yep, we're aware 2022 CTR software doesn't work with 2023 :)
As for 'soon', we're doing our final checks right now and will likely have the release public sometime tomorrow.

@Daltz333 I would actually recommend letting each vendor PR their respective links to 2023 software releases (unless you happen to beat them to the punch).  Historically not all vendors actually release at the same time (sometimes not even before kickoff).

Also related, does it make sense to encourage/enforce putting a year next to the vendor's entry?  ie, either list it as "2023 CTRE Phoenix Framework" OR put "2023 Season: " before the online vendordep URL.  That way next year when folks go looking it's more obvious that it hasn't been updated yet.

dawonn:
👍 @JCaporuscio

PR #2111, opened 2023-01-06
Run linkcheck on changed files only on PRs

TheTripleV:


Makefile
```diff
@@ -2,7 +2,7 @@
 #
 # You can set these variables from the command line.
 
-SPHINXOPTS    = -W --keep-going -q -T
+SPHINXOPTS    = -W --keep-going -T
```
Daltz333, line 5: Is this change intended?

TheTripleV: No, I'll undo it later. It's impossible for me to debug in quiet mode so I make that change. I accidentally commited it.

TheTripleV: Ok undid that change

Daltz333 (approved):

PR #2131, opened 2023-01-12
Fix custom data logging example

gaborszita:
DataLogManager should be accessed through the frc namespace.


PeterJohnson (approved):

gaborszita:
Can someone please bypass the check-links check? It's failing the check but my PR clearly didn't cause the problem. My PR only added "frc::" in front of a class, so there is no way this introduced a link check problem.

gaborszita:
I mean there's no need to bypass it if you can just merge it, I just wanted to say that a check is failing but my PR isn't the problem.

PR #2166, opened 2023-01-20
Update the JNI dependencies for ntcore

varun7654:
Added all the platforms that are available here: https://frcmaven.wpi.edu/ui/native/release/edu/wpi/first/ntcore/ntcore-jni/2023.2.1/

The older list of platforms no longer works because Gradle can't find some of the dependencies.


PeterJohnson:
The -jni artifacts are broken, so a more fundamental fix is needed here.

Daltz333:
This has been superceded by #2173

PR #2173, opened 2023-01-25
Update NT Client to use StandaloneAppSamples

sciencewhiz:


Starlight220:
If we have the code in a repository, perhaps RLI it?

Starlight220:
Also, add an Inspector config for them (versioning with a commit hash should be fine).

Starlight220 (approved):

sciencewhiz:
Supercedes #2166

PR #2175, opened 2023-01-27
Correct profiledpid example code

AnotherUser12:
The acceleration parameter provided to the feedforward object was being calculated with the incorrect velocity setpoint (now corrected). Teams might reference this example when integrating profiledpid and feedforward for position control of an arm mechanism this season.


calcmogul:
Specifically, the current acceleration calculation uses the ProfiledPIDController's velocity setpoint and lastSpeed, but those two values are identical before the call to calculate(). Moving the acceleration calculation to after calculate() makes it use the next and current velocity (stored in lastSpeed) like it should.

calcmogul (approved):

sciencewhiz (approved):

PR #2179, opened 2023-02-01
Fix Scheduler Event RLIs

Starlight220:


Starlight220:
This should fix the broken links check that's failing on other PRs.

sciencewhiz:
Since allwpilib has already been tagged for 2023.3.1, you can use that rather then a hash

sciencewhiz (approved):

PR #2181, opened 2023-02-02
Pin black version

sciencewhiz:


.github/workflows/CI.yml
```diff
@@ -180,6 +180,7 @@ jobs:
     - uses: actions/setup-python@v4
       with:
         python-version: 3.9
-    - uses: psf/black@22.1.0
+    - uses: psf/black@stable
       with:
         options: "--check"
+        version: "22.1.0"
```
auscompgeek, line 186: Should we pin to the latest 22.x release, i.e. 22.12, given the format style is stable for the entire year?

sciencewhiz: I was planning to update to 23 and reformat, after proving that this works, so I figured it was less likely to introduce problems to run the version that had previously intended to be run

Daltz333 (approved):

PR #2183, opened 2023-02-04
Schedule() Method typo

juliashimshock:
There is a typo in which under The schedule() Method there is only an open parentheses and no close parentheses when linking Java and C++. This was extremely distracting and I had trouble focusing on reading when seeing this. Thank you for your time.


sciencewhiz:
As per https://docs.wpilib.org/en/stable/docs/contributing/frc-docs/contribution-guidelines.html#creating-a-pr PRs should be targeted to main not stable. I updated the base branch.

sciencewhiz (approved):

PR #2185, opened 2023-02-05
Use poetry pinned black over GH action

Daltz333:


pyproject.toml
```diff
@@ -5,6 +5,7 @@ description = ""
 authors = []
 
 [tool.poetry.dependencies]
+black = "==23.1.0"
```
auscompgeek, line 8: Any reason to not use `^23.1.0`, given [Black's stability policy](https://black.readthedocs.io/en/stable/the_black_code_style/index.html#stability-policy)?

sciencewhiz (approved):

PR #2237, opened 2023-04-02
Add page for wiring Rev Pneumatic Hub

### LINKED ISSUE #2155: Wiring Pneumatics only covers PCM
https://docs.wpilib.org/en/stable/docs/hardware/hardware-basics/wiring-pneumatics.html

sciencewhiz:
Update PCM for wiring with PDH
Fixes #2155


Daltz333 (commented):
Is the image from REV docs? Can we get the original vector for it?

sciencewhiz:
It's a bastardized version of the vector we already have, with a few manual edits after conversion to png, since that's what I'm more comfortable editing.

Daltz333:
It's unfortunate, as I'd rather have the original modified vector. I won't have that blocking this PR though, so I should be fine once the conflicts be resolved.

Daltz333 (approved):

PR #2246, opened 2023-04-26
Add correct header to non-static command factories

RobinsonZ:
The "Static Command Factories" subheader mistakenly has an example of non-static command factories under it as well. This adds the proper header and makes the example a bit clearer.

### DIFF
--- a/source/docs/software/commandbased/organizing-command-based.rst
+++ b/source/docs/software/commandbased/organizing-command-based.rst
@@ -261,7 +261,7 @@ Then, elsewhere in our code, we can instantiate an single instance of this class
 
     Command driveAndIntake = autoRoutines.driveAndIntake();
     Command driveThenIntake = autoRoutines.driveThenIntake();
-    
+
     Command drivingAndIntakingSequence = Commands.sequence(
       autoRoutines.driveAndIntake(),
       autoRoutines.driveThenIntake()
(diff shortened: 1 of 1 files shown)


Oblarg (approved):

RobinsonZ:
@Starlight220 This code and the design pattern it references was already in the docs before this PR, just badly mislabeled (under the "Static Command Factories" header). It's not supposed to be showing the superstructure pattern. Note that `Drivetrain` and `Intake` are two unrelated subsystems which can be operated independently; they have only been captured in the `AutoRoutines` class to save a bunch of method arguments.

I'm also a bit unsure why you're talking about in-subsystem instance factories — this PR doesn't change any of the docs around command factories in subsystems. The `AutoRoutines` class is more like a builder pattern, and has no internal state whatsoever aside from holding two references to the requisite subsystems. It doesn't act as a mutex or relate to any physical hardware like a `Subsystem` would — multiple `AutoRoutines` instances could be created in different locations with no issues. The code example I added is just intended to show that unlike static factories (which take a list of subsystems on every invocation), `AutoRoutines` needs to be instantiated with the subsystem list and then can be used to generate multiple different commands.

I agree that general revision of the command-based documentation to be more cohesive and idiomatic (and include examples of the superstructure pattern) is needed, and I'll hopefully be able to do some work on that after my final exams are done, but this PR is just to fix an obvious error in the heading structure.

I'm also unclear on what you mean by "unidiomatic" re: `RobotContainer`, but looking at this code again I'm realizing that I can probably get my point across without bothering to set up a whole `RobotContainer` class. I'll make some edits to that effect.

Oblarg (approved):

PR #2279, opened 2023-07-08
Add instructions for disabling VS Code line length linting

sciencewhiz:


auscompgeek (commented):
Can we add a doc8 config to the repo instead of forcing contributors to work around in their editor settings?

sciencewhiz:
the VS Code extension doesn't appear to work with a doc8 config file in the root directory. Looks like there'd need to a config in every directory, which doesn't sound reasonable. https://github.com/vscode-restructuredtext/vscode-restructuredtext/issues/128#issuecomment-446605285

Daltz333 (approved):

PR #2337, opened 2023-10-09
Documentation for XRP

zhiquanyeo:


source/index.rst
```diff
@@ -394,6 +394,13 @@ Community translations can be found in a variety of languages in the bottom-left
 
    docs/romi-robot/index
 
+.. toctree::
+   :maxdepth: 1
+   :caption: XRP Robot
+   :hidden:
+
+   docs/xrp-robot/index
+
```
Daltz333, line 403: Very well done PR. My only issue is the index here. If the romi is no longer supported, we should hide it from the TOC and point to these docs. If we plan to support both, can they both be in this section and the section title "additional platform support" or another alternative name.

zhiquanyeo: Yeah that makes sense. Maybe something like "Romi and XRP support"?

zhiquanyeo: Also updated the "squares"

PeterJohnson (approved):

PR #2348, opened 2023-10-16
Update vendor and battery information for XRP

zhiquanyeo:


PeterJohnson:
Looks like you probably have to add DigiKey to the linkcheck exclusions list? https://github.com/wpilibsuite/frc-docs/blob/main/source/conf.py#L143

zhiquanyeo:
yeah looks like adding to the exclusion list "fixed" it

PeterJohnson (approved):

PR #2360, opened 2023-10-23
Update to sphinx 7

GrahamSH-LLK:
Currently first pass


sciencewhiz:
Unfortunately, It's fairly common for things to break on big upgrades like this in the past.

Here's what I've noticed:

- colors on the front page panels shifted slightly (not necessarily a problem, but a possible indication of other CSS issues)s
- 404 page is all messed up
- sphinx-rtd-theme v2 (which looks to add support for Sphinx 7) should be released soon. Not sure what might be broken with the older version.

I'm not sure if there's a way to see if translations or pdf/html still works without merging @Daltz333

Daltz333:
You can build translations locally to test, but I'd just rely on CI to test PDF builds.

sciencewhiz:
Looks like translations failed

sciencewhiz:
The colors on the front panels seem back to normal

Daltz333:
Needs a rebase

GrahamSH-LLK:
> Needs a rebase

Rebased and builds succeeding.

sciencewhiz (approved):

GrahamSH-LLK:
@sciencewhiz Strange that it failed. I can repro the problem locally by running the command that rtd runs for whatever part of the build it is that failed. However, I can't figure out where the problem is as the stack trace is fairly unhelpful.

sciencewhiz:
@TheTripleV identified the problem as being from https://github.com/wpilibsuite/frc-docs/blob/main/source/_extensions/rtd_patch.py

TheTripleV:
Adding `*args, **kwargs` to both inits (definition and call) should make it work on current sphinx and sphinx 7.

GrahamSH-LLK:
Fixed PR in #2395

PR #2370, opened 2023-10-28
Set default image width/alignment in LaTeX build

abidingabi:
This produces much better looking presentation for the PDF build of gm0, and perhaps a similar idea can be applicable here. However, it will likely require some tweaking to better match the different style of content.


abidingabi:
Before:
image

After:
image

Daltz333:
Pending #2360

Daltz333 (approved):

PR #2390, opened 2023-11-06
Add no-op to the glossary

### LINKED ISSUE #2388: Add "no-op" to glossary
It's used in the docs and users may not know what it means.

kshitijarvikar:
## Added Hyperlinks to "no-op", for better understanding of no-op behavior to readers.
resolved #2388


rzblue:
Instead of linking to an external document it should be added to the glossary (https://github.com/wpilibsuite/frc-docs/blob/main/source/docs/software/frc-glossary.rst) and linked inline like so: `:term:\`no-op\``

kshitijarvikar:
look for updated PR, or are there any changes suggested.!

source/docs/software/frc-glossary.rst
```diff
@@ -178,3 +178,6 @@ FRC Glossary
 
    transitory
       In :term:`NetworkTables`, a :term:`topic` that will disappear after the last :term:`publisher` stops publishing.
+
+   no-op
+      No op means no operation, which is a computer instruction that specifies no operation but takes up a small amount of storage space. When the computer processor encounters a no op instruction, it simply moves to the next sequential instruction. Read more about no-op on `Wikipedia <https://en.wikipedia.org/wiki/NOP_(code)>`_ .
```
sciencewhiz, line 183: I think the part about taking up a small amount of storage space is irrelevant and confusing. I'd remove it

sciencewhiz (commented):
As rzblue said, existing uses need to use the term syntax. I see at least two uses, one in the command scheduler article and one in the new for 2024 article

PR #2395, opened 2023-11-12
Upgrade to Sphinx 7

GrahamSH-LLK:
Same changes as previous PR, but I fixed RTD builds. Tested locally.


TheTripleV:
Why is rtd theme using a release candidate?

GrahamSH-LLK:
> Why is rtd theme using a release candidate?

The older version didn't play nicely with the 404 page. I haven't found any issues with the theme so far.

sciencewhiz:
We had to use an rc for the rtd theme when updating to Sphinx 5 also. The release timeline is really slow.

sciencewhiz (approved):

PR #2397, opened 2023-11-14
Add detail to Java Units documentation

agasser:
This adds more detail to the Java units library. Related to https://github.com/wpilibsuite/2024Beta/issues/28


source/docs/software/basic-programming/java-units.rst
```diff
@@ -68,13 +117,28 @@ Unit conversions can be done by calling ``Measure#in(Unit)``. The Java type syst
...
+The ``Measure`` class has both ``toShortString()`` and ``toLongString()`` methods that can be used to get a human-readable representation of the measure. This feature is useful to display a measurement on a dashboard or in logs.
+
+Mutibility and Object Creation
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Memory Usage and the Garbage Collector
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+To reduce the number of object instances you create, and reduce memory usage, a special ``MutableMeasure`` class is available. You may want to consider using mutable objects if you are using the units library repeatedly, such as in the robot's periodic loop.
 
-The roboRIO is a severely memory-constrained runtime environment and the Java garbage collector has to run to keep memory usage to a reasonable level. The garbage collector will have to pause the robot program in order to free unused objects, which manifests as loop overruns and potentially jittery controls. To avoid this issue, allocate as few ``Measure`` objects as possible in areas of code that will run periodically, such as in a Command's ``execute`` or one of TimedRobot's periodic methods.
```
GrahamSH-LLK, line 76: I think this was an important section to keep.

agasser: Thanks for reviewing, but I disagree. There aren't similar warnings for other classes in WPILib, such as `Translation2d`. At best, this seems like a general comment that belongs someplace else.
See my discussion in topic https://github.com/wpilibsuite/2024Beta/discussions/29

calcmogul: Then it should be moved to a more general location instead of deleted outright in this PR.

agasser: Again, I disagree because it never existed before and just slipped in with the java-units documentation. However, I'll see if I can find a good spot for the warning.

max-niederman: Moving this to a general location would be good, but even so I think it would be good to leave in a sentence motivating the use of the mutable methods. Otherwise it could be confusing: "Why is there the option to mutate in the first place? Should I be mutating?"

agasser: I'm working on adding a page, and I will link it from this documentation. But, I did include such a statement at the beginning of the section: "To reduce the number of object instances you create, and reduce memory usage..."

source/docs/software/basic-programming/java-units.rst
```diff
@@ -1,58 +1,107 @@
...
++--------------------------------------------------------+------------------------------------------------------------------------------------------------------------------------+
+| ``Measure<Per<Voltage, Velocity<Velocity<Angle>>>``    | ``VoltsPerRadianPerSecondSquared``                                                                                     |
++--------------------------------------------------------+------------------------------------------------------------------------------------------------------------------------+
+
+Creating Measures
+~~~~~~~~~~~~~~~~~
+
+The ``Measure`` class is a generic type that represents a physical quantity with its corresponding unit of measurement. It provides a consistent and type-safe way to handle different types of measurements, such as distance, angle, velocity, and more. To create a ``Measure`` object, you call the ``Unit.of`` method on the appropriate unit object. For example, to create a ``Measure<Distance>`` object representing a distance of 6 inches, you would write:
```
max-niederman, line 63: The way you've phrased this I think it would be easy to mistakenly assume that `Measure` is parameterized by the unit. I think you should explicitly say that `Measure` is parameterized by the dimension rather than a particular unit, since that's half the point of the library.

Maybe something like:
> The measure class is a generic type representing a physical quantity. It tracks the type of measurement (distance, angle, mass, etc.), but abstracts away the particular unit of measure (e.g. meter vs. inch) unless you explicitly convert to a particular unit.

source/docs/software/basic-programming/java-units.rst
```diff
@@ -1,58 +1,107 @@
...
+
+.. code-block:: java
+
+  Measure<Distance> distance1 = Inches.of(10);
+  Measure<Distance> distance2 = Meters.of(0.254);
 
-Because the ``Measure`` objects are Java objects, arithmetic must be done by calling methods on the objects. Note that each operation will return a new ``Measure`` object.
+  Measure<Distance> totalDistance = distance1.plus(distance2);
+
+This code will automatically convert the units of measurement to the same unit before adding the two distances. The resulting ``totalDistance`` object will be a new ``Measure<Distance>`` object that has a value of 0.508 meters, or 20 inches.
```
max-niederman, line 91: This is a useful explanation, but I think it should be mentioned that the unit conversion is entirely internal and the whole point of the library is to abstract this away. Maybe you could preface it with "Internally, this works by" or something?

source/docs/software/basic-programming/java-units.rst
```diff
@@ -1,58 +1,107 @@
...
+
+A more advanced example combines the wheel diameter and gear ratio to calcualate the distance per rotation of the wheel:
```
max-niederman, line 93: The reader already knows the next example will be more advanced; no need to say it.

source/docs/software/basic-programming/java-units.rst
```diff
@@ -68,13 +117,28 @@ Unit conversions can be done by calling ``Measure#in(Unit)``. The Java type syst
      maxAcceleration.in(MetersPerSecondPerSecond)
    );
 
-.. note:: Due to restrictions of the Java type system, acceleration is modeled as ``Velocity<Velocity<X>>>`` instead of having its own class.
```
max-niederman, line 71: This note isn't sufficient either, but I think it's important to explain or at least mention `Velocity`.

source/docs/software/basic-programming/java-units.rst
```diff
@@ -68,13 +117,28 @@ Unit conversions can be done by calling ``Measure#in(Unit)``. The Java type syst
...
+.. code-block:: java
+
+  Measure<Distance> armLength = Feet.of(3).plus(Inches.of(4.25));
+  Measure<Distance> endEffectorX = armLength.times(Math.cos(getArmAngle().in(Radians)));
+  Measure<Distance> endEffectorY = armLength.times(Math.sin(getArmAngle().in(Radians)));
+
+Human-readable Formatting
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``Measure`` class has both ``toShortString()`` and ``toLongString()`` methods that can be used to get a human-readable representation of the measure. This feature is useful to display a measurement on a dashboard or in logs.
```
max-niederman, line 134: It might be helpful to add an example which shows the difference between `toShortString` and `toLongString`.

source/docs/software/basic-programming/java-units.rst
```diff
@@ -68,13 +117,28 @@ Unit conversions can be done by calling ``Measure#in(Unit)``. The Java type syst
...
+
+Mutibility and Object Creation
```
max-niederman, line 136: typo: "mutability" not "mutibility"

max-niederman (changes requested):

source/docs/software/basic-programming/java-units.rst
agasser: I like that, I'll add the bit about abstracting the unit of measure. It's under the heading "Creating objects" so I'm going to leave out "unless you explicitly convert to a particular unit"

source/docs/software/basic-programming/java-units.rst
agasser: This is not my note, it was already there.
I agree, I had to read it three times before I understood `X` was actually `Velocity` in the example. If such a note is valuable, it probably belongs under "Using the Units Library". because it applies to all of the crazy looking types like `Measure<Per<Voltage, Velocity<Velocity<Distance>>>>`

source/docs/software/basic-programming/java-units.rst
agasser: Is there rationale for not overriding `toString()` that should be added to this section? That seems like an oversight, but maybe there was a reason.

source/docs/software/basic-programming/java-units.rst
max-niederman: Yeah, that would probably be a good place to put it. Possibly with an explanation that it's shorthand for `Per<Time, _>`.

agasser:
Ready for final review

source/docs/software/basic-programming/java-gc.rst
```diff
@@ -0,0 +1,35 @@
...
+The performance impact of creating a lot of small objects is not limited to GC overhead. Creating and destroying objects also involves CPU cycles, which can further impact application performance. Additionally, accessing and manipulating large numbers of small objects can lead to cache misses, which can further slow down the application.
+
+Design Considerations
+~~~~~~~~~~~~~~~~~~~~~
+
+If you anticipate your application creating a large number of small objects, it is important to consider design strategies to mitigate the potential memory and performance issues. Here are some strategies to consider:
+
+- Minimize object creation: Carefully evaluate the need for each object creation. If possible, reuse existing objects or use alternative data structures, such as arrays or primitives, to avoid creating new objects.
+
+- Object pooling: Implement object pooling to reuse frequently created objects. This involves creating a pool of objects that are available for use when needed. When an object is no longer needed, it is returned to the pool instead of being destroyed.
```
sciencewhiz, line 31: While certainly a general design pattern, I'm struggling to think of a use case for an FRC robot. I'm thinking this is probably not something who'd reading about the GC for the first time would be prepared to implement, so I'd lean towards not covering in this artcile

max-niederman (approved):
LGTM

source/docs/software/basic-programming/java-units.rst
```diff
@@ -1,58 +1,122 @@
...
 -----------------------
 
 The Java units library is available in the ``edu.wpi.first.units`` package. The most relevant classes are `edu.wpi.first.units.Units <https://github.wpilib.org/allwpilib/docs/beta/java/edu/wpi/first/units/Units.html>`__, which contains a set of predefined units; and `edu.wpi.first.units.Measure <https://github.wpilib.org/allwpilib/docs/beta/java/edu/wpi/first/units/Measure.html>`__, which is used to tag a value with a unit. It is recommended to static import ``edu.wpi.first.units.Units.*`` to get full access to all the predefined units.
 
-The library comes with predefined SI and imperial units for distance, angle, time, velocity and acceleration (both linear and angular), mass, voltage, current, power, energy, and temperature.
+The library comes with many predefined units:
+
++--------------------------------------------------------+------------------------------------------------------------------------------------------------------------------------+
+| Type                                                   | Units                                                                                                                  |
```
Daltz333, line 23: I'd rather not us maintain an independent list of units.

source/docs/software/basic-programming/java-gc.rst
```diff
@@ -0,0 +1,33 @@
+Java Garbage Collection
+=======================
```
Daltz333, line 2: I'm not a fan of placing this under basics, even with it being the catchall that it is. I wonder if it makes more sense to create a subsection called language concepts.

source/docs/software/basic-programming/java-gc.rst
```diff
@@ -0,0 +1,33 @@
...
+
+Java garbage collection is an automatic process, which means that the programmer does not need to explicitly deallocate memory. The garbage collector keeps track of which objects are in use and which are not, and it periodically reclaims unused objects.
+
+Object Creation
+---------------
+
+Creating a large number of small objects in Java can lead to memory and performance issues. While the Java Garbage Collector (GC) is designed to handle memory management efficiently, creating too many small objects can overwhelm the GC and cause performance degradation.
+
+Memory Concerns
+~~~~~~~~~~~~~~~
```
Daltz333, line 13: Per style guide, should be = - ^ ~

Daltz333 (changes requested):

sciencewhiz:
@SamCarlberg

source/docs/software/basic-programming/java-gc.rst
agasser: It snuck into the basics article for Java Units, so it is here now. I agree it didn't belong here. The other reviewers feel it's critical information when dealing with the units library, for some reason, even though WPILib has several other core immutable types that are designed to be recreated on every period, such as `Pose2d` and `ModuleStates`.

Maybe it should be removed from this PR (and the Java Units library documentation), and I can write up a separate issue for it.

calcmogul: > The other reviewers feel it's critical information when dealing with the units library, for some reason, even though WPILib has several other core immutable types that are designed to be recreated on every period, such as Pose2d and ModuleStates.

For typical math, unit containers get instantiated _way_ more than any of our other value types.

agasser: > For typical robot code math, unit container value types would get instantiated way more than any of our other value types

It's beyond the scope of this document PR, but this comment makes me question if there should be arithmetic functions that return Measures on the Units library at all, if object instantiation is such an issue. Remember, our target audience is 13-18 year old students with little to no Java experience. Troubleshooting the problem starting on line 245 in this doc will be more difficult for a student or CSA than debugging a classic unit error. This feature just swaps one problem that takes careful consideration (units) for another (object ownership and mutability.)

It might be best to only offer a `replace()` mutation function, and give guidance on appropriate use of the Units library. Another option would be to only offer arithmetic functions where the called must pass a unit and it returns double, to avoid object instantiation, like `double add(double, Unit)` and `double add(Measure<Unit>, Unit)`.

To bring this back on topic, should I leave the java-gc page in the Basics section, or remove it and create a new PR?

source/docs/software/basic-programming/java-units.rst
agasser: This is a valid concern, but in the interest of making this document a one-stop-shop, I think it adds a lot of value. If the maintainers feel we should remove it, I will change the wording to encourage users to follow the Javadoc link to the Units class. Honestly, I missed it when I tried to use the library the first time and I really struggled to figure out how to use velocity measures. It's really the reason I created this PR in the first place.
What would the maintainers like?

source/docs/software/basic-programming/java-units.rst
```diff
@@ -1,58 +1,135 @@
...
++--------------------------------------------------------+------------------------------------------------------------------------------------------------------------------------+
+
+Java Generics
+^^^^^^^^^^^^^
+Units of measurement can be complex expressions involving various dimension, such as distance, time, and velocity. Nested `generic type parameters <https://docs.oracle.com/javase/tutorial/java/generics/index.html>`__ allow for the definition of units that can represent such complex expressions. Generics are used to keep the library concise, reusable, and extensible, but it tends to be verbose due to the syntax for Java generics.
+
+For instance, consider the type ``Measure<Velocity<Distance>>``. This type represents a measurement for velocity, where the velocity itself is expressed as a unit of distance per unit of time. This nested structure allows for the representation of units like meters per second or feet per minute. Similarly, the type ``Measure<Per<Voltage, Velocity<Distance>>>`` represents a measurement for a ratio of voltage to velocity. This type is useful for representing quantities like volts per meter per second.
```
SamCarlberg, line 77: Should mention what V/m/s would be useful for so readers know it's not contrived just for sake of an example. This would also be a good place to make the distinction between `Velocity` and `Per<, Time>`

source/docs/software/basic-programming/java-units.rst
```diff
@@ -1,58 +1,135 @@
...
-       public static final Measure<Distance> kTrackWidth = Inches.of(26);
-     }
-   }
+.. code-block:: java
+
+  Measure<Mass> kArmMass = Kilograms.of(1.423);
+  Measure<Distance> kArmLength = Inches.of(32.25);
+  Measure<Angle> kMinArmAngle = Degrees.of(5);
+  Measure<Angle> kArmMaxTravel = Rotations.of(0.45);
+  Measure<Velocity<Distance> kMaxSpeed = MetersPerSecond.of(2.5);
```
SamCarlberg, line 105: :+1: good examples

source/docs/software/basic-programming/java-units.rst
```diff
@@ -1,58 +1,135 @@
...
+
+Performing Calculations
+^^^^^^^^^^^^^^^^^^^^^^^
 
-Because the ``Measure`` objects are Java objects, arithmetic must be done by calling methods on the objects. Note that each operation will return a new ``Measure`` object.
+The ``Measure`` class also supports arithmetic operations, such as addition, subtraction, multiplication, and division. These are done by calling methods on the objects. These operations always ensure that the units are compatible before performing the calculation, and they return a new ``Measure`` object. For example, you can add two ``Measure<Distance>`` objects together, even if they have different units:
```
SamCarlberg, line 110: This section would be a good place to link to the GC article, since the default immutable operations could be a source of unexpected object allocations in hot loops. Or link to the mutability & object creation section below

source/docs/software/basic-programming/java-units.rst
SamCarlberg: `Velocity` isn't shorthand. It's incompatible with `Per` because of limitations in the Java type system. The `Unit` class takes the self type as a generic argument, meaning `Velocity<D extends Unit<D>> extends Per<D, Time>` would be out of bounds.

Kotlin teams can use type aliasing and may be able to completely ignore the `Velocity` class, but that's out of scope for our documentation

source/docs/software/basic-programming/java-units.rst
```diff
@@ -68,13 +145,31 @@ Unit conversions can be done by calling ``Measure#in(Unit)``. The Java type syst
...
+.. code-block:: java
+
+  Measure<Distance> armLength = Feet.of(3).plus(Inches.of(4.25));
+  Measure<Distance> endEffectorX = armLength.times(Math.cos(getArmAngle().in(Radians)));
+  Measure<Distance> endEffectorY = armLength.times(Math.sin(getArmAngle().in(Radians)));
+
+Human-readable Formatting
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The ``Measure`` class has both ``toLongString()`` and ``toShortString()`` methods that can be used to get a human-readable representation of the measure. This feature is useful to display a measure on a dashboard or in logs.
```
SamCarlberg, line 162: Both implementations of the `Measure` interface implement `toString` as `toShortString`, eg `Inches.of(6).toString()` will return `"6 in"`

source/docs/software/basic-programming/java-units.rst
```diff
@@ -230,12 +325,12 @@ Defining New Units
 
 There are four ways to define a new unit that isn't already present in the library:
 
-- Using the ``Unit#per`` or ``Unit#mult`` methods to create a composite of two other units;
+- Using the ``Unit.per`` or ``Unit.mult`` methods to create a composite of two other units;
 - Using the ``Milli``, ``Micro``, and ``Kilo`` helper methods;
 - Using the ``derive`` method and customizing how the new unit relates to the base unit;
-- Subclassing ``Unit`` to define a new type of unit
+- Subclassing ``Unit`` to define a new diimension
```
SamCarlberg, line 331: typo

```suggestion
- Subclassing ``Unit`` to define a new dimension
```

SamCarlberg (changes requested):
Thanks for the updates! Had a few comments

source/docs/software/basic-programming/java-gc.rst
```diff
@@ -0,0 +1,33 @@
+Java Garbage Collection
+=======================
+Java garbage collection is the process of automatically managing memory for Java objects. The Java Virtual Machine (JVM) is responsible for creating and destroying objects, and the garbage collector is responsible for identifying and reclaiming unused objects.
+
+Java garbage collection is an automatic process, which means that the programmer does not need to explicitly deallocate memory. The garbage collector keeps track of which objects are in use and which are not, and it periodically reclaims unused objects.
+
+Object Creation
+---------------
+
+Creating a large number of small objects in Java can lead to memory and performance issues. While the Java Garbage Collector (GC) is designed to handle memory management efficiently, creating too many small objects can overwhelm the GC and cause performance degradation.
```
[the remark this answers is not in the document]
agasser: I have no numbers, but I'll try to add some behavior examples. Starlight mentioned he once did some benchmarking, but I think this is largely theoretical.

source/docs/software/basic-programming/java-gc.rst
```diff
@@ -0,0 +1,33 @@
...
+Design Considerations
+^^^^^^^^^^^^^^^^^^^^^
+
+If you anticipate your application creating a large number of small objects, it is important to consider design strategies to mitigate the potential memory and performance issues. Here are some strategies to consider:
+
+- Minimize object creation: Carefully evaluate the need for each object creation. If possible, reuse existing objects or use alternative data structures, such as arrays or primitives, to avoid creating new objects.
+
+- Efficient data structures: Use data structures that are well-suited for the type of data you are working with. For example, if you are dealing with a large number of primitive values, consider using arrays or collections specifically designed for primitives.
+
+In general, it is advisable to avoid creating a large number of small objects in Java unless it is necessary for the application's functionality. By carefully considering design strategies and optimizing object creation, you can minimize the memory and performance impact of creating small objects.
```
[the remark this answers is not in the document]
agasser: > What design strategies are recommended?
"it is advisable to avoid creating a large number of small objects" and "Use data structures that are well-suited for the type of data you are working with"


> Can you provide code examples of how object creation can be optimized?

"Minimize object creation: Carefully evaluate the need for each object creation. If possible, reuse existing objects or use alternative data structures, such as arrays or primitives, to avoid creating new objects."

I think you're making the point this is repetitive. It's a summary, but I'll remove it.

source/docs/software/basic-programming/java-units.rst
```diff
@@ -1,58 +1,135 @@
...
+
+For instance, consider the type ``Measure<Velocity<Distance>>``. This type represents a measurement for velocity, where the velocity itself is expressed as a unit of distance per unit of time. This nested structure allows for the representation of units like meters per second or feet per minute. Similarly, the type ``Measure<Per<Voltage, Velocity<Distance>>>`` represents a measurement for a ratio of voltage to velocity. This type is useful for representing quantities like volts per meter per second.
+
+It's important to note that not all measurements require such complex nested types. For example, the type ``Measure<Distance>`` is sufficient for representing simple units like meters or feet. However, for more complex units, the use of nested generic type parameters is essential.
+
+For local variables, you may choose to use Java's `var` keyword instead of including the full type name. For example, these are equivalent:
```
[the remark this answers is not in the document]
agasser: I prepared this text to add, but it really seems out of place.
> The ``var`` keyword is a type inference feature that allows you to declare local variables (variables declared within a method, constructor, or block) without explicitly specifying their type. The compiler will infer the type of the variable based on the initializer expression. In this case, the initializer expression is ``VoltsPerMeterPerSecond.of(8)``, so the compiler will infer that the type of the variable ``v`` is ``Measure<Per<Voltage, Velocity<Distance>>>``.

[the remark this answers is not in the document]
agasser: Also, `var` is used in several other examples in frc-docs without explanation.

SamCarlberg (approved):
LGTM

source/docs/software/basic-programming/java-gc.rst
sciencewhiz: Maybe we should rename basic programming to miscellaneous. It's really a grab bag of topics that aren't really basic programming

agasser: So I should I remove the article and create a new issue for this debate?

agasser:
Are you waiting for anything from me on this PR?

agasser:
@Daltz333 Is this ready to merge? It's showing changes still requested from you, but I think I addressed everything.

PR #2401, opened 2023-11-17
Fix version warning

GrahamSH-LLK:
The version warning somehow got on an old version in all the git merging. This fixes it.


sciencewhiz:
1.1.4 was released 4 hours ago, so don't think it was missed

sciencewhiz (approved):

PR #2407, opened 2023-11-29
Add RobotPy related image processing documentation

virtuald:
Migrating from https://robotpy.readthedocs.io/en/stable/vision/code.html#


virtuald:
I'm not exactly sure where to put https://robotpy.readthedocs.io/en/stable/vision/other.html. Maybe it belongs as documentation for the cscore API instead?

virtuald:
The examples are old, but presumably if we get a better CPU in the future it'll become relevant again.

source/docs/software/vision-processing/roborio/using-multiple-cameras.rst
```diff
@@ -113,12 +160,20 @@ If you're using some other dashboard, you can change the camera used by the came
          prevTrigger = joy1.GetTrigger();
        }
 
+    .. code-block:: python
+
+       # Setting the source directly via joystick isn't possible in Python, you
```
sciencewhiz, line 165: The previous example is only really applicable with SmartDashboard. Can NT be used to communicate between the two processes?

virtuald: The previous example uses `NetworkTableInstance.getDefault().getTable("").getEntry("CameraSelection")`... which is NetworkTables? So yes.

sciencewhiz: I guess my point is that it isn't so obvious what the code would look like. There needs to be NT code in the joystick section that isn't needed in the previous example.

virtuald: I don't understand. The code is exactly the example above. See line 78-90 of this file. It's using a joystick to set a NetworkTables value, which controls the camera.

sciencewhiz (approved):

PR #2408, opened 2023-11-30
Driverstation Article for Getting Alliance Color

jasondaming:
Should I include more of a description about other things the `DriverStation` class does (like Joysticks and game state etc)?

I thought this class also did the Game data but I don't see that anymore did we remove that for some reason?

Any other examples from this class?


jasondaming:
okay still there we just removed the article about game data using it.

So I guess the question is do we even mention it or just accept that when the game calls for game data there will be a separate article (or maybe an addition to this one?) that will cover it?

calcmogul:
I'd prefer there always being a page (or pages) for it.

sciencewhiz:
There is a separate article for game data. It's just hidden

sciencewhiz:
I'm thinking it would be more discoverable if the article title was about the alliance color, rather then driverstation class, which can also be confused with the driverstation software. Since there's a ton of stuff in the driverstation class, if more things are added in the future it's probably better split across several articles.

Alternately, the link in the changelog should point to the specific anchor with the article with a ref link rather then to the full article. But I prefer making this article specific to alliance color.

sciencewhiz (approved):

PR #2411, opened 2023-12-01
Added RLIs for python to preferences page

jasondaming:


jasondaming:
Will probably want to update when this issue gets resolved
https://github.com/robotpy/examples/issues/79

sciencewhiz (changes requested):
The RLIs need to be to a tag, not main, to avoid unintended breakage in the future if the example changes.

sciencewhiz (approved):

PR #2412, opened 2023-12-01
Update the Control System Docs to include Python

jasondaming:
Removed sentence about object oriented.

I believe python is currently unsupported in Robot Builder right?


sciencewhiz:
Correct, Python isn't supported by RobotBuilder and there isn't plans to add support, but PRs are welcome.

source/docs/controls-overviews/control-system-software.rst
```diff
@@ -26,7 +26,7 @@ Visual Studio Code
 .. image:: images/control-system-software/visual-studio-code.png
    :alt: Welcome screen of Visual Studio Code.
 
-Visual Studio Code is the supported development environment for C++ and Java (the other two supported languages). Both are object-oriented text based programming languages. A guide to getting started with C++ or Java for FRC, including the installation and configuration of Visual Studio Code can be found :ref:`here <docs/zero-to-robot/step-2/wpilib-setup:WPILib Installation Guide>`.
+Visual Studio Code is the supported development environment for C++, Java, and Python (the other supported languages).  A guide to getting started with C++, Java, and Python for FRC, including the installation and configuration of Visual Studio Code can be found :ref:`here <docs/zero-to-robot/step-2/wpilib-setup:WPILib Installation Guide>`.
```
sciencewhiz, line 29: Given that it's December and the extension doesn't yet support python, I'd be tempted to not include thse changes, and if @ThadHouse pulls a miracle, then we can add it.

jasondaming: I agree it would be great if the extension works, but I don't think this promises that it will.  It links to step 2 which will be updated to include whatever pythons install process is.  VSCode can still be a recommended dev environment.

My suggestion would be to leave it as is and if / when the extension doesn't get done we could add more text explicitly calling out that the extension doesn't support python.

sciencewhiz: It's more then just the extension, no where in the docs do we talk about python in vs code. The ZTR first program is agnostic to IDE. If VS Code really is preferred, then basic documentation should exist for it, and that PR could then say it's supported

sciencewhiz (approved):

PR #2421, opened 2023-12-07
Fix which port the ADIS16470 is plugged into by default

Technologyman00:

image


sciencewhiz:
Per https://docs.wpilib.org/en/stable/docs/contributing/frc-docs/contribution-guidelines.html#creating-a-pr PRs should be targeted to main. Please rebase on main.

Technologyman00:
Ok I made the change. Apologies

sciencewhiz:
the commit needs to be rebased on main, not just the base branch changed, since it now picks up unrelated changes

PR #2425, opened 2023-12-10
ADUS16470 default port says MXP it should say SPI

Technologyman00:


sciencewhiz:
Unfortunately, this is still based on stable

PR #2430, opened 2023-12-11
Added python for kinematics and odometry

jasondaming:


source/docs/software/kinematics-and-odometry/differential-drive-odometry.rst
```diff
@@ -9,7 +9,7 @@ Creating the Odometry Object
 The ``DifferentialDriveOdometry`` class constructor requires three mandatory arguments and one optional argument. The mandatory arguments are:
 
 * The angle reported by your gyroscope (as a ``Rotation2d``)
-* The initial left and right encoder readings. In Java, these are each a ``double``, and must represent the distance traveled by each side in meters.  In C++, the :doc:`units library </docs/software/basic-programming/cpp-units>` must be used to represent your wheel positions.
+* The initial left and right encoder readings. In Java / Python, these are each a ``double``, and must represent the distance traveled by each side in meters.  In C++, the :doc:`units library </docs/software/basic-programming/cpp-units>` must be used to represent your wheel positions.
```
virtuald, line 12: Python doesn't have doubles, it only has a float type which is usually implemented as a double in C.

virtuald (changes requested):
Only one objection, everything else looks good.

source/docs/software/kinematics-and-odometry/differential-drive-odometry.rst
jasondaming: So what is the best way to fix this?  We could add a whole new sentence? Like:

In Java these are each a ``double``, while Python uses a ``float``.  Both must represent the distance traveled by each side in meters.

Daltz333 (approved):

PR #2432, opened 2023-12-11
ADIS16470 fix wrong default port listed

Technologyman00:


rzblue:
Looks like you've got some extraneous commits; those need to be removed

PR #2434, opened 2023-12-11
Adding Python to swerve kinematics and odometry

jasondaming:
I find it odd that the first example uses the m_ notation, but the rest of the examples (which often "depend" on the kinematics object) do not use that variable naming.  I think we should standardize but wanted to have agreement and direction.

Waiting on the swervebot example in Python to be done also.


source/docs/software/kinematics-and-odometry/swerve-drive-odometry.rst
```diff
@@ -67,6 +67,40 @@ The fourth optional argument is the starting pose of your robot on the field (as
...
+      # For both the Kinematics and Odometry classes
+
+      from wpimath.geometry import Translation2d
+      from wpimath.kinematics import SwerveDrive4Kinematics
+      from wpimath.kinematics import SwerveDrive4Odometry
+      from wpimath.geometry import Pose2d
+      from wpimath.geometry import Rotation2d
+
+      # Locations for the swerve drive modules relative to the robot center.
+      m_frontLeftLocation = Translation2d(0.381, 0.381)
```
virtuald, line 82: I would put these all in

```
class MyRobot:
  def robotInit(self):
    self.frontLeftLocation = ...
```

virtuald (changes requested):

virtuald (approved):

sciencewhiz (dismissed):

sciencewhiz (approved):

PR #2439, opened 2023-12-14
Update control-system-hardware.rst

ab12gu:
change color from white to red thats all

red tint to outer light plastic, so must be red


rzblue:
That says "while", not "white"

ab12gu:
oh my bad, getting old man eyes i guess lmfao

PR #2443, opened 2023-12-16
Expanding the Glossary

jasondaming:
Based on numerous CD threads there are a lot of acronyms used.  Wanted to expand this glossary to include them.

Thoughts on abbreviation structure?  We could instead list the term and then include the abbreviation in the description.

Do you think all of this should go into the glossary?  I have heard arguments about breaking it up.  To me a glossary is not really something you read it is instead something that is searched for to answer a specific need.  In this case the size is irrelevant but the completion only becomes more helpful with more entries.


jasondaming:
@Kevin-OConnor I did make a few changes to ZTR figured I should tag you but it is really just correcting an abbreviation that was wrong.  PH->PDH

sciencewhiz:
Seems like a good place to fix #2388

sciencewhiz:
I like all the entries except EPA and OPR. Those seem to be getting too far away from the purpose of frc-docs

jasondaming:
Fixes #2388 

I think EPA and OPR are used often enough in the FRC context to warrant inclusion.  I am not going to die on this hill however so am interested in what others think.

sciencewhiz (approved):

PR #2456, opened 2023-12-24
Update gradle tasks

### LINKED ISSUE #2431: Gradlew Tasks are outdated
There are a variety of examples of this. One example of this is that GradleRIO no longer use Embedded Tools but deploy-utils instead. S

sciencewhiz:
Fixes #2431

### DIFF
--- a/source/docs/software/advanced-gradlerio/gradlew-tasks.rst
+++ b/source/docs/software/advanced-gradlerio/gradlew-tasks.rst
@@ -7,14 +7,15 @@ Build tasks
 -----------
 
 ``./gradlew build`` - Assembles and tests this project. Useful for prebuilding your project without deploying to the roboRIO.
+
 ``./gradlew clean`` - Deletes the build directory.
 
 CompileCommands tasks
 ---------------------
 
-``./gradlew generateCompileCommands`` - Generate compile_commands.json. This is a configuration file that is supported by many Integrated Development Environments.
+``./gradlew generateCompileCommands`` - Generate compile_commands.json for C++ programs. This is a configuration file that is supported by many Integrated Development Environments.
 
-EmbeddedTools tasks
+DeployUtils tasks
 -------------------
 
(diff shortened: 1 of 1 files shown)


source/docs/software/advanced-gradlerio/gradlew-tasks.rst
```diff
@@ -7,14 +7,15 @@ Build tasks
...
 
 ``./gradlew build`` - Assembles and tests this project. Useful for prebuilding your project without deploying to the roboRIO.
+
 ``./gradlew clean`` - Deletes the build directory.
 
 CompileCommands tasks
 ---------------------
 
-``./gradlew generateCompileCommands`` - Generate compile_commands.json. This is a configuration file that is supported by many Integrated Development Environments.
+``./gradlew generateCompileCommands`` - Generate compile_commands.json for C++ programs. This is a configuration file that is supported by many Integrated Development Environments.
```
spacey-sooty, line 16: compile_commands.json is supported by the clangd language server maybe that should be mentioned here?

sciencewhiz: I've seen as many reports of problems with clangd language server as I have successes, so I'm not sure it makes sense to call attention to it specifically

spacey-sooty: I think it might be worth it for the sake of accuracy as it isn't the IDE which supports the compile_commands.json file but instead the Clangd language server. If we don't want to specifically mention Clangd it might be worth it to just say something like: "supported by the language server run in many Integrated Developement Environments" or something along those lines?

sciencewhiz, line 16:
```suggestion
``./gradlew generateCompileCommands`` - Generate compile_commands.json for C++ programs. This is a configuration file that is supported by many Integrated Development Environments and build tools.
```

PR #2468, opened 2023-12-31
Update location of Data Logs.

sciencewhiz:
Fix note notation so they show up

### DIFF
--- a/source/docs/software/telemetry/datalog-download.rst
+++ b/source/docs/software/telemetry/datalog-download.rst
@@ -44,7 +44,7 @@ Downloading Files
 
 After the connection is successfully established, a simplified file browser will be displayed.  This is used to navigate the remote filesystem and select which files to download.  The first text box shows the current directory.  A specific directory can be navigated to by typing it in this text box and pressing Enter.  Alternatively, directory navigation can be performed by clicking on one of the directories that are listed below the remote dir textbox.  Following the list of directories is a table of files.  Only files with a ``.wpilog`` extension are shown, so the table will be empty if there are no log files in the current directory.  The checkbox next to each data log file indicates whether the file should be downloaded.
 
-.. note: On the roboRIO, log files are typically saved to either ``/home/lvuser`` or ``/u`` (USB stick location).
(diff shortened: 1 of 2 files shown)


source/docs/software/telemetry/datalog-download.rst
```diff
@@ -44,7 +44,7 @@ Downloading Files
 
 After the connection is successfully established, a simplified file browser will be displayed.  This is used to navigate the remote filesystem and select which files to download.  The first text box shows the current directory.  A specific directory can be navigated to by typing it in this text box and pressing Enter.  Alternatively, directory navigation can be performed by clicking on one of the directories that are listed below the remote dir textbox.  Following the list of directories is a table of files.  Only files with a ``.wpilog`` extension are shown, so the table will be empty if there are no log files in the current directory.  The checkbox next to each data log file indicates whether the file should be downloaded.
 
-.. note: On the roboRIO, log files are typically saved to either ``/home/lvuser`` or ``/u`` (USB stick location).
+.. note:: On the roboRIO, log files are typically saved to either ``/home/lvuser/logs`` or ``/u/logs`` (USB stick location).
```
jasondaming, line 47: Do we want to direct them to the /mnt location instead so they can find it without having to hunt through the alphabet if it gets a new letter.

sciencewhiz: Is /mnt any better if there's more then one drive mounted?

sciencewhiz: The logging specifies /u https://github.com/wpilibsuite/allwpilib/blob/e7c9f2768377763c72c649cc2d63cbee0532ef46/wpilibj/src/main/java/edu/wpi/first/wpilibj/DataLogManager.java#L215-L220

I'm not sure how you'd end up with logs in something other then /u. I guess you'd have to start with only one drive plugged in that gets logged to. And then plug in a second drive and reboot and that second drive gets assigned /u?

jasondaming: it was the /media folder and not /mnt I was thinking of.  They are just an alternate way to get to /u.  I couldn't reproduce the issues I remember seeing way back when with flash drives so all good.

jasondaming (approved):

PR #2487, opened 2024-01-05
Python for glass docs

jasondaming:


source/docs/software/dashboards/glass/command-based-widgets.rst
```diff
@@ -23,6 +23,13 @@ The :guilabel:`Command Selector`  widget allows you to start and cancel a specif
       MyCommand command{...};
       frc::SmartDashboard::PutData("My Command", &command);
 
+   .. code-block:: python
+
+      from wpilib import SmartDashboard
```
auscompgeek, line 28: Did we want to use from-imports, or `import wpilib`? I recall having a discussion around this and thought we landed on using `import wpilib`, since our newer examples do that.

jasondaming: I like introducing teams to both and this allows us to match the other examples.

auscompgeek (approved):
Python bits LGTM

sciencewhiz (approved):

PR #2491, opened 2024-01-06
Remove old year references

jasondaming:


source/docs/software/hardware-apis/motors/wpi-drive-classes.rst
```diff
@@ -32,7 +32,7 @@ Drive Class Conventions
 Motor Inversion
 ^^^^^^^^^^^^^^^
 
-As of 2022, the right side of the drivetrain is **no longer** inverted by default. It is the responsibility of the user to manage proper inversions for their drivetrain. Users can invert motors by calling ``setInverted()``/``SetInverted()`` on their motor objects.
+The right side of the drivetrain is **no longer** inverted by default. It is the responsibility of the user to manage proper inversions for their drivetrain. Users can invert motors by calling ``setInverted()``/``SetInverted()`` on their motor objects.
```
sciencewhiz, line 35: If this warning is important enough to keep, then I think the year is helpful. I'm not sure that we need to keep a warning that only applies to someone importing code from 2021

jasondaming: I could see people remembering that it was inverted in the past and this makes it clear that it is no longer.  The year is only important when importing code and I think the number of complete code imports from 2021 (and before) should be very few

spacey-sooty: Even if people do remember it was inverted in the past (and don't know its changed) or are importing code from 2021 (both of which seem reasonably rare to me due to the fact its a couple of years ago now then it would be reasonably clear what the issue is after testing no? I think that this could probably be entirely removed.

jasondaming:
Looks good to me

sciencewhiz (approved):

PR #2496, opened 2024-01-06
Remove note about using an old drill as a speed controller

sciencewhiz:
Dangerous with the proliferation of brushless motors


jasondaming:
I think this is still fine.  We do this.  Maybe expand it to specifically state brushed motors?

PR #2505, opened 2024-01-08
Fix adv scope image width

TheTripleV:


jasondaming:
doesn't look like auto width is supported.  Everywhere else it seems like we just pick a number.

jasondaming (approved):

PR #2532, opened 2024-01-15
Rename advanced-controls to controls

DeltaDizzy:
Calling it "advanced" scares people away from its contents, even though the docs that could make them more understandable are under the "advanced" section.


jasondaming:
I really don't mind calling them advanced controls, but since it is already under an "Advanced Programming" section calling it advanced controls implies there is a "basic" controls.  This is just "controls" and yes it is "advanced" for the average team.

So I am good with the change.  This does touch a bunch of files so I think we should have multiple reviews.

sciencewhiz:
We generally don't like to move things around in-season. While the redirects make it so that links aren't broken, there's still people's muscle memory as well as merge issues with PRs

sciencewhiz:
There's conflicts to resolve

PR #2533, opened 2024-01-15
Use same variable name between two code blocks on the same SysID page

Anhysteretic:
The two code blocks use two different variable names for the SysIdRoutine, adjust the naming to be the same variable to reduce confusion for readers.


sciencewhiz:
Per https://docs.wpilib.org/en/latest/docs/contributing/frc-docs/contribution-guidelines.html#creating-a-pr PRs should be created against the main branch, not stable. I changed the branch to main.

sciencewhiz (approved):

PR #2536, opened 2024-01-16
Add known issue for DS MemFree instead of MemAvailable

sciencewhiz:


sciencewhiz:
I could potentially see documenting this somewhere else (in the roborio section, or maybe in the GC article), but this seems as good of place as any and probably will get more views then those other places.

jasondaming:
Have we heard from NI if this is intended behavior or a bug?  That would be my determination if this should go in known issues and elsewhere.

sciencewhiz:
It's more a quirk of linux. NI has been reporting the same thing forever, but it's become more important over the last year or two as people have started using more RAM.

jasondaming (approved):

PR #2537, opened 2024-01-16
Coordinate system

agasser:
Opening PR early to gather feedback. I still plan to include:
- Information about the field coordinate system
  - Mirrored (2023, 2024) vs. rotated (2020, 2022)
  - Discussion about keeping the origin blue regardless of alliance vs. moving the origin based on your alliance. As well as programming tips for how to deal with these.
- Practical programming tips:
  - Joystick input inversion for proper driving with ChassisSpeeds, etc.
  - Gyro rotation may be backwards
  - Gyro and encoder rotation is often continuous
  - Expected coordinate signs (+/-) for swerve kinematics (this is a common starting point for errors that cascade.)

I also plan to replace a few places where coordinate systems are mentioned with links to this section. Those may include the Geometry Classes section, and Drive Class Conventions.


jasondaming:
Great info and idea I think this will help a lot of teams.  I am not sure we (WPILib) have come to a conclusion (or have even through through) on the "correct" method on dealing with different field types.  I am curious to hear what your thoughts are.

I would also think there are other community members that need to be involved with this and have spent a lot of time dealing with coordinate problems.  @mjansen4857 comes to mind but there are probably others too.

This may affect #2531.

agasser:
As far as origin and alliance, I was planning to document both options with some pros and cons, and leave it up to teams to decide. The community seems to be converging on keeping the origin blue, but I'm sure there will be teams that want to move it based on their alliance, especially if they code that already handles it.

As it stands, I would keep #2531 as it is since it matches the surrounding text. I plan for this PR to replace that section and probably include two diagrams.

jasondaming:
I am not saying we have to (or even really can) force users to do it one way, but the sooner we get consensus on what is preferable we can make each of the associated programs interoperate better.

I guess you can't mention an organization?  SlepnirGroup @calcmogul @nlaverdure @jlbabilino @JoshuaAN

calcmogul:
Choreo assumes the origin is blue with NWU coordinates. WPILib's AprilTagFieldLayout loader also assumes this by default; setOrigin() can move it elsewhere via either an arbitrary Pose2d or one of the enum presets.

agasser:
> it's typically a good idea to make your origin near points of interest (e.g., making the origin one of the tag poses you visit frequently).

Can you explain this further? I had no plan to include anything like that in the document, but I can, if you can provide more details.

calcmogul:
You can run into alignment issues with scoring objectives if the tags weren't placed to field-spec. Measuring everything relative to a tag near the point of interest will have less error overall. 971 has a tag calibration system to make this moot.

jasondaming:
I agree everyone seems to be standardizing for this year and I think that is great.  For most teams I really think they will go with what we recommend so long as it makes sense and all of the tools / support is built with that in mind.

What does everyone think would be best for a rotationally symmetric game?  Does that hold true for "half field" games?  Can we say for example always blue 0,0?  Are there any games in the past that would cause insurmountable issues?

agasser:
With the introduction of AprilTags, I'm leaning toward always using blue as the origin. This simplifies pose estimation. Last year we used our alliance corner as the origin, and it meant these things had to be handled:
- The field image in Shuffleboard doesn't flip, so if you want it to be accurate, you have to flip your pose before sending it to shuffleboard.
- Your alliance can change. In the shop, the driver can select the alliance on the driverstation. At a competition, the robot may see a tag before connecting to the FMS. This means code has to deal with the fact that the origin just moved, so the pose estimate needs to change. The tags you saw before the change were actually at different coordinates, so you need to flip your pose and reset your pose estimator.

Keeping the origin blue means the code doesn't have to do anything special for pose estimation, even when the alliance changes. The code only needs to handle field relative driver controls to take the driver's perspective into account.

In both cases, paths still need to be flipped since the field is different, just in different ways.

source/docs/software/coordinate-system.rst
```diff
@@ -0,0 +1,171 @@
...
+   :scale: 200
+   :alt: Unit circle
+
+The figure above shows the unit circle with common angles labeled in degrees (°) and radians (rad). Notice that rotation to the right is negative, and the range for the whole unit circle is -180° to 180° (-Pi radians to Pi radians).
+
+There are some places you may choose to use a different range, such as 0° to 360° or 0 to 1 rotation, but be aware that many core WPILib classes and FRC tools are built with the unit circle above.
+
+.. note:: The range is (-180, 180], meaning it is exclusive of -180° and inclusive of 180°.
+
+Joystick and XBox Controller coordinate system
```
sciencewhiz, line 38: This applies to more then just Joysticks and Xbox controllers, right? Ie PS4Controllers are another class built into WPILib

source/index.rst
```diff
@@ -355,6 +355,7 @@ Community translations can be found in a variety of languages in the bottom-left
    docs/software/what-is-wpilib
    docs/yearly-overview/index
    docs/software/vscode-overview/index
+   docs/software/coordinate-system.rst
```
sciencewhiz, line 358: This feels like it should go under basic programming to me

source/docs/software/coordinate-system.rst
```diff
@@ -0,0 +1,171 @@
...
+
+In most cases in WPILib programming, 0° is aligned with the positive X axis, and 180° is aligned with the negative X axis. CCW rotation is positive, so 90° is aligned with the positive Y axis, and -90° is aligned with the negative Y axis.
+
+.. image:: images/coordinate-system/rotation.svg
+   :scale: 200
+   :alt: Unit circle
+
+The figure above shows the unit circle with common angles labeled in degrees (°) and radians (rad). Notice that rotation to the right is negative, and the range for the whole unit circle is -180° to 180° (-Pi radians to Pi radians).
+
+There are some places you may choose to use a different range, such as 0° to 360° or 0 to 1 rotation, but be aware that many core WPILib classes and FRC tools are built with the unit circle above.
```
sciencewhiz, line 34: It seems like it may also be appropriate that some devices (such as NavX use positive rotation as opposite). Not sure about other common IMUs

agasser: I'll add a warning here. I'm also still planning to add more information about IMUs being CW positive and continuous.

agasser:
This no longer WIP, the document is complete.
Please give it a thorough review. Some of the diagrams and sentences are dense and it's easy to get a label wrong here, or an axis wrong there.

source/docs/software/basic-programming/coordinate-system.rst
```diff
@@ -0,0 +1,325 @@
...
+- pushing forward on the joystick (toward the positive X axis) is a CW rotation around the Y axis, so you get a negative Y value.
+- pushing to the right (toward the postivie Y axis) is a CCW rotation around the X axis, so you get a positive X value.
+- twisting the joystick CW (toward the positive Y axis) is a CCW rotation around the Z axis, so you get a positive Z value.
+
+Using Joystick and controller input to drive a robot
+---------------------------------------------------------
+
+You may have noticed, the coordinate system used by WPILib for the robot is not the same as the coordinate system used for joysticks and controllers. Care needs to be taken to understand the difference, and properly pass driver input to the drive subsystem.
+
+Non-holonomic drivetrain example
```
sciencewhiz, line 70: I think it would be more keeping with the rest of the docs to say Differential drivetrain, rather then non-holonomic, especially since that is the only non-holonomic drivetrain that WPILib supports

source/docs/software/basic-programming/coordinate-system.rst
```diff
@@ -0,0 +1,325 @@
...
+.. figure:: images/coordinate-system/field-blue-alliance.svg
+   :alt: CHARGED UP with blue origin
+   :scale: 200
+
+   CHARGED UP with blue origin
+
+Some advantages to this approach are:
+
+- Pose estimation with AprilTags is simplified. AprilTags throughout the field are unique. If you keep the coordinate system the same regardless of alliance, there is no need for special logic to deal the location of AprilTags on the field relative to your alliance.
+- Many of the tools and libraries used in FRC follow this convention. Some of the tools include: PathPlanner, Choreo, and the ShuffleBoard Field2d widget.
```
sciencewhiz, line 252:
```suggestion
- Many of the tools and libraries used in FRC follow this convention. Some of the tools include: PathPlanner, Choreo, and the ShuffleBoard & Glass Field2d widget.
```

source/docs/software/basic-programming/coordinate-system.rst
```diff
@@ -0,0 +1,325 @@
...
+
+In order to use this approach for field oriented driving, driver input needs to consider the alliance color. When your alliance is red and the driver is standing behind the red alliance wall, they will want the robot to move toward down field toward the blue alliance wall. However, when your alliance is blue, the driver will want the robot to go down field toward the red alliance wall.
+
+A simple way to deal with field oriented driving is to check the alliance color reported by the `DriverStation` class, and invert the drivers controls based on the alliance.
```
sciencewhiz, line 256: This probably needs a callout about how to set the driver station color in the driver station for practice. And also a warning about not checking once in robotInit. They are down below, but really apply to both approaches, right?

source/docs/software/basic-programming/coordinate-system.rst
```diff
@@ -0,0 +1,325 @@
...
+
+Some things you need to consider when using this approach are:
+
+- There are cases where your alliance may change (or appear to change) after the code is initialized. When you are not connected to the FMS at a competition, you can change your alliance station in the Driver Station application at any time. Even when you are at a competition, your robot will usually initialize before connecting to the FMS so you will not have alliance information. If you are not using AprilTags, you may not have anything to adjust when the alliance changes. However, if you are using AprilTags and your robot has seen a tag and used it for pose estimation, you will need to adjust your origin and reset your estimated pose.
+- The field image in the ShuffleBoard Field2d widget follows the *Always blue origin* approach. If you want the widget to display the correct pose for your robot, you will need to change the origin for your estimated pose before sending it to the dashboard.
```
sciencewhiz, line 322:
```suggestion
- The field image in the ShuffleBoard and Glass Field2d widget follows the *Always blue origin* approach. If you want the widget to display the correct pose for your robot, you will need to change the origin for your estimated pose before sending it to the dashboard.
```

source/docs/software/basic-programming/coordinate-system.rst
agasser: I agree. I did that heading when I wasn't sure where I was going to put swerve. I'll update it.

source/docs/software/basic-programming/coordinate-system.rst
agasser: Good ideas. I'll add info about changing the alliance in Driver Station for practice.
With the Always Blue approach, the Driver Station is checked every iteration and an alliance change is handled automatically.
I didn't go into detail on how to handle an alliance change with for the other approach, because it's complicated and varied by code base.
Maybe I can add some detail about about that, I'll have to give it some thought.

sciencewhiz: It was implied that it's checked every iteration because it was in the same code block, but it wasn't stated unless I missed it. My worry is that someone would see that and think they could move it operate once.

agasser: Oh, I see what you mean. The snippet doesn't have context that it's in a block that gets called every iteration, like `execute()` , so it's not clear that `invert` is being set every iteration. I'll find a way to make that more clear.

agasser:
It would be nice to get this reviewed and merged. I've seen two teams today on CD that I would like to refer to this page.
https://www.chiefdelphi.com/t/swerve-x-and-y-flipped-in-odometry/451670
https://www.chiefdelphi.com/t/mecanumdrive-unintended-behavior/451565

sciencewhiz:
> It would be nice to get this reviewed and merged. 

@jasondaming 

> I've seen two teams today on CD that I would like to refer to this page.

you can point them to this page and ask for feedback: https://frc-docs--2537.org.readthedocs.build/en/2537/docs/software/basic-programming/coordinate-system.html

auscompgeek:
> What does everyone think would be best for a rotationally symmetric game?

In 2022 (Rapid React) my team used the centre of the field as the origin. This made a lot of sense for us:

- Since the field was rotationally symmetric, our targeting/pose estimation didn't need to account for alliance.
- We always wanted to aim at the hub, so making it the origin simplified a lot of our calculations.

agasser:
What do the maintainers think, should I add a section about having an arbitrary origin based on a goal? My opinion is that this is an entirely acceptable approach, and there's nothing preventing it, but it is an advanced solution. This document is covering the basics.

sciencewhiz (approved):

sciencewhiz:
I think that can be a different PR. It's kind of niche.

PR #2563, opened 2024-01-29
Make NT Stale Data more discoverable

sciencewhiz:
Move from a sub-section in SmartDashboard into NetworkTables, and add references in each dashboard

### DIFF
--- a/source/docs/software/networktables/stale-data.rst
+++ b/source/docs/software/networktables/stale-data.rst
@@ -1,0 +1,8 @@
+Stale Data
+==========
+
+NetworkTables acts as a distributed table of name and value pairs. If a name/value pair is added to either the server (robot) or any client (dashboard, co-processor, limelight, etc) it is replicated to all devices. If a name/value pair is deleted from, say, the robot but any NetworkTables clients are still running, then when the robot is restarted, the old values will still appear in the client because they never stopped running and continue to have those values in their tables. When the robot restarts, those old values will be replicated to the robot.
+
+To ensure that the robot and clients are showing current values, it is necessary to shut down all NetworkTables clients and robot at the same time. That way, old values that one is holding won't get replicated to the others. Once everything is shut down, the robot and clients can be restarted.
+
+This usually isn't a problem if the program isn't constantly changing, but if the program is in development and the set of keys being added to NetworkTables is constantly changing, then it might be necessary to do the restart of everything to accurately see what is current.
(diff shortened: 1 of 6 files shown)


source/docs/software/networktables/stale-data.rst
```diff
@@ -0,0 +1,8 @@
+Stale Data
+==========
+
+NetworkTables acts as a distributed table of name and value pairs. If a name/value pair is added to either the server (robot) or any client (dashboard, co-processor, limelight, etc) it is replicated to all devices. If a name/value pair is deleted from, say, the robot but any NetworkTables clients are still running, then when the robot is restarted, the old values will still appear in the client because they never stopped running and continue to have those values in their tables. When the robot restarts, those old values will be replicated to the robot.
```
PeterJohnson, line 4: This is more nuanced than stated.  NT4 has different categories of topic value preservation: persistent, retained, and transient.  Only persistent and retained topic values are preserved by the server when all publishers are gone, and only persistent topic values are preserved through a server reboot. As NT4 defaults to transient publishes, and a client disconnect acts as an “unpublish all”, stale values are rare unless a client specifically marks a topic as retained.

NT3 clients mark all their published topics as retained, so there’s a greater likelihood those values will be stale.  But this still only applies to topics the NT3 client actually writes to.

PeterJohnson: It’s probably also worth noting here that the timestamp column in OutlineViewer can be used to see what topics are being regularly updated.

jasondaming: I agree it is more nuanced than written but I don't think its much.

So with Limelight for example (presumably still on NT3? will never know till (if?) we get a new image) would leave all of its stale values?  I am just thinking of one of the most common use cases for teams.

source/docs/software/networktables/stale-data.rst
```diff
@@ -0,0 +1,8 @@
...
+
+To ensure that the robot and clients are showing current values, it is necessary to shut down all NetworkTables clients and robot at the same time. That way, old values that one is holding won't get replicated to the others. Once everything is shut down, the robot and clients can be restarted.
```
jasondaming, line 6:
```suggestion
To ensure that the robot and clients are showing current values, it is necessary to shut down all NetworkTables clients and the robot at the same time. That way, old values one is holding won't get replicated to the others. Once everything is shut down, the robot and clients can be restarted.
```

source/docs/software/networktables/stale-data.rst
sciencewhiz: However, Limelight isn't changing the data that it sends often. like changing robot programs

sciencewhiz:
Since this is only really an issue with NT3, going to close this and keep in SD

PR #2564, opened 2024-01-29
Add intended use case and footguns to ProxyCommand docs

DeltaDizzy:


Oblarg (approved):

source/docs/software/commandbased/command-compositions.rst
```diff
@@ -204,19 +204,29 @@ By default, composition members are run through the command composition, and are
 
 ``ProxyCommand`` (`Java <https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/wpilibj2/command/ProxyCommand.html>`__, `C++ <https://github.wpilib.org/allwpilib/docs/release/cpp/classfrc2_1_1_proxy_command.html>`__), also creatable using the ``.asProxy()`` decorator (`Java <https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/wpilibj2/command/Command.html#asProxy()>`__, `C++ <https://github.wpilib.org/allwpilib/docs/release/cpp/classfrc2_1_1_command_ptr.html#aa45784053431393e3277e5bc5ae7f751>`__), schedules a command "by proxy": the command is scheduled when the proxy is scheduled, and the proxy finishes when the command finishes. In the case of "forking off" from a command composition, this allows the group to track the command's progress without it being in the composition.
 
+.. warning:: Do not use ``ProxyCommand`` unless you are sure of what you are doing and there is no other way to accomplish your need! Proxying is only intended for use as an escape hatch from command group requirement aggregation.
```
sciencewhiz, line 207: I'm not sure that people will understand by the term command group requirement aggregation. This would probably be more clear if it was after the next paragraph and used similar language.

source/docs/software/commandbased/command-compositions.rst
```diff
@@ -204,19 +204,32 @@ By default, composition members are run through the command composition, and are
...
 
+
+Command groups and compositions inherit the union of their compoments' requirements and requirements are immutable. Therefore, a SequentialCommandGroup that intakes a game piece, indexes it, aims a shooter, and shoots it would reserve all three subsystems (the intake, indexer, and shooter), precluding any of those subsystems from performing other operations in their "downtime". To solve this, the subsystems that should only be reserved for the composition while they are participating in it should have their commands proxied.
```
Starlight220, line 208: "SequentialCommandGroup" should be code-formatted and linked to the appropriate section in the compositions page (the link is more important than the formatting, if it's one or the other).

I'm not a fan of the "solve" phrasing, as it implies that this is necessarily a problem. There are use cases in both directions, so "if this is not desired" would be preferable here.
Additionally, "participating" isn't well-defined here.

Starlight220 (changes requested):
"[command] compositions" is the new-ish term for "command groups". Don't use the outdated one, and definitely don't use both (that leads to more confusion).

Starlight220 (approved):
I meant to replace only the `s/group/composition/` cases in scope of this PR...
Not a blocker imo.

PR #2566, opened 2024-02-02
Update Known Issues for Game Tools Patch 1

sciencewhiz:
Move print error and tags to fixed
Add updating DS to first troubleshooting step
Add updating Wi-Fi drivers as a step

### DIFF
--- a/source/docs/yearly-overview/known-issues.rst
+++ b/source/docs/yearly-overview/known-issues.rst
@@ -177,7 +177,7 @@ On Linux/macOS execute the following:
 
 After running those, you’ll need to refresh Java intellisense in VS Code for it to pick up the new files. You can do so by running the ``Clean Java Language Server Workspace`` command in VS Code.
 
-Fixed in Game TGols 2024 Patch 1
+Fixed in Game Tools 2024 Patch 1
 --------------------------------
 
 Driver Station internal issue with print error and tags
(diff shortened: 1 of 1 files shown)


source/docs/yearly-overview/known-issues.rst
```diff
@@ -178,3 +176,13 @@ On Linux/macOS execute the following:
    curl https://repo.maven.apache.org/maven2/org/junit/junit-bom/5.10.1/junit-bom-5.10.1.module -o ~/wpilib/2024/maven/org/junit/junit-bom/5.10.1/junit-bom-5.10.1.module
 
 After running those, you’ll need to refresh Java intellisense in VS Code for it to pick up the new files. You can do so by running the ``Clean Java Language Server Workspace`` command in VS Code.
+
+Fixed in Game TGols 2024 Patch 1
```
jasondaming, line 180:
```suggestion
Fixed in Game Tools 2024 Patch 1
```

jasondaming:
Maybe we should also add a line about the fix to https://github.com/wpilibsuite/allwpilib/issues/6206.  I know this wasn't on the known issues before but I think it is good to show all (or at least as many that might be relevant) of the things they are fixing.

sciencewhiz:
There were a few other issues fixed that were only reported by 1 or two people. Don't think those rise to being in the known issues, just close the issues.

PeterJohnson (approved):

PR #2573, opened 2024-02-17
Update linkcheck user agent and update links

sciencewhiz:


sciencewhiz:
I commented out link exclusions that pass now, in case they're flaky, but I ran a few trials and they all passed.

PR #2581, opened 2024-02-19
Update field oriented inversion example snippet

Theinatorinator:
Previously, the field-oriented inversion example would multiply chassis speeds to zero if the driver station alliance was blue. Now it properly multiplies by one if the alliance is blue and by negative one if the alliance is red.


sciencewhiz:
Per https://docs.wpilib.org/en/stable/docs/contributing/frc-docs/contribution-guidelines.html#creating-a-pr, PRs should be targeted to main not stable. I updated the base branch.

PR #2592, opened 2024-03-05
Add note that the DS 'random disables' won't occur on FMS

rzblue:


sciencewhiz (approved):

jasondaming:
What do you think about adding a :term: to FMS so it will show the glossary on that.  I think most people reading this would know the FMS but never know

sciencewhiz (approved):

PR #2594, opened 2024-03-07
Update measuring bandwidth utilization article

rzblue:
Corrected max bandwidth utilization to be the newer 4MBit/s (previously 7MBit/s)
Old screenshots were from windows 7 and the wireshark section was no longer correct


rzblue:
That's true, hadn't thought of that. Way easier than the performance monitor.

sciencewhiz (approved):

PR #2596, opened 2024-03-10
Update slew-rate-limiter.rst for c++ docs

Dannykcw:


source/docs/software/advanced-controls/filters/slew-rate-limiter.rst
```diff
@@ -52,7 +52,7 @@ Once your filter has been created, using it is easy - simply call the ``calculat
   .. code-block:: c++
 
     // Calculates the next value of the output
-    filter.Calculate(input);
+    filter.Calculate(units::voltage::volt_t{input}).value();
```
spacey-sooty, line 55: This doesn't make sense, why are you creating a unit then calling .value on the newly created unit?

source/docs/software/advanced-controls/filters/slew-rate-limiter.rst
```diff
@@ -82,7 +82,7 @@ A typical use of a SlewRateLimiter is to limit the acceleration of a robot's dri
     drivetrain.ArcadeDrive(forward, turn);
 
     // Slew-rate limits the forward/backward input, limiting forward/backward acceleration
-    drivetrain.ArcadeDrive(filter.Calculate(forward), turn);
+    drivetrain.ArcadeDrive(filter.Calculate(units::voltage::volt_t{forward}).value(), turn);
```
spacey-sooty, line 85: Same point here

spacey-sooty (changes requested):

spacey-sooty:
> The c++ function does not return a double nor require one, but rather requires an units::voltage::volt_t object and returns an object of the same type. Maybe I can take out the .value method call, but for the purpose of getting a double as the return, it makes more sense.
> 
> We should have gone with Java...

If the C++ filter does not take a double then calling .value is 100% wrong here. I don't think the conversion is necessary in the docs as it is specified in the function signature and most C++ functions operate within the Units library.

Dannykcw:
The reason .value() is called is because of `frc::DifferentialDrive::ArcadeDrive`. So for reference, these are the ArcadeDrive params: `void frc::DifferentialDrive::ArcadeDrive(double xSpeed, double zRotation, bool squareInputs = true)`.

So both `xSpeed` and `zRotation` are both double. But `frc::SlewRateLimiter` for c++ looks like this:
`inline units::voltage::volt_t frc::SlewRateLimiter<units::voltage::volts>::Calculate(units::voltage::volt_t input)
`
Where the output is `units::voltage::volt_t`. This means the output is incompatible. I will say I am not too familiar with the units library, but I do know that `value()` returns a double, which would be used in the `ArcadeDrive` call.

In any case, 
image
would be a demo of the above, where the absence of the `value()` would return an error at compile time. Unless java does auto conversions, or if there is some other way to pass it as a double, this is what the user would have to do.

I think you are valid in that the call on line 55 is less sensible. I called` value()` because I don't want to confuse rookies, where I am doing this thing in one place and this thing in another without saying why. It would make sense to remove the call if it causes confusion, which it did.

Also, one more thing, is ArcadeDrive take in three argument, but in my commit I missed one. That would be my fault.

Also, if the implementation for `calculate` change to align with that of java: `public double calculate (double input)` where the return is a double, then we wouldn't have the issue. I can make a pr on the actual wpilib repo if that is more preferable.

In any case, let me know if I misunderstood anything. If you think this doesn't make sense, then I guess I should be moving on to something else...

Dannykcw:
The reason `value()` is called is because of `frc::DifferentialDrive::ArcadeDrive`. So for reference, these are the `ArcadeDrive` params: `void frc::DifferentialDrive::ArcadeDrive(double xSpeed, double zRotation, bool squareInputs = true)`.

So both `xSpeed` and `zRotation` are both double. But `frc::SlewRateLimiter` for c++ looks like this:
`inline units::voltage::volt_t frc::SlewRateLimiter<units::voltage::volts>::Calculate(units::voltage::volt_t input)` 
Where the output is `units::voltage::volt_t`. This means the output is incompatible. I will say I am not too familiar with the units library, but I do know that value() returns a double, which would be used in the ArcadeDrive call.

In any case,
image

would be a demo of the above issue where the absence of the value() would return an error at compile time. Unless java does auto conversions, or if there is some other way to pass it as a double, this is what the user would have to do.

I think you are valid in that the call on line 55 is less sensible. I called `value()` because I don't want to confuse rookies, where I am doing this thing in one place and this thing in another without saying why. It would make sense to remove the call if it causes confusion, which it did.

Also, one more thing, is `ArcadeDrive` take in three argument, but in my commit I have two. That would be my fault and something that I would have to fix.

Also, if the implementation for calculate change to align with that of java: `public double calculate (double input)` where the return is a double, then we wouldn't have the issue. I can make a pr on the actual wpilib repo if that is more preferable.

In any case, let me know if I misunderstood anything. If you think this doesn't make sense, then I guess I should be moving on to something else...

source/docs/software/advanced-controls/filters/slew-rate-limiter.rst
Dannykcw: Please see the new comment for clarification.

spacey-sooty:
> The reason `value()` is called is because of `frc::DifferentialDrive::ArcadeDrive`. So for reference, these are the `ArcadeDrive` params: `void frc::DifferentialDrive::ArcadeDrive(double xSpeed, double zRotation, bool squareInputs = true)`.
> 
> So both `xSpeed` and `zRotation` are both double. But `frc::SlewRateLimiter` for c++ looks like this: `inline units::voltage::volt_t frc::SlewRateLimiter<units::voltage::volts>::Calculate(units::voltage::volt_t input)` Where the output is `units::voltage::volt_t`. This means the output is incompatible. I will say I am not too familiar with the units library, but I do know that `value()` returns a double, which would be used in the ArcadeDrive call.
> 
> In any case, image
> 
> would be a demo of the above issue where the absence of the value() would return an error at compile time. Unless c++ does auto conversions, or if there is some other way to pass it as a double, this is what the user would have to do.
> 
> I think you are valid in that the call on line 55 is less sensible. I called `value()` because I don't want to confuse rookies, where I am doing this thing in one place and this thing in another without saying why. It would make sense to remove the call if it causes confusion, which it did.
> 
> Also, if the implementation for calculate change to align with that of java: `public double calculate (double input)` where the return is a double, then we wouldn't have the issue. I can make a pr on the actual wpilib repo if that is more preferable.
> 
> In any case, let me know if I misunderstood anything. If you think this doesn't make sense, then I guess I should be moving on to something else...

It taking a double makes sense, but it was confusing that you would wrap it in a constructor for a `units::volt_t` then call .value as this is redundant (your value should already be a double or you can make the .value call directly). It certainly isn't desirable to make this use doubles instead of units but this doesn't adrress this issue in the docs properly.

source/docs/software/advanced-controls/filters/slew-rate-limiter.rst
spacey-sooty: The new comment doesnt address this the whole change here should be removed from what I've seen but this definitely isn't correct

source/docs/software/advanced-controls/filters/slew-rate-limiter.rst
spacey-sooty: This is
1. Redundant
2. Doesn't address the issue you described

spacey-sooty (changes requested):

Dannykcw:
I removed the first value() call as you suggested. 

The reason I made this pr is because the original code does not compile because the java and c++ implementation of the `Calculate` method is different. For the purpose of using the slew rate limiter in `ArcadeDrive`, you would need to pass in `units::voltage::volt_t` instead of double.
image

My pr would fix the compilation issue, and demonstrate the use of the Calculate function for the purpose of damping the input to the `ArcadeDrive` method that would compile, as originally intended.

source/docs/software/advanced-controls/filters/slew-rate-limiter.rst
spacey-sooty, line 85: This still hasn't been fixed wrapping something in a constructor only to get the underlying double HAS to be wrong

spacey-sooty (changes requested):
Does the docs page show where you are getting the input from? If not I don't think this change should be made. If something takes in a double and is returning a volts then it should have .value directly called on it

Dannykcw:
> Does the docs page show where you are getting the input from? If not I don't think this change should be made. If something takes in a double and is returning a volts then it should have .value directly called on it

Calculate() in c++ takes in a units object type when instantiated, and uses the units object of the specified type in its methods. On the other hand, Java stores the property as a double, and uses double in its methods, which is why the returns for the same method are different. 

https://github.wpilib.org/allwpilib/docs/release/cpp/_slew_rate_limiter_8h_source.html
https://paste.me/paste/6815e735-98dc-4309-5000-6566a734b411#4d462b7190529a03317a2d90956bb5050ee6a34ef446f56a16cf9f75cdc7226e
(Java implimentation)

I would expect that even if the code isn't "as is", but the original wouldn't even compile due to the varying implementation, and that is the crust of my issue.

> wrapping something in a constructor only to get the underlying double HAS to be wrong

The problem is that in c++, Caculate() doesn't return a double, but the unit that the slewRateLimiter is constructed with. ArcadeDrive requires a double, and the only way to convert any units object into a double would be to use value(). This is also the way wpilib docs [recommends](https://docs.wpilib.org/en/stable/docs/software/basic-programming/cpp-units.html#the-c-units-library). Of course, I would like to learn what the more "correct" way to do this. And if you dislike my implementation, feel free to change it so that would align to your standards.

> Does the docs page show where you are getting the input from? If not I don't think this change should be made.

You are right that the docs does not tell you where the input comes from. But I don't see why this is the issue. The user is supposed to figure out what to use as input, and I don't see how that would invalidate my point.

source/docs/software/advanced-controls/filters/slew-rate-limiter.rst
spacey-sooty, line 85:
```suggestion
    drivetrain.ArcadeDrive(filter.Calculate(forward).value(), turn);
```

spacey-sooty: If you are getting the input in most cases it should be a units::volt_t as that is the correct way to do things in C++ so it should be like this

source/docs/software/advanced-controls/filters/slew-rate-limiter.rst
```diff
@@ -52,7 +52,7 @@ Once your filter has been created, using it is easy - simply call the ``calculat
   .. code-block:: c++
 
     // Calculates the next value of the output
-    filter.Calculate(input);
+    filter.Calculate(units::voltage::volt_t{input});
```
spacey-sooty, line 55:
```suggestion
    filter.Calculate(input);
```
Your input should be a units::volt_t as C++ should be using the units library

spacey-sooty (changes requested):
I just had a look at the ArcadeDrive function as well and it appears to want a speed as a double meaning even if your input to the filter.Calculate needed to be converted to units (it shouldn't) then it would be done with a speed unit not a voltage unit

Dannykcw:
Your suggestion was definitely cleaner then mine. 

I think a few more changes would help also. For instance, changing the constructor to use meter / sec in the example and write a notice that input should be of the same object type as the constructor, or something similar. But they aren't necessary.

I have committed your suggestions. Thanks for your input.

sciencewhiz:
I'm confused as to why this is needed. the note above says `The C++ example below templates the filter on units::scalar for use with doubles, since joystick values are typically dimensionless.` I'm guessing you aren't using units::scalar?

Dannykcw:
@sciencewhiz You are right that I did not use units:scalar. I did read the handle, but since the first code example used volts, and the units library page made no mention of scalar or dimensionless.h, I did not realize that such a thing even exist.

I changed the slewratelimiter type to scalar, and it seems to work fine without the value() call. I still think the docs uses of volts is misleading, so I changed the example to use scalar instead. It works directly with the example of using it to limit the Arcadedrive input.

Since I think this issue is more of me not understanding what I was supposed to do, if there isn't a lot of value in me changing it, please let me know and I will drop the pr.

And I like to apologize to both you and @spacey-sooty for the trouble.

PR #2597, opened 2024-03-11
expand getAlliance documentation

bovlb:
* Emphasize tri-state nature
* Explain how to set alliance color at home
* Explain when it is and is not available on FMS


ThadHouse:
There's actually no behavior difference between FMS and home use. If the ds has not connected, you don't have an alliance, and it will be none. Thats the same if you're on the field vs at home.

bovlb:
> There's actually no behavior difference between FMS and home use. If the ds has not connected, you don't have an alliance, and it will be none. Thats the same if you're on the field vs at home.

OK.  Improved, I think.

source/docs/software/basic-programming/alliancecolor.rst
```diff
@@ -3,6 +3,8 @@ Get Alliance Color
 
 The ``DriverStation`` class (`Java <https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/wpilibj/DriverStation.html>`__, `C++ <https://github.wpilib.org/allwpilib/docs/release/cpp/classfrc_1_1_driver_station.html>`__, :py:class:`Python <robotpy:wpilib.DriverStation>`) has many useful features for getting data from the Driver Station computer.  One of the most important features is ``getAlliance`` (Java & Python) / ``GetAlliance`` (C++).
 
+Note that there are three cases: red, blue, and no color yet.  It is important that code handles the third case correctly because the alliance color will not be available until the DriverStation connects.  In particular, code should not assume that the alliance color will be available during constructor methods or `robotInit`, but it should be available by the time `autoInit` or `teleopInit` is called.  FMS will set the alliance color automatically; when not connected to FMS, the alliance color can be set from the Driver Station (see :ref:`"Team Station" on the Operation Tab <docs/software/driverstation/driver-station:Operation Tab>`).
```
Starlight220, line 6:
```suggestion
Note that there are three cases: red, blue, and no color yet.  It is important that code handles the third case correctly because the alliance color will not be available until the Driver Station connects.  In particular, code should not assume that the alliance color will be available during constructor methods or `robotInit`, but it should be available by the time `autoInit` or `teleopInit` is called.  FMS will set the alliance color automatically; when not connected to FMS, the alliance color can be set from the Driver Station (see :ref:`"Team Station" on the Operation Tab <docs/software/driverstation/driver-station:Operation Tab>`).
```
Unless it's the class being mentioned (and then it should be in code formatting), "Driver Station" should be two words.

sciencewhiz (approved):

PR #2599, opened 2024-03-11
[docs] Make C++ example for advantagescope swerve states compilable

nobody5050:
When I was initially implementing this in my codebase the example in the docs confused me because although it seemed straightforward it wouldn’t actually compile. The publisher needs a std::span of states rather than a vector as indicated by the docs. While this is definitely not the most concise implementation, (it’s ripped from our season code) it’s a lot more helpful in my opinion.

### DIFF
--- a/source/docs/software/kinematics-and-odometry/swerve-drive-kinematics.rst
+++ b/source/docs/software/kinematics-and-odometry/swerve-drive-kinematics.rst
@@ -303,14 +303,12 @@ By recording a set of swerve module states using :ref:`NetworkTables <docs/softw
 
         void Periodic() {
           // Periodically send a set of module states
-          swervePublisher.Set(
-            std::vector{
-              frontLeftState,
-              frontRightState,
-              backLeftState,
-              backRightState
-            }
-          );
+          std::vector states_vec = {m_frontLeft.GetState(), m_frontRight.GetState(),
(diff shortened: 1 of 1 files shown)


source/docs/software/kinematics-and-odometry/swerve-drive-kinematics.rst
```diff
@@ -303,14 +303,12 @@ By recording a set of swerve module states using :ref:`NetworkTables <docs/softw
...
-              backLeftState,
-              backRightState
-            }
-          );
+          std::vector states_vec = {m_frontLeft.GetState(), m_frontRight.GetState(),
+                            m_rearLeft.GetState(), m_rearRight.GetState()};
+         std::span<frc::SwerveModuleState, 4> states(states_vec.begin(),
+                                              states_vec.end());
+
+         publisher.Set(states);
```
PeterJohnson, line 311: It shouldn't be necessary to explicitly create a std::span.  It should work to just pass states_vec directly to the publisher, as there's an implicit conversion from std::vector to std::span.

PeterJohnson: It should also be possible to simply use double-braces here, although I've not tested it:
```c++
swervePublisher.Set(
            {{
              frontLeftState,
              frontRightState,
              backLeftState,
              backRightState
            }}
          );
```

PR #2609, opened 2024-03-18
Update introduction-to-pid.rst with DigiKey PID Video

horner:


calcmogul:
This video is good until the 10 minute mark when it starts talking about the integral and derivative terms.

## Integral control

I disagree with its assertion that integral control is better than a constant offset (aka feedforward). A model-based feedforward + P controller gives superior performance compared to a PI controller, and an accurate feedforward is usually easy to make; `u = Kₛsgn(v) + Kᵥv` works for most things with a DC motor.

Also, there's better ways to compensate for unmodeled disturbances like [input error estimation](https://file.tavsys.net/control/controls-engineering-in-frc.pdf#subsection.6.7.2), which only integrates if the system is deviating from its model. That is, you integrate a model of the system, then output a corrective action based on the difference between the expected position and actual position. The model is reasonably easy to get in FRC with [SysId](https://docs.wpilib.org/en/stable/docs/software/advanced-controls/system-identification/index.html) because most FRC systems fit into `dx/dt = Ax + Bu -> dx/dt = −Kᵥ/Kₐx + 1/Kₐu`. In any case, pure integrators should have exceedingly small controller contributions, if any.

## Derivative control

The video's interpretation of derivative control does nothing to explain how motion profiles work, which are common in FRC. That's why our existing docs use the [alternate interpretation](https://docs.wpilib.org/en/stable/docs/software/advanced-controls/introduction/introduction-to-pid.html#derivative-term) that for a position system, Kp is proportional control of position, and Kd is proportional control of velocity (the velocity setpoint is implicitly set by the rate of change of the position setpoint). This gives the gains accurate physical meaning in more cases that better informs how to tune them.

## The video's audience

I am aware many students these days prefer videos to text, but the videos need to have the right information for our audience. This video seems aimed at PLC users who don't have a system model, and if they do start modeling, they'll do so with transfer functions (classical control theory). FRC can take advantage of feedforwards and modern control theory¹ instead, so I think this video is counterproductive for the rest of our docs.  

¹ Modern control theory is easier to learn and more powerful than classical. The only place you'd use classical is for frequency shaping, like EE power circuit design or rocket vibrational dampening.

sciencewhiz:
Thanks for the contribution! Besides those issues, I think the overall length of the video may also be an issue. We're definitely open to other learning mediums, but we need to make sure they're consistent with the rest of the content. I'm going to close this but if you have another suggestion, feel free to open a new PR.

horner:
@calcmogul Thanks for your feedback on the video.  I appreciate you taking the time to explain the befits of a PD over a PID and modern control theory. I also learned a lot based on the PDF article you shared [input error estimation](https://file.tavsys.net/control/controls-engineering-in-frc.pdf#subsection.6.7.2). I found that document to be very helpful.  I edited my PR to include that link.   There is a lot of good videos in the PDF but none for the PID or control theory so I am going to make some PR's over there!

I think it's worth pointing out that the video on https://github.com/wpilibsuite/frc-docs/blob/main/source/docs/software/advanced-controls/introduction/pid-video.rst?plain=1 is similar to the DigiKey PID Video in that it does not cover the modern control theory.

Check out this video (particularly 6:45 where feedforward is introduced)
https://www.youtube.com/watch?v=FW_ay7K4jPE

Curious if you think it would be a better resource.

Thanks again for sharing !

horner:
https://github.com/wpilibsuite/frc-docs/compare/main...horner:frc-docs:patch-2?expand=1

PR #2622, opened 2024-03-30
Added a note explaining that Measure math is done using the "Base Unit" of a dimension

falOn-Dev:
I added a note to `java-units.rst` explaining that the various measure math methods use the "Base Unit" of the dimension that the measure is for.


sciencewhiz:
I'm not sure I understand the need for the note. 

When I run the code in the example that adds meters to inches, and add a print, I get "TotalDistance: 2.000e+01 in". The note would imply it should be in the base unit (meters). Of course you can get any other distance unit.

```
    Measure<Distance> distance1 = Inches.of(10);
    Measure<Distance> distance2 = Meters.of(0.254);
    
    Measure<Distance> totalDistance = distance1.plus(distance2);
    System.out.println("TotalDistance: " + totalDistance);
```

falOn-Dev:
I just think it's important to be transparent about underlying systems, in case a team doesn't want to do math with radians, meters, etc.

SamCarlberg:
> in case a team doesn't want to do math with radians, meters, etc.

I'm curious, in what cases do you think this could be problematic?

falOn-Dev:
It isn't a matter of problematic IMO, it's a matter of transparency in underlying systems, there's no reason not to add a sentence in that explains how the library handles math or other things.

PR #2638, opened 2024-05-01
Add link to NT specification

chondl:
[Recent CD post](https://www.chiefdelphi.com/t/networktables-node-js/465025/3) from @PeterJohnson contained link to the specification which wasn't directly referenced from the documentation.  At this time, Google search results for "frc network tables specification" do not include the current spec (at least when I do the search).  The current spec should be readily available from the documentation for easy reference.


chondl:
Apologies for the whitespace changes introduce by editing in the GitHub UI.  Not sure how to remove.

PeterJohnson (approved):

PR #2655, opened 2024-05-15
Update for five digit team numbers

sciencewhiz:


jasondaming:
I suggested to modify the traditional TE.AM to XTE.AM, what do you think? @sciencewhiz 
https://www.chiefdelphi.com/t/team-numbers-have-officially-hit-5-digits-what-now/466213/24

PeterJohnson:
I liked the suggestion of `10.STE.AM.x` later in the thread.

jasondaming:
I was worried that would change the "obviousness" that it was the team number.  I do think that would prevent confusion around `10.XTE.AM.X` which less ideally uses X twice (we could also change that variable, we already use Y and Z in other places).  Either way I think both of these are much better than just TE.AM

sciencewhiz:
I don't like STE.AM, it uses something unrelated in a confusing way. 

I think TE.AM with explanation is sufficient, but if I were to choose something else it'd prefer TEA.M#

PeterJohnson:
I think just sticking with TE.AM works too with a simple explanation that "AM" is always the last two digits of the team number, so "TE" is 3 digits for 5-digit teams.

source/docs/networking/networking-introduction/ip-configurations.rst
```diff
@@ -6,14 +6,18 @@ IP Configurations
 TE.AM IP Notation
 -----------------
 
-The notation TE.AM is used as part of IPs in numerous places in this document. This notation refers to splitting your four digit team number into two digit pairs for the IP address octets.
+The notation TE.AM is used as part of IPs in numerous places in this document. This notation refers to splitting your five digit team number into digits for the IP address octets. where AM is the last two digits of the team number, and TE is the first three digits. Leading zeros are optional. This scheme supports team numbers up to 25599.
```
jasondaming, line 9:
```suggestion
The notation TE.AM is used as part of IPs in numerous places in this document. This notation refers to splitting your five digit team number into digits for the IP address octets. Where AM is the last two digits of the team number, and TE is the first three digits. Leading zeros are optional. This scheme supports team numbers up to 25599.
```

jasondaming (approved):

PR #2662, opened 2024-06-11
Simplify base-branch checking

sciencewhiz:
The action has been broken for a while, and this provides a better message


sciencewhiz:
Example run with a failure. https://github.com/sciencewhiz/frc-docs/actions/runs/9469866648/job/26089507043?pr=8

Issue #2651, opened 2024-05-09
Replace RAMSETE documentation


sciencewhiz:
https://github.com/wpilibsuite/allwpilib/pull/6494 deprecates RAMSETE controller. The following articles and tutorials need to be updated

- [ ] https://docs.wpilib.org/en/stable/docs/software/pathplanning/trajectory-tutorial/index.html
- [ ] https://docs.wpilib.org/en/stable/docs/software/advanced-controls/trajectories/ramsete.html

PR #2668, opened 2024-06-16
Pin Ramsete example links to last tag

sciencewhiz:
Removed in main. Issue #2651 exists for updating


Starlight220:
Are there RLIs that should be pinned as well?

sciencewhiz:
No, because RLIs are already pinned

Starlight220:
I meant that they'll be ignored and not automatically updated.

PR #2673, opened 2024-07-09
Update index.rst to include static method example

ArchdukeTim:
Include static method reference


sciencewhiz:
Per https://docs.wpilib.org/en/stable/docs/contributing/frc-docs/contribution-guidelines.html#creating-a-pr PRs should be based on main, not stable. I updated the base branch of the PR.

sciencewhiz (approved):

PR #2686, opened 2024-08-02
Add additional steps for DS performance

sciencewhiz:
keeping DS visible, and setting power plan to best performance

### DIFF
--- a/source/docs/software/driverstation/driver-station-best-practices.rst
+++ b/source/docs/software/driverstation/driver-station-best-practices.rst
@@ -71,11 +71,11 @@ Before Each Match
 2. Close programs that aren’t needed during the match – e.g., Visual Studio Code or LabView – when you are competing.
 3. Bring your laptop charger to the field. Power is provided for you in each player station.
 4. Fasten your laptop with hook-and-loop tape to the player station shelf. You never know when your alliance partner will have an autonomous programming issue and blast the wall.
-5. Ensure driver station is visible during the match for best performance, and to ease troubleshooting.
-6. Ensure joysticks and controllers are assigned to the correct USB ports.
+5. Ensure joysticks and controllers are assigned to the correct USB ports.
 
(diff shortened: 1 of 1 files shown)


source/docs/software/driverstation/driver-station-best-practices.rst
```diff
@@ -70,10 +71,11 @@ Before Each Match
 2. Close programs that aren’t needed during the match – e.g., Visual Studio Code or LabView – when you are competing.
 3. Bring your laptop charger to the field. Power is provided for you in each player station.
 4. Fasten your laptop with hook-and-loop tape to the player station shelf. You never know when your alliance partner will have an autonomous programming issue and blast the wall.
-5. Ensure joysticks and controllers are assigned to the correct USB ports.
+5. Ensure driver station is visible during the match for best performance, and to ease troubleshooting.
```
rzblue, line 74: This should emphasize that the DS window should be fully visible and that not doing so can cause significant performance issues

sciencewhiz: Doesn't it only have to be partially visible?

https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-setprocessinformation#remarks says the throttling occurs when the window is fully occluded. Since it's trying to keep gui responsiveness, it makes sense that if part of it is still visible, it shouldn't be throttled because otherwise it would be noticeable.

TheTripleV: No...

https://learn.microsoft.com/en-us/windows/win32/procthread/quality-of-service

Explicitly setting process info should bypass all of this according to the table.

But fully visible + active window, partially visible, and fully occluded are separate tiers.

I would recommend making sure the ds window is the active app (clicked on last)

sciencewhiz: I read that as for active and fully visible to have an effect, it also needs to have the process info set. But there's too many and and ors and not enough parentheses to evaluate ;-)

PR #2688, opened 2024-08-03
Add an article documenting auto importing for Java code

### LINKED ISSUE #2675: Document usage of VSCode's auto importer for Java

WispySparks:
Resolves #2675


source/docs/software/vscode-overview/organize-imports.rst
```diff
@@ -0,0 +1,16 @@
+Organize Imports
+================
+
+Visual Studio Code provides many settings to speed up programming. The Java extensions installed by WPILib allow for a setting to automatically insert the correct imports and organize them in your Java code when that file is saved. To enable this feature, first navigate to the user settings by opening the command palette with :kbd:`Ctrl+Shift+P`, then type User Settings (JSON) and press enter. Finally, add this section of code to the end of the user settings.
```
sciencewhiz, line 4: In most cases, the imports are put in automatically as code is written. The main benefit seems to be the organizational aspect, But it seems like someone who cares that much would also want to use a full code formatter, which we have an article for. https://docs.wpilib.org/en/stable/docs/software/advanced-gradlerio/code-formatting.html

sciencewhiz (commented):
I'm struggling to see why this is useful enough for an frc-docs article. This seems fairly niche to me.

source/docs/software/vscode-overview/organize-imports.rst
WispySparks: In my experience a lot of my imports are put in automatically through this setting and less so through the intellisense. The move for this article came from a discussion on the discord that a majority of people aren't aware of the auto import feature which this article aims to document. It also alleviates concerns about moving classes to different packages and breaking users' imports

source/docs/software/vscode-overview/organize-imports.rst
```diff
@@ -0,0 +1,16 @@
...
+
+.. code-block:: json
+
+    "editor.codeActionsOnSave": {
```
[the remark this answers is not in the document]
WispySparks: I'm not sure as I haven't heard of that option before

Gold856 (changes requested):
A note should be added that syntax errors can cause auto-importing to delete imports that are actually used. Basically, any syntax error in a given scope will cause the extension to not register referenced imports in that scope, so if the only use of an import is in the scope where the syntax error is, that import will be removed. Once the syntax error is fixed, auto-import will fix up everything.

spacey-sooty (changes requested):
If this is common enough to be mentioned in frc-docs then IMO it should instead be added to the templates vscode settings

source/docs/software/vscode-overview/organize-imports.rst
spacey-sooty: Whats the benefit of people knowing this option exists? Why does it justify maintaining a docs article

WispySparks:
> If this is common enough to be mentioned in frc-docs then IMO it should instead be added to the templates vscode settings

I asked about that but it seems like people would rather have it documented and not be the default.

source/docs/software/vscode-overview/organize-imports.rst
WispySparks: I believe it is helpful when coding and helps people from having to type imports manually which is pushback for WPILib to move classes around to different packages. Since it is such a small section I would be more than happy to add it to an existing VSCode article I'm just not sure which one would be best. I don't believe the maintenance would be high as this option has stayed the same the entire time I've been using VSCode (4 years).

Issue #2679, opened 2024-07-29
Remove usage of robotInit


spacey-sooty:

sciencewhiz:
There's 109 instances, not including the ones in RemoteLiteralIncludes

```
Search "robotInit" (109 hits in 25 files of 442 searched)
  C:\Users\chris\Documents\Robotics\frc-docs\source\docs\networking\networking-utilities\portforwarding.rst (4 hits)
	Line 16:       public void robotInit() {
	Line 22:       void Robot::RobotInit {
	Line 42:       public void robotInit() {
	Line 48:       void Robot::RobotInit {
  C:\Users\chris\Documents\Robotics\frc-docs\source\docs\software\advanced-controls\trajectories\trajectory-generation.rst (1 hit)
	Line  70: .. note:: Generating a typical trajectory takes about 10 ms to 25 ms. This isn't long, but it's still highly recommended to generate all trajectories on startup (``robotInit``).
  C:\Users\chris\Documents\Robotics\frc-docs\source\docs\software\basic-programming\alliancecolor.rst (1 hit)
	Line  6: Note that there are three cases: red, blue, and no color yet.  It is important that code handles the third case correctly because the alliance color will not be available until the Driver Station connects.  In particular, code should not assume that the alliance color will be available during constructor methods or `robotInit`, but it should be available by the time `autoInit` or `teleopInit` is called.  FMS will set the alliance color automatically; when not connected to FMS, the alliance color can be set from the Driver Station (see :ref:`"Team Station" on the Operation Tab <docs/software/driverstation/driver-station:Operation Tab>`).
  C:\Users\chris\Documents\Robotics\frc-docs\source\docs\software\basic-programming\java-gc.rst (1 hit)
	Line 145:     public void robotInit() {
  C:\Users\chris\Documents\Robotics\frc-docs\source\docs\software\basic-programming\reading-stacktraces.rst (30 hits)
	Line  45:          Error at frc.robot.Robot.robotInit(Robot.java:24): Unhandled exception: java.lang.NullPointerException
	Line  46:                   at frc.robot.Robot.robotInit(Robot.java:24)
	Line  62:          * ``robotInit`` was the name of the method executing when the error happened.
	Line  64:       * ``robotInit`` is a function in the ``frc.robot.Robot`` package (AKA, your team's code)
	Line  66:       * ``robotInit`` was called from a number of functions from the ``edu.wpi.first.wpilibj`` package (AKA, the WPILib libraries)
[...]

PR #2689, opened 2024-08-04
Remove usage of robotInit from FRC Docs

DJB755:
Remove usage of robotInit #2679


DJB755:
Oop realized I did not go throught the examples linked in the docs, will do that before submitting the PR

DJB755:
Remove usage of robotInit https://github.com/wpilibsuite/frc-docs/issues/2679 replaced with different constructor on the FRCDocs side, should I fork allwpilib too and change the examples there too before?

source/docs/software/advanced-controls/trajectories/trajectory-generation.rst
```diff
@@ -67,7 +67,7 @@ Here is an example of generating a trajectory using clamped cubic splines for th
 
 .. note:: The Java code utilizes the `Units <https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/math/util/Units.html>`_ utility, for easy unit conversions.
 
-.. note:: Generating a typical trajectory takes about 10 ms to 25 ms. This isn't long, but it's still highly recommended to generate all trajectories on startup (``robotInit``).
+.. note:: Generating a typical trajectory takes about 10 ms to 25 ms. This isn't long, but it's still highly recommended to generate all trajectories on startup (``Robot``).
```
sciencewhiz, line 70: This should say something like ``Robot`` constructor

source/docs/software/basic-programming/alliancecolor.rst
```diff
@@ -3,7 +3,7 @@ Get Alliance Color
...
+Note that there are three cases: red, blue, and no color yet.  It is important that code handles the third case correctly because the alliance color will not be available until the Driver Station connects.  In particular, code should not assume that the alliance color will be available during constructor methods or `Robot`, but it should be available by the time `autoInit` or `teleopInit` is called.  FMS will set the alliance color automatically; when not connected to FMS, the alliance color can be set from the Driver Station (see :ref:`"Team Station" on the Operation Tab <docs/software/driverstation/driver-station:Operation Tab>`).
```
sciencewhiz, line 6: This shouldn't be changed, as now it's redundant

source/docs/software/basic-programming/reading-stacktraces.rst
```diff
@@ -42,8 +42,8 @@ To start, search above the ``unexpected error has occurred`` for the stack trace
 
       .. code-block:: text
 
-         Error at frc.robot.Robot.robotInit(Robot.java:24): Unhandled exception: java.lang.NullPointerException
-                  at frc.robot.Robot.robotInit(Robot.java:24)
+         Error at frc.robot.Robot.Robot(Robot.java:24): Unhandled exception: java.lang.NullPointerException
```
sciencewhiz, line 45: I suspect that this is not what the stack trace would look like

sciencewhiz (changes requested):
In general, this needs more nuanced changes. I commented on a few places at the beginning, but these "rules" apply many more places.

- 
- Can't say Robot method or function, need to update those to say constructor
- When there's a description of a side effect that the user needs to be aware of, it's still appropriate to reference robotInit, as users can still use robotInit, we just don't want to encourage it's use with examples.
- There's cases where code samples say `void Robot()` these need to be updated since Robot isn't a method
- I suspect that the python changes aren't correct
- For the stacktraces article, I suspect the stack traces look significantly different when run from the constructor. This can't be updated just with find and replace. Per the previous point, I don't think this article is important to update as showing robotInit is ancillary to the point of explaining stack traces, but it would be bad to update the way it is that makes it wrong.

spacey-sooty:
Resolves https://github.com/wpilibsuite/frc-docs/issues/2679

spacey-sooty:
> Remove usage of robotInit #2679 replaced with different constructor on the FRCDocs side, should I fork allwpilib too and change the examples there too before?

allwpilib has already removed this https://github.com/wpilibsuite/allwpilib/pull/6623

sciencewhiz:
Superseded by #2841

PR #2696, opened 2024-08-14
Update API links to use development

sciencewhiz:


sciencewhiz:
Linkcheck will fail until wpilibsuite/allwpilib#6963 is merged

PR #2699, opened 2024-08-16
Clarify python version information

virtuald:
The RoboRIO python version is just extra information and doesn't really belong here. It confused a user on CD, remove it.


TheTripleV:
Maybe just say 3.12 and let it be a happy accident from the user's perspective when an older version works?

sciencewhiz:
The CD user drew the wrong conclusion, but I'm not sure that not having that info would have helped them draw the right conclusion.

sciencewhiz:
The build failures will be fixed with a merge of main

PR #2690, opened 2024-08-08
Convert links to ReDown

TheTripleV:

PR #2702, opened 2024-08-17
Fix link syntax that was broken in #2690

sciencewhiz:


TheTripleV:
`:ref:[CTRE Power Distribution Panel](docs/software/can-devices/power-distribution-module:Power Distribution Module)` is supposed to work. I'll take a look at the generation

sciencewhiz:
It does work, but the only ones that were switched were the ones that were on the same line as an external link. This makes them all consistent again, so we can switch them all later.

TheTripleV (dismissed):

jasondaming:
I think we can do this if you resolve conflicts

sciencewhiz:
I found more categories of issues I'm working on

sciencewhiz:
Supersedes #2701

PR #2703, opened 2024-08-18
Remove Axon and GRIP docs

sciencewhiz:
They've been hidden for 2 and 1 years respectively, and are unlikely to be revived. They are still availible in the respective year's tags/branches if needed in the future.


jasondaming (changes requested):
Do you think we could remove the references to GRIP [here](https://docs.wpilib.org/en/latest/docs/software/networktables/reading-array-values-published-by-networktables.html)

jasondaming (approved):

PR #2711, opened 2024-08-25
Add page for persistent alerts

jwbonner:
Documentation for the new alerts [API](https://github.com/wpilibsuite/allwpilib/pull/6735) and [widget](https://github.com/wpilibsuite/shuffleboard/pull/805).


source/docs/software/telemetry/persistent-alerts.rst
```diff
@@ -0,0 +1,55 @@
+# Recording Faults with Persistent Alerts
+
+Robots encounter a variety of fault conditions: disconnected sensors/motors, invalid mechanism states, initialization failures, etc. While the FRC Driver Station provides a :ref:`console interface <docs/software/driverstation/driver-station:Messages Tab>` for instantaneous alerts, the risk of missing important messages makes it poorly suited to communicate faults that persistent over time.
```
jasondaming, line 3:
```suggestion
Robots encounter a variety of fault conditions: disconnected sensors/motors, invalid mechanism states, initialization failures, etc. While the FRC Driver Station provides a :ref:`console interface <docs/software/driverstation/driver-station:Messages Tab>` for instantaneous alerts, the risk of missing important messages makes it poorly suited to communicate faults that are persistent over time.
```

Could also be "faults that persist"

source/docs/software/telemetry/persistent-alerts.rst
```diff
@@ -0,0 +1,55 @@
...
+
+Instead, WPILib provides an API for managing persistent alerts published via Network Tables. Alerts are assigned a priority (*error*, *warning*, or *info*) and can be *activated* or *deactivated* in robot code. The set of active alerts can be displayed on a dashboard, as shown below in Shuffleboard.
```
jasondaming, line 5: I think this would be a good place to link to the API docs for this.  Especially as we later reference them as the ideal way to learn about the alert levels

source/docs/software/telemetry/persistent-alerts.rst
```diff
@@ -0,0 +1,55 @@
...
+
+.. image:: images/alerts.png
+   :alt: A screenshot of the alerts widget in Shuffleboard, with several active alerts.
+
+Active alerts are automatically displayed in order based on priority and how recently they were activated, with newer and more critical alerts at the top of the list. This provides a mechanism for drivers (on the field) and programmers (off the field) to quickly assess the state of the robot and determine whether it is in a "match-ready" state. Highly descriptive alerts also allow drive teams to adapt to faults in real time, such as switching strategies in case of a subsystem failure.
+
+## API Usage
+
+Alerts are created and managed using the ``Alert`` class. Typically, an alert should be instantiated once and stored as a class member. Periodically, the ``set`` method should be called to update the state of the alert based on relevant data.
```
jasondaming, line 14: What is the thinking behind this periodic suggestion?  What things do we expect teams to update?

Seems to me most alerts would be activate when it starts and deactivate when it stops and I don't see the use case for the periodic updates in between.  Is there some sort of timeout or something on the alert that I missed?  Wouldn't this just create a swapping effect of the alerts in the widget not really related to when they actually started.

jasondaming (changes requested):

source/docs/software/telemetry/persistent-alerts.rst
jasondaming: I guess my suggestion here might be to phrase this in a way that encourages the user to periodically use the set for simplicity sake.

source/docs/software/telemetry/persistent-alerts.rst
jasondaming: See [here](https://docs.wpilib.org/en/latest/docs/software/advanced-controls/controllers/pidcontroller.html) for an example of linking to the API.

source/docs/software/telemetry/persistent-alerts.rst
```diff
@@ -0,0 +1,57 @@
...
+
+Active alerts are automatically displayed in order based on priority and how recently they were activated, with newer and more critical alerts at the top of the list. This provides a mechanism for drivers (on the field) and programmers (off the field) to quickly assess the state of the robot and determine whether it is in a "match-ready" state. Highly descriptive alerts also allow drive teams to adapt to faults in real time, such as switching strategies in case of a subsystem failure.
+
+## API Usage
+
+Alerts are created and managed using the ``Alert`` class. Typically, an alert should be instantiated once and stored as a class member. The ``set`` method is used to activate or deactivate the alert. Change detection is built-in, so calling ``set`` multiple times with the same value does affect the order in which alerts are displayed.
+
+Alert states are often expressed most easily as a conditional, such as whether the latest signal from a CAN device is up-to-date. In this case, we recommend calling the ``set`` method in a periodic loop and relying on the built-in change detection to active the alert.
```
jasondaming, line 16:
```suggestion
Alert states are often expressed most easily as a conditional, such as whether the latest signal from a CAN device is up-to-date. In this case, we recommend calling the ``set`` method in a periodic loop and relying on the built-in change detection to activate the alert.
```

jasondaming:
I think the link is failing because if you want to use the "new" syntax I think you might need the full link.  So either revert to the old style or change it to the full link and see if that works.  See this s[ection](https://docs.wpilib.org/en/latest/docs/contributing/frc-docs/style-guide.html#python-api-links)

jwbonner:
It looks like CI is failing because of the missing Python class. Should we update the bindings before merging this? (I assume docs are based on the dev builds of robotpy).

jwbonner:
I switched to the full link. Even with the old syntax it fails because the Python class doesn't exist yet.

jasondaming:
Yeah makes sense I guess we just wait then till that happens.  This looks ready to go otherwise to me.

jasondaming (dismissed):

sciencewhiz:
I think I'd prefer to merge without a python link, and open an issue to have it added later when available.

source/docs/software/telemetry/persistent-alerts.rst
```diff
@@ -0,0 +1,57 @@
...
+
+.. image:: images/alerts.png
+   :alt: A screenshot of the alerts widget in Shuffleboard, with several active alerts.
+
+Active alerts are automatically displayed in order based on priority and how recently they were activated, with newer and more critical alerts at the top of the list. This provides a mechanism for drivers (on the field) and programmers (off the field) to quickly assess the state of the robot and determine whether it is in a "match-ready" state. Highly descriptive alerts also allow drive teams to adapt to faults in real time, such as switching strategies in case of a subsystem failure.
+
+## API Usage
+
+Alerts are created and managed using the ``Alert`` class. Typically, an alert should be instantiated once and stored as a class member. The ``set`` method is used to activate or deactivate the alert. Change detection is built-in, so calling ``set`` multiple times with the same value does affect the order in which alerts are displayed.
```
sciencewhiz, line 14: Nowhere is it mentioned that set takes a boolean, Especially without any code in the examples, I think that's important.

sciencewhiz (approved):

PR #2713, opened 2024-09-01
Compress PNG files with Zopfli

Gold856:
Also corrected the file extension on a few images. Ran [oxipng](https://github.com/shssoichiro/oxipng) using the zopfli compressor for maximal compression. About 15 MiB was saved (~25% reduction from 90 MB in the installer.)


sciencewhiz:
The LabVIEW block diagrams should not be stripped because they contain code. https://knowledge.ni.com/KnowledgeArticleDetails?id=kA03q000000x2PkCAI&l=en-US

Gold856:
I finished rerunning oxipng without stripping any metadata. I will say that it appears only [this one image](https://github.com/wpilibsuite/frc-docs/blob/main/source/docs/software/labview/resources/images/local-and-global-variables-in-labview-for-frc/6.png) has any LabVIEW metadata. I checked basically every image under source/docs/software/labview/resources/images and found no LabVIEW metadata.

sciencewhiz (approved):

PR #2718, opened 2024-09-01
Update list of supported OSes for 2025

sciencewhiz:


PeterJohnson:
We should also add a warning regarding Windows 10 EOL coming up in Oct 2025.  We plan to keep supporting it in 2026, but may drop support in 2027.

PeterJohnson (approved):

PR #2727, opened 2024-09-06
Create Python Script + CI to check for language coverage in code blocks

falOn-Dev:
I wanted to build a quick tool that would allow people to view what percentage of code blocks contain which languages, as well as letting people know which blocks were missing languages.

This is a python script that I also added to run with the `CI.yml` workflow, although I'm not married to it running during CI, and would be fine if it was just a tool that can be ran.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -186,3 +186,18 @@ jobs:
     - name: Format
       run: |
         black --check .
+
+  check-codeblock-coverage:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: 3.9
+      - name: Check Coverage
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/CI.yml
```diff
@@ -186,3 +186,18 @@ jobs:
     - name: Format
       run: |
         black --check .
+
+  check-codeblock-coverage:
+    runs-on: ubuntu-latest
```
spacey-sooty, line 191: You should pin the version

spacey-sooty (changes requested):

.github/workflows/CI.yml
falOn-Dev: I might be missing something but none of the other jobs specify an ubuntu version?

falOn-Dev:
I have discovered some weird issues with this that I can't get working, so I am going to close this for the time being, as I think putting it in, in a semi-functional state would be a detriment.

PR #2731, opened 2024-09-07
Script for checking codeblock coverage

falOn-Dev:
I wrote a simple tool for getting data on how many codeblocks cover each language. By default it will check for Python, C++, and Java, but the RegEx is generated at runtime based on what languages you pass in, so it still works for checking the coverage of other languages.

This script contains its own documentation on how to run it, but I'm happy to add a page to the frc-docs contributing guide if needed.

(I made a similar PR earlier but I wanted to make a new one as that one contains differing information from this newer version of the script)


spacey-sooty (changes requested):
image
Uhhh?

scripts/check_codeblock_coverage.py
```diff
@@ -0,0 +1,210 @@
+"""
+This script checks the code block coverage in .rst files in a specified directory.
+It counts the number of code blocks for each language and outputs the coverage percentage for each language.
+If the verbose flag is set, it also outputs detailed information about missing code blocks.
+
+Arguments for the script:
+--dir: The directory to search for .rst files (required)
+--verbose: Include detailed information about missing code blocks, as well as output to file rather than terminal (optional, default=False)
+--output: The path to the output file (optional, default="output.txt")
+--langs: The languages to check for (optional, default=["java", "python", "c++"])
```
spacey-sooty, line 10: --help should output this or something closer to it

spacey-sooty (changes requested):
Could you add an option to link to the non covered parts?

falOn-Dev:
> Could you add an option to link to the non covered parts?

That's what --verbose does, unless I'm misunderstanding?

spacey-sooty:
Should check for when no arguments are provided and display help

spacey-sooty:
> > Could you add an option to link to the non covered parts?
> 
> That's what --verbose does, unless I'm misunderstanding?

Verbose just breaks it for me
image

falOn-Dev:
Ah, it's working it just doesn't say anything, okay I'll fix both of those things

TheTripleV:
I'm concerned about this script.

1. It might be better as a one off tracking issue listing everything missing
2. As there are unique features for each language (coroutines in progress for python and java, epilogue for java), not all code blocks should contain all 3 languages.
3. If this ends up being a ci check, I don't want prs to fail because the author doesn't know cpp. Separate prs for each lang are ok.
3. I'd be more comfortable if it was implemented with Sphinx's api instead of regex.

falOn-Dev:
Implemented both of @spacey-sooty requests.

falOn-Dev:
> I'm concerned about this script.
> 
>     1. It might be better as a one off tracking issue listing everything missing
> 
>     2. As there are unique features for each language (coroutines in progress for python and java, epilogue for java), not all code blocks should contain all 3 languages.
> 
>     3. If this ends up being a ci check, I don't want prs to fail because the author doesn't know cpp. Separate prs for each lang are ok.
> 
>     4. I'd be more comfortable if it was implemented with Sphinx's api instead of regex.

I do not think this should be a CI thing, or at least it shouldn't fail PRs. It's intended to be a tool for people to just check if there are any glaring gaps in language support. I can understand wanting it in Sphinx's API but I have never used that, so I used regex, if it's absolutely necessary to not be regex that's fine

spacey-sooty:
> 3\. If this ends up being a ci check, I don't want prs to fail because the author doesn't know cpp. Separate prs for each lang are ok.

I feel like this is a recipe to have documentation fall further behind for Python and make C++ start falling behind substantially

falOn-Dev:
I feel like if the tool has proper warnings (which I'm happy to add more of) then not putting it in over the _possibilities_ of misuse would be a bad decision, because it would still be useful

PR #2745, opened 2024-09-15
change report issue to report docs issue

spacey-sooty:
This could easily be confused by people as the place to report bugs in WPILib and since the majority of WPILib exists in allwpilib it makes more sense to point it here.

### DIFF
--- a/source/index.rst
+++ b/source/index.rst
@@ -445,6 +445,6 @@ Community translations can be found in a variety of languages in the bottom-left
    :caption: Issues
    :hidden:
 
-   Report an Issue <https://github.com/wpilibsuite/allwpilib/issues>
+   Report a Documentation Issue <https://github.com/wpilibsuite/frc-docs/issues>
 
 .. todolist::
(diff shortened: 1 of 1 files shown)


jasondaming:
I disagree with this change.  There is no reason to think that reporting an issue on the docs page would bring you to allwpilib instead of docs.

This is evidenced by the very rare issue that is mistakenly created on docs that is an allwpilib issue.

sciencewhiz:
Agree. There is space, so for clarity the title could change to report a docs issue

source/index.rst
```diff
@@ -445,6 +445,6 @@ Community translations can be found in a variety of languages in the bottom-left
    :caption: Issues
    :hidden:
 
-   Report an Issue <https://github.com/wpilibsuite/frc-docs/issues>
+   Report an Issue <https://github.com/wpilibsuite/allwpilib/issues>
```
spacey-sooty, line 448:
```suggestion
   Report a Documentation Issue <https://github.com/wpilibsuite/frc-docs/issues>
```

sciencewhiz (approved):

PR #2752, opened 2024-09-19
Document TimesliceRobot template

### LINKED ISSUE #1446: Document `TimesliceRobot` template
2022 added a `TimesliceRobot` template. This should get documented alongside the other templates

leongmichael:
resolve #1446

### DIFF
--- a/source/docs/software/vscode-overview/creating-robot-program.rst
+++ b/source/docs/software/vscode-overview/creating-robot-program.rst
@@ -4,7 +4,7 @@ Once everything is installed, we're ready to create a robot program.  WPILib com
 
 ## Choosing a Base Class
 
-To start a project using one of the WPILib robot program templates, users must first choose a base class for their robot.  Users subclass these base classes to create their primary :code:`Robot` class, which controls the main flow of the robot program.  There are three choices available for the base class:
+To start a project using one of the WPILib robot program templates, users must first choose a base class for their robot.  Users subclass these base classes to create their primary :code:`Robot` class, which controls the main flow of the robot program.  There are various choices available for the base class:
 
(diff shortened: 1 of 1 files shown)


leongmichael:
i realize there are various other examples not documented. would it be best to merge this first or do them all in this one PR?

jasondaming:
Looks like you have an annoying extra whitespace linting error probably just a space at the end of a line or on a blank line.

I would say it is up to you on the organization of this vs multiple PRs.  I think the only two template types that aren't documented are "skeletons" and "educational".  Skeletons could be generically mentioned in one paragraph towards the beginning.  Educational has a very narrow specific use case that I'm not sure it would be worth documenting here.

If you are looking for additional ideas for improving this document the 2 highest on my list would be:
1) Advice up front about making (or not making) a selection.  This is always difficult to do impartially but is often the most critical piece for helping teams.  In this case what I would stress is that starting from a template is IMO a really bad idea for a new user.  It leaves them with a lot of blank page fears and gives them little (other than the template comments) to guide them on the proper structure.  Instead new teams should be looking at an understanding the example most relevant to them and/or past team code with goals similar to theirs.  I think it would also be good to link to ["What is Command Based"](https://frcdocs.wpi.edu/en/latest/docs/software/commandbased/what-is-command-based.html) instead of referencing a tutuorial that no longer exists and linking to the index page.
2) Adding Python.  It has been a slow process for us to get Python added everywhere but we should do our best to link to it also in all situations where it has the same features and Java/C++ are explicitly linked.

sciencewhiz (changes requested):

sciencewhiz (approved):

PR #2760, opened 2024-09-24
Add XRP to creating robot programs

leongmichael:


source/docs/software/vscode-overview/creating-robot-program.rst
```diff
@@ -86,6 +86,22 @@ The ``Romi - Command Bot`` template provides a ``RomiDrivetrain`` subsystem that
 
 This subsystem also provides functions for retrieving and resetting the Romi's onboard encoders.
 
+### XRP
+
+Teams using a :ref:`XRP <docs/xrp-robot/index:Getting Started with XRP>` should use the ``XRP - Timed`` or ``XRP - Command Bot`` template.
+
+#### XRP - Timed
+
+The ``XRP - Timed`` template provides a ``XRPDrivetrain`` class that exposes an ``arcadeDrive(double xaxisSpeed, double zaxisRotate)`` method. It's up to the user to feed this arcadeDrive function.
```
jasondaming, line 95: We should be exact with the naming here.  It looks like they are called ``XRP - Command Robot`` and ``XRP - Timed Robot``.  This is also a problem with the Romi section.  I realize you didn't write that but I think improving it the same way would help!

source/docs/software/vscode-overview/creating-robot-program.rst
```diff
@@ -86,6 +86,22 @@ The ``Romi - Command Bot`` template provides a ``RomiDrivetrain`` subsystem that
...
+
+This class also provides functions for retrieving and resetting the XRP's onboard encoders.
+
+#### XRP - Command Bot
+
+The ``XRP - Command Bot`` template provides a ``XRPDrivetrain`` subsystem that exposes an ``arcadeDrive(double xaxisSpeed, double zaxisRotate)`` method. It's up to the user to feed this arcadeDrive function.
+
+This subsystem also provides functions for retrieving and resetting the XRP's onboard encoders.
```
jasondaming, line 103: All of this feels just a duplication of the Romi stuff.  While it is all true I am wondering if we couldn't get across all the same information with a single combined section that use this text once for all.  It would need to say "subsystem/class".

jasondaming (changes requested):

PR #2763, opened 2024-10-02
Update supported OS versions

rzblue:
WPILib requires macOS >13.3
Also updated robotpy requirements- I assumed they (mostly) matched wpilib.


virtuald:
Yes, they match.

PeterJohnson (approved):

PR #2684, opened 2024-08-01
Rewrite command controller articles to show composition

Gold856:
Control subsystems are full of footguns and are poor abstractions since they don't save any typing, yet they actively bypass the command mutex system. Control commands are also poor abstractions as they are easy to mess up, yet hide internal details away from the user, making debugging harder. They also don't save any typing.

PR #2771, opened 2024-10-07
Add more details to control in Command-based articles

Gold856:
Extension of #2684


jasondaming:
This details in depth much more what is happening.  The issue I have is that I am unsure what the user is learning from reading this.  They either have the skills to be able to look at the code and see the same things for themselves or they don't in which case I don't think they can read or make sense of these paragraphs.  It also doesn't describe any of the "why" these things are being done.

PR #2774, opened 2024-10-08
Add LED patterns to LED documentation

### LINKED ISSUE #2773: Addressable LED Pattern documentation
WPILIb added an API for animating LED patterns https://github.com/wpilibsuite/allwpilib/pull/6344 The addressable LED article should be updated for this. The changes to the AddressableLED example mean the RLIs need extensive rework.

SamCarlberg:
Resolves #2773


source/docs/software/hardware-apis/misc/addressable-leds.rst
```diff
@@ -19,29 +19,74 @@ After the length of the strip has been set, you'll have to create an ``Addressab
    .. tab-item:: Java
       :sync: Java
 
-      .. remoteliteralinclude:: https://raw.githubusercontent.com/wpilibsuite/allwpilib/v2024.3.2/wpilibjExamples/src/main/java/edu/wpi/first/wpilibj/examples/addressableled/Robot.java
+      .. remoteliteralinclude:: https://raw.githubusercontent.com/wpilibsuite/allwpilib/main/wpilibjExamples/src/main/java/edu/wpi/first/wpilibj/examples/addressableled/Robot.java
```
sciencewhiz, line 22: These should use the 2025 alpha-2 tag, assuming that's recent enough, and if not, a full commit hash, rather then main.

source/docs/software/hardware-apis/misc/addressable-leds.rst
```diff
@@ -107,50 +186,49 @@ LEDs can be set with the ``setHSV`` method that takes 4 arguments: index of the
...
       :sync: C++
 
-      .. remoteliteralinclude:: https://raw.githubusercontent.com/wpilibsuite/allwpilib/v2024.3.2/wpilibcExamples/src/main/cpp/examples/AddressableLED/cpp/Robot.cpp
+      .. remoteliteralinclude:: https://raw.githubusercontent.com/wpilibsuite/allwpilib/main/wpilibcExamples/src/main/cpp/examples/AddressableLED/cpp/Robot.cpp
          :language: c++
          :lines: 15-20
          :linenos:
          :lineno-start: 15
+
+.. image:: images/rainbow.gif
```
sciencewhiz, line 233: I wonder if specifying the :width: will fix the pdf error

source/docs/software/hardware-apis/misc/addressable-leds.rst
SamCarlberg: I went with `main` mostly to be sure it had the newest features, and to make it easy to find and replace later before merge

source/docs/software/hardware-apis/misc/addressable-leds.rst
SamCarlberg: 🤔 Hmm. I'll need to test that. I don't have LaTeX tools installed locally so it'll need to be on CI, which could be frustrating.

The style guide _does_ explicitly call out gifs as being *un*acceptable for storage and accessibility concerns, but I don't see a way around it since the whole point of these docs are to demonstrate animations

source/docs/software/hardware-apis/misc/addressable-leds.rst
sciencewhiz: inspector automates updating from commit hashes, but not main. But that's OBE now that beta 2 is out

sciencewhiz:
this should remove the ignore from inspector.json

sciencewhiz:
Setting the width to 900 makes mobile scroll

sciencewhiz:
If you put in a static png with the same name as the video files, I think that will fix the pdf build

source/docs/software/hardware-apis/misc/addressable-leds.rst
```diff
@@ -19,29 +19,74 @@ After the length of the strip has been set, you'll have to create an ``Addressab
...
          :lineno-start: 11
 
-      .. remoteliteralinclude:: https://raw.githubusercontent.com/wpilibsuite/allwpilib/v2024.3.2/wpilibcExamples/src/main/cpp/examples/AddressableLED/cpp/Robot.cpp
+      .. remoteliteralinclude:: https://raw.githubusercontent.com/wpilibsuite/allwpilib/v2025.0.0-alpha-2/wpilibcExamples/src/main/cpp/examples/AddressableLED/cpp/Robot.cpp
          :language: c++
          :lines: 7-13
          :linenos:
          :lineno-start: 7
 
 .. note:: The roboRIO only supports only 1 ``AddressableLED`` object. As WS2812B LEDs are connected in series, you can drive several strips connected in series from from ``AddressableLED`` object.
```
sciencewhiz, line 43: I think this not can be deleted since it's said in the new section

SamCarlberg:
> If you put in a static png with the same name as the video files, I think that will fix the pdf build

As in, `images/scroll-relative.mp4` should have a corresponding `images/scroll-relative.png` file?

SamCarlberg:
`only` or `ifconfig` may also work, to conditionally render a video element or an image element based on the build target

source/docs/software/hardware-apis/misc/addressable-leds.rst
```diff
@@ -107,50 +187,678 @@ LEDs can be set with the ``setHSV`` method that takes 4 arguments: index of the
...
+
+      // Write the data to the LED strip
+      m_led.SetData(m_ledBuffer);
+      ```
+
+### Modifying effects
+
+.. note:: The built in animating effects like blinking and scrolling are based on the time returned by ``WPIUtilJNI.now()`` - in effect, they will play as if they started when the robot booted. Because all built in animation patterns are periodic, this means that the *first* period of a pattern may be truncated at any arbitrary point between 0% and 100%, and every period after that will play normally.
+
+.. note:: Animating effects use the :ref:`Java units library <docs/software/basic-programming/java-units:The Java Units Library>` and the :ref:`C++ units library <docs/software/basic-programming/cpp-units:The C++ Units Library>` for speeds and durations.
```
sciencewhiz, line 465: This note feels late, since the rainbow already used units for distance

source/docs/software/hardware-apis/misc/addressable-leds.rst
```diff
@@ -107,50 +187,678 @@ LEDs can be set with the ``setHSV`` method that takes 4 arguments: index of the
...
+   :poster: ../../../../_images/discontinuous-gradient.png
+   :width: 900
+
+Blinking can be done in one of three ways:
+
+1. Symmetrically, where an equal amount of time is spent in the "on" and "off" states per cycle
+2. Asymetrically, where the time spent "on" can be configured independently from the time spent "off"
+3. Synchronously, where the time spent on and off is synchronized with an external source (for example, the state of the RSL)
+
+.. note:: The time-based blinking modifiers are synced with the robot's clock
```
sciencewhiz, line 665: At first I wasn't sure what this meant. After reading the note in modifying effects, it makes more sense, but since that note comes later, maybe more of it should be up here.

source/docs/software/hardware-apis/misc/addressable-leds.rst
```diff
@@ -107,50 +187,678 @@ LEDs can be set with the ``setHSV`` method that takes 4 arguments: index of the
...
+   :loop:
+   :poster: ../../../../_images/discontinuous-gradient.png
+   :width: 900
+
+.. video:: images/scroll-absolute.mp4
+   :loop:
+   :poster: ../../../../_images/discontinuous-gradient.png
+   :width: 900
+
+Scrolling can be controlled in two different ways: either at a speed as a function of the length of the buffer or view to which it is applied (i.e., the scrolling speed is in terms of percentage per second, or a similar unit), or as a function of the density of the phsyical LED strips (i.e. scrolling speed is in meters per second, or a similar unit). Relative velocities are particularly useful when a scrolling pattern is applied to different LED strips with different LED spacing (such as one strip with 120 LEDs per meter daisy chained to a second strip with 60 or 144 LEDs per meter), when prototyping before having a particular LED strip in mind (where the density isn't yet known), or when LED strips are quickly changed out. Scrolling at a fixed real-world speed (eg ``InchesPerSecond.of(2)``) may be more understandable to readers, but will move faster or slower when applied to an LED strip with a lower or higher pixel density, respectively.
```
sciencewhiz, line 565:
```suggestion
Scrolling can be controlled in two different ways: either at a speed as a function of the length of the buffer or view to which it is applied (i.e., the scrolling speed is in terms of percentage per second, or a similar unit), or as a function of the density of the physical LED strips (i.e. scrolling speed is in meters per second, or a similar unit). Relative velocities are particularly useful when a scrolling pattern is applied to different LED strips with different LED spacing (such as one strip with 120 LEDs per meter daisy chained to a second strip with 60 or 144 LEDs per meter), when prototyping before having a particular LED strip in mind (where the density isn't yet known), or when LED strips are quickly changed out. Scrolling at a fixed real-world speed (eg ``InchesPerSecond.of(2)``) may be more understandable to readers, but will move faster or slower when applied to an LED strip with a lower or higher pixel density, respectively.
```

sciencewhiz:
> > If you put in a static png with the same name as the video files, I think that will fix the pdf build
> 
> As in, `images/scroll-relative.mp4` should have a corresponding `images/scroll-relative.png` file?

yes

sciencewhiz:
> > > If you put in a static png with the same name as the video files, I think that will fix the pdf build
> > 
> > 
> > As in, `images/scroll-relative.mp4` should have a corresponding `images/scroll-relative.png` file?
> 
> yes. That will work for image files, but I'm not sure if sphinx will automatically fall back from a video file to an image file

This won't work, looks like ..only would probably be best.

sciencewhiz:
On chrome and edge, the videos don't show a play. Maybe because they are not tall enough? Works ok on Firefox

image

SamCarlberg:
I'm not sure, both firefox and chrome showed the controls overlaid on the videos in my testing. Annoyingly, the controls obscure the videos, and there doesn't seem to be any options for moving the (eg to be below the video). Maybe we could have some custom JS to make the controls appear on hover? https://stackoverflow.com/q/40284816

TheTripleV:
The video controls are there. The slider is the progress slider. If you click on the very very top few pixels of the video and press space, it will start to play and the controls will fade away.

Since this is looping, I would rather put in gifs for the website and png for the pdf.
APNG is support by all browsers https://caniuse.com/apng so maybe if you put in an apng, it will work on both web and pdf

SamCarlberg:
> Since this is looping, I would rather put in gifs for the website and png for the pdf.

That was my preference too, but Daltz says it's an accessibility concern to have autoplaying gifs

TheTripleV:
I guess the fix then is making the videos a lot taller so the controls render properly

sciencewhiz:
> I guess the fix then is making the videos a lot taller so the controls render properly

We might be able to do something like this, rather then make new videos: https://stackoverflow.com/a/65116261

zeroClearAmerican:
A bit of feedback from our beta testing: LEDPattern examples sometimes refer to calling a function using the Units library, like `scrollAtRelativeSpeed(Percent.per(Second).of(25))`, however they don't specify that the Units library should be used. To a new student, this could be confusing because VSCode would report that an import for Percent would be missing when typed like the example.

Updating the example to include the Units reference before `Percent`, like `scrollAtRelativeSpeed(Units.Percent.per(Second).of(25))` allows intellisense to understand and suggest the correct import and the user understands just by reading the example that `Percent` is part of the Units library.

sciencewhiz:
this has conflicts now since some of the css was merged in the a-stop PR

sciencewhiz:
\inspector fix all

sciencewhiz:
There's a conflict with inspector.json

sciencewhiz (commented):
It's not obvious that these are videos. Mousing over them gives the same hand that the static images give, which in a static image opens the static image. I think if you add `:muted:` and `:autoplay:` they'll autoplay, and then clicking on them will stop them (see the DS A-Stop). I think that would be preferred.

Both firefox and chrome seem to render the poster different sizes from each other, and also different sizes from the video (so that when you click the video to start it, it resizes). I think it's probably ok to not use a poster.

sciencewhiz:
\inspector fix all

sciencewhiz (approved):

PR #2775, opened 2024-10-09
Reenable delta

sciencewhiz:
Install sphinx-build-compatibility plugin


sciencewhiz:
The compatibility plugin doesn't work except in RTD environments

Issue #989, opened 2020-12-09
Cleanup Root TOC


Daltz333:
- Condense Contributing
- Condense Networking?
- Split up Dashboards and Condense

jasondaming:
This also ties in to #879 
I thought about this some more and here are some of my suggestions:
- Reduce the Size of the logo.  My proposal would be the current height of WPI & FRC part to now also include the WPILib gear
- Remove Contributing section.  Either place as a single bullet under issues or just keep the documentation on latest.  
- Condense Software Tools to: Driver Station, Dashboards, and Misc
- Move Old Command Based under Command Based
- Remove Convenience Features and place the only article, Scheduling Functions under roboRIO or something
- Place all of networking in one section under Hardware or under the RoboRIO section
- Instead of "What is WPILib" make the section "WPILib" and move the stuff from 2021 Overview and support resources under it
- Make the current "Basic Programming" section called Misc Programming, it is wildly misleading as is.
- Make a new section called "Common Robot Problems" or something similar.  This will both be for things that aren't quite big enough for a tutorial like: Joysticks, Autonomous, Unit Testing, etc .  The new FAQ.  Also a few of the articles that don't currently fit in their spot like: Addressable LEDs and Using WPILib Classes to Drive your Robot under actuators.

Daltz333:
> Reduce the Size of the logo. My proposal would be the current height of WPI & FRC part to now also include the WPILib gear
Is there a particular reason for this? I don't think this is necessary.
> Remove Contributing section. Either place as a single bullet under issues or just keep the documentation on latest.
This would only save 1 level of space, and make it less contributor friendly in the future. Also we can't keep it on latest, as latest is (master) and stable is (master) during build season. Beta branches have been decided against in the past.
> Condense Software Tools to: Driver Station, Dashboards, and Misc
I agree with this
> Move Old Command Based under Command Based
I disagree with this, this makes it more confusing. We also plan to deprecate old-commandbased in the future so this isn't the route forward.
> Place all of networking in one section under Hardware or under the RoboRIO section
I also disagree with this. While networking is hardware, this section doesn't really feature the hardware side as much as just general networking
> Instead of "What is WPILib" make the section "WPILib" and move the stuff from 2021 Overview and support resources under it
[...]

PR #2804, opened 2024-10-16
Change Programming Basics to Miscellaneous Programming

spacey-sooty:
See https://github.com/wpilibsuite/frc-docs/issues/989


sciencewhiz:
WPI has started an effort to comprehensively reorganize the docs, so I don't think we'd make organization changes while that is in progress. @arbessette

arbessette:
Yes, I recommend we not make any organizational changes.

PR #2808, opened 2024-10-18
Remove warning about non-default PowerDistribution ID

Gold856:
My team ran a REV PDH with a non-default ID and the DS logs still had voltage and current data.


sciencewhiz:
Did you also verify with a PDP? The note covers both

Gold856:
Yes. I can see voltage and current data for both.

sciencewhiz (approved):

Issue #2812, opened 2024-10-20
Document NT protobuf/struct support


sciencewhiz:
https://docs.wpilib.org/en/latest/docs/software/telemetry/telemetry.html

> While NetworkTables does not yet support serialization of complex data types (this is tentatively scheduled for 2024), mutable types from user code can be easily extended to interface directly with WPILib dashboards via the Sendable interface, whose usage is described in the next article.

PR #2814, opened 2024-10-21
Document protobufs and structs

spacey-sooty:
Resolves https://github.com/wpilibsuite/frc-docs/issues/2812


PeterJohnson:
As currently written, this doesn't actually document how to use them or create your own.  I would expect to see documentation of the pros/cons of each of these approaches (sendable, protobuf, struct), as well as code examples of how to use each, and code examples of how to create custom ones.

PR #2816, opened 2024-10-28
Document AStop

Kevin-OConnor:
Add documentation for new DS AStop to DS article and beta testing task


Kevin-OConnor:
Added the imgconverter extension to make sure it works properly in CI, but if we go this direction I should also add the ImageMagick dependency to the Build Instructions page before merging. With it being an easily available open source tool and only needed for PDF I don't think it's a big deal to add. Thoughts @Daltz333 or @sciencewhiz ?

sciencewhiz:
There's a similar discussion happening in #2774

TheTripleV:
I think no one builds pdf locally so a new dep for pdf is fine.

This affects Sam's pr with gifs too.
Is the repo standard to use mp4 instead of gifs everywhere for filesize?

source/docs/software/driverstation/driver-station.rst
```diff
@@ -21,6 +21,9 @@ The FRC Driver Station can be launched by double-clicking the icon on the Deskto
   * :kbd:`[` + :kbd:`]` + :kbd:`\\` - Enable the robot (the 3 keys above Enter on most keyboards)
   * :kbd:`Enter` - Disable the Robot
   * :kbd:`Space` - Emergency Stop the robot. After an emergency stop is triggered the roboRIO will need to be rebooted before the robot can be enabled again.
+  * :kbd:`Backspace` - "A-Stop" the robot when in Practice Mode - Autonomous Enabled. The robot will be disabled until the Practice Mode reaches Teleop then will be automatically re-enabled. While disabled, the application background will flash orange indicating the robot will re-enable automatically.
+
+ .. image:: images/driver-station/ds-astop.gif
```
sciencewhiz, line 26: This should have an :alt:

sciencewhiz:
> Is the repo standard to use mp4 instead of gifs everywhere for filesize?

Both for filesize and for accessibility. It's not possible to disable an animated gif. If someone were to be sensitive to the blinking in this screenshot, they wouldn't be able to disable it, unlike a video.

sciencewhiz:
Everything looks to be working with mp4

source/docs/software/driverstation/driver-station.rst
```diff
@@ -21,6 +21,21 @@ The FRC Driver Station can be launched by double-clicking the icon on the Deskto
...
+      :loop:
+      :width: 900
+      :autoplay:
+      :muted:
+      :alt: Driver Station flashing yellow in A-Stop mode
+
+.. only:: not html
+
+    .. image:: images/driver-station/ds-astop.png
+       :alt: Driver Station flashing yellow in A-Stop mode
```
SamCarlberg, line 38: Should there be a subtitle or something to direct to the web docs to see an example of the animation playing?

jasondaming:
I think the visualization is a great idea and we should do more of that.

For this one however I had trouble following it mostly because I didn't understand when it stopped and where it began.  Because it auto replayed and there was little difference between the start and end states it looped seamlessly for me.  Two solutions I can think of: is there a way we can make is wait 5? seconds before repeating? or make is a little longer at the end so that it is more obvious when it repeats.

Kevin-OConnor:
> For this one however I had trouble following it mostly because I didn't understand when it stopped and where it began. 
Re-recorded longer and with showing the practice timing at the beginning so the loop is more obvious. Does that help?

jasondaming:
Was it shorter than 7 seconds before?  That is the video I see currently.

Not a big deal the orange blinking is the primary thing you are trying to highlight.

Kevin-OConnor:
> Was it shorter than 7 seconds before? That is the video I see currently.
> 
> Not a big deal the orange blinking is the primary thing you are trying to highlight.

That might be a cache issue? I see the updated 14s video showing here: https://frc-docs--2816.org.readthedocs.build/en/2816/docs/software/driverstation/driver-station.html

PR #2832, opened 2024-11-07
Update more CTRE links

sciencewhiz:
Update Pigeon code examples for Phoenix 6

### DIFF
--- a/source/docs/software/hardware-apis/sensors/gyros-software.rst
+++ b/source/docs/software/hardware-apis/sensors/gyros-software.rst
@@ -127,22 +127,16 @@ The navX uses the :code:`AHRS` class.  See the [navX documentation](https://pdoc
 
 ## Pigeon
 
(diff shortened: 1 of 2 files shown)


source/docs/software/hardware-apis/sensors/gyros-software.rst
```diff
@@ -127,22 +127,16 @@ The navX uses the :code:`AHRS` class.  See the [navX documentation](https://pdoc
 
 ## Pigeon
```
jasondaming, line 128:
```suggestion
## Pigeon 2
```

jasondaming (changes requested):
Pigeon is pretty different from Pigeon 2 and don't want to get them confused.  I am okay if we just want to document the 2 since it is newer and the most likely to be used.

jasondaming (approved):

PR #2837, opened 2024-11-09
Add New in 2025 Devices

jasondaming:


sciencewhiz:
The motor controllers and PDP 2.0 also need to be added to https://docs.wpilib.org/en/latest/docs/controls-overviews/control-system-hardware.html

sciencewhiz:
> PDP 2.0 shouldn't be added since it isn't a can enabled devices butg the PDP 1.0 and PDH should be.

This page is for CAN devices that need extra rivers. PDP and PDH are built into WPILib and don't need to be listed on the third part CAN device page.

sciencewhiz:
Motor controllers should be added to https://docs.wpilib.org/en/latest/docs/software/hardware-apis/index.html#frc-legal-motor-controllers

jasondaming:
Replaced by #2923

PR #2853, opened 2024-11-24
Update formatting to avoid blockquote on furo

sciencewhiz:


sciencewhiz:
This isn't complete, but has no side effects, and already had one merge error, so makes sense to merge now and more articles can be done in subsequent PRs.

PR #2855, opened 2024-11-28
Add deprecation notice for PathWeaver

Gold856:


sciencewhiz:
Needs conflicts resolved

source/docs/software/pathplanning/pathweaver/introduction.rst
```diff
@@ -1,6 +1,8 @@
 # Introduction to PathWeaver
 
-.. note:: Users may find :doc:`Choreo </docs/software/pathplanning/choreo/index>` or [PathPlanner](https://github.com/mjansen4857/pathplanner) more useful. They both have an intuitive user interface and swerve support. Users should already have access to Choreo as it is distributed with the WPILib installer.
+## Deprecation Notice
```
sciencewhiz, line 3: I assume this header is primarily to get an anchor to point to in the Pathweaver PR? Unfortunately it breaks up the article, the text below shouldn't be a part of the deprecation notice. Assuming that's the reason, I think using a :ref: would do the same thing without breaking up the document. https://www.sphinx-doc.org/en/master/usage/referencing.html#role-ref

Gold856: Yep, using a label works for getting something to link to, thanks!

sciencewhiz (approved):

Issue #2420, opened 2023-12-07
The Wiring Diagram of the Control System has the CAN colors backwards


Technologyman00:
Control System: ![image](https://github.com/wpilibsuite/frc-docs/assets/16891521/5e705a76-c574-489e-8fd9-793362876373)

Roborio: 
![image](https://github.com/wpilibsuite/frc-docs/assets/16891521/60ee7b39-30a7-4ad0-ae23-afd446319498)

Technologyman00:
Apologies I should have made that significantly clearer

jasondaming:
I messaged Stefen Acepcion (the creator of the image) on his website to see if he was able to help.

sciencewhiz:
Are the PDH labels backward?

![REV-11-1850-PDH-BirdsEye-FINAL__88635](https://github.com/wpilibsuite/frc-docs/assets/4885091/a6572d33-8a07-49d9-b82f-a36f5afbbaf8)

Technologyman00:
> Are the PDH labels backward? they show H for Yellow also. https://www.revrobotics.com/rev-11-1850/
> 
> ![REV-11-1850-PDH-BirdsEye-FINAL__88635](https://github.com/wpilibsuite/frc-docs/assets/4885091/a6572d33-8a07-49d9-b82f-a36f5afbbaf8)
> 

No that's correct. It's just that the Orientation of the wires is wrong. The Roborio has L or Green on top

Gold856:
Did the new diagrams in #2862 fix this issue?

Technologyman00:
Yep. Closing

Issue #2635, opened 2024-04-23
Implement Documentation links for the Vivid Hosting Radios


Suave101:
At worlds, everyone was given a brand new shiny VH-109 radio and everyone had their radio configured to the field at a kiosk. Can we have a documentation link to show how to have the radio broadcast its own signal. Here is the link to the website:
https://frc-radio.vivid-hosting.net/
The website is kind of confusing for newer teams when it comes to "Team Access Point Mode" and "Team Client/STA Mode". Maybe we could document what each mode means exactly. Furthermore, there are FCC regulations that must be followed with this new radio. There should be a warning to teams on our documentation that lets them know the FCC will track them down if they use the radio in certain modes in certain places. Not to mention other countries guidelines!

jasondaming:
There will be a large PR related to all of the radio changes.  I believe some of these configurations are still being worked out. Thanks for making an issue to track this.

Suave101:
Is there one currently being worked on that can be linked here for others to contribute to?

jasondaming:
It can be linked when it gets started.  We have to be extra careful (it affects a lot of teams and is part of the Zero to Robot guide) with how this is done so I wouldn't start on this yourself.

We haven't even finalized where the documentation will live or if it will be partially duplicated.

victoriadekalita:
We just got the new radio over the weekend and I cannot find the configuration website, the only thing I can find tells me to go to "radio.local" which never loads?? I am missing something and really need to get things rolling on the software, Where do I go, how do I get the software downloaded?

jasondaming:
I would suggest creating an issue on the docs page so we can help you on that and it will be more visible.

In the mean time how do you have it connected to your computer?  Did you see the[ troubleshooting in the docs](https://docs.wpilib.org/en/stable/docs/zero-to-robot/step-3/radio-programming.html#cannot-reach-the-configuration-page-at-radio-local) that talks about setting a static IP address and getting to it at: http://192.168.69.1/

PR #2862, opened 2024-12-11
Replacement of Control System Diagrams with new ones for the new radio

stefacep:
https://www.chiefdelphi.com/t/2025-unofficial-frc-control-system-diagrams/476263/14


sciencewhiz:
Since these are also used on the wiring robot page, there also needs to be updates to that page for the new radio, because otherwise they'll be out of sync. #2635

Gold856:
This fixes #2420 right?

jasondaming (approved):

PR #2863, opened 2024-12-11
Add AdvantageScope to Software Overview

sciencewhiz:


jasondaming:
That image looks really grainy.  I am not opposed to pushing without changing it but to me it stands out as overly compressed or something

sciencewhiz:
I see a little bit of what looks like resizing artifacts, but I wouldn't call it very grainy. It's the same picture that's used elsewhere in the docs.

jasondaming (approved):

PR #2867, opened 2024-12-14
Update compiler args doc for 2025

sciencewhiz:


jasondaming:
We should update that tag to get the changes from: https://github.com/wpilibsuite/vscode-wpilib/pull/732/files

jasondaming (approved):

PR #2874, opened 2024-12-16
Update what-is-wpilib.rst

arbessette:
Added link to wpilib.org


sciencewhiz:
Was the extra space between parentheses necessary? I don't think it should be

arbessette:
nope, I thought I had deleted it

sciencewhiz (approved):

sciencewhiz:
Since you're an admin, you have to be careful, because using the web editor will create a branch in the repo rather then in your fork by default.

PR #2883, opened 2024-12-21
Check for errors from requests.get

sciencewhiz:


TheTripleV:
Need parenthesis on both?

TheTripleV (approved):

PR #2913, opened 2025-01-04
Add known issue for LabVIEW showing 2024

sciencewhiz:
Fix alt for NI Package Manager image


jasondaming:
won't that workaround just confirm that you have the 2025 game tools?

sciencewhiz:
yes, so you can ignore the year and know that you have the right version installed. Otherwise, you could still be using 2024.

jasondaming:
What is preventing teams from just installing 2025 game tools and not 2025 LV.  Will it not launch or something if they just installed game tools?

sciencewhiz:
This is only an issue for LabVIEW

jasondaming (approved):

PR #2923, opened 2025-01-08
Updated Control System Hardware for 2025

Fenghuang4:
Fixed broken links to Rev docs.
Changed TalonFX image to be of the current version and specified its use on Kraken X60 and X44.
Added:
ATO breakers
PDP 2.0
Spark Flex
TalonFXS
Nova
Koors 40
Servo Hub

Reordered motor controllers based on 2024 data.


jasondaming:
Can we crop that ATO breaker image so that its not so "small" compared to the picture of the old breaker above it.

jasondaming:
We also need to add a The Thrifty Bot image credit at the bottom for the Nova

jasondaming:
looks like there is an extra duplicate ATO-breaker under the images directory.  Otherwise good to go

sciencewhiz:
please use lower case for the file extensions

jasondaming (approved):

PR #2927, opened 2025-01-09
Add wpi and wpilib logos

jasondaming:


arbessette (changes requested):
Remove white space and decrease size of logos, move the WPILIb logo above the paragraph to avoid the similar but different reds

sciencewhiz:
The WPILib logo already is in the /assets folder, can reference that one

arbessette (approved):

sciencewhiz:
Looks better, but don;t need to add wpilib logo, can use the one in /assets folder

jasondaming:
I explicitly adjusted the WPILib logo to get the right spacing needed.  I don't think using that logo will give the same spacing it will be back to the scenario where it is too close to the paragraph after it like it was before.

sciencewhiz:
Spacing looks good to me, so I think you can just delete the duplicate logo

sciencewhiz (approved):

PR #2932, opened 2025-01-10
updates the LED types that work

jasondaming:


sciencewhiz:
If 2815s work, should also PR allwpilib so the API docs indicate it

jasondaming:
Both [AM](https://www.andymark.com/products/12v-individually-addressable-led-light-strip-16-4-ft) and [Thrifty](https://www.thethriftybot.com/products/12v-individually-addressable-leds?_pos=1&_sid=418f82550&_ss=r) sell them for this.

sciencewhiz (approved):

PR #2947, opened 2025-01-15
Add article for JVM MSVC runtime error

sciencewhiz:


source/docs/software/advanced-gradlerio/jvm-runtime.rst
```diff
@@ -0,0 +1,56 @@
+# JVM Runtime error
+
+Programs compiled with Microsoft Visual C++ (MSVC) require a MSVC runtime that is equal to or newer then the newest library being executed. WPILib is compiled with a recent version of MSVC.
+
+In the past, this hasn’t really been used, but a change in the runtime last year actually took advantage of this requirement.
```
Starlight220, line 5: Replace "last year" with the actual year

source/docs/software/advanced-gradlerio/jvm-runtime.rst
```diff
@@ -0,0 +1,56 @@
...
+
+```console
+gradlew -Dorg.gradle.java.home="C:\\Users\\Public\\wpilib\\YEAR\\jdk" ...
+```
+
+### gradle.properties
+
+It can also be done by putting ``gradle.properties`` in the project root with the following contents (where YEAR is the current WPILib installation):
+
+.. warning:: This will only work if you only use your project on Windows, as the location is hardcoded to the location on Windows.
```
Starlight220, line 48: Mention what the Linux path is rather than limiting this to Windows

ThadHouse: I would rather just either leave this as is, or remove this option entirely. 

Basically what this is saying is that if you do this, your project will _only_ work on Windows. We have no way around that.

sciencewhiz: I added a recommendation not to check into SCM

ThadHouse (approved):

PR #2960, opened 2025-01-24
Update OpenMesh Tool to 25.0.0

Kevin-OConnor:


sciencewhiz:
Does this also update where it looks for Java? If so could update the note in line 38.

Kevin-OConnor:
> Does this also update where it looks for Java? If so could update the note in line 38.

Thanks Joe, it does add 2025 (and 2026 for future proofing) to the search list in the wrapper, I'll update that line.

PR #2975, opened 2025-02-09
Really update Python version info for 2025

auscompgeek:
Linux ARM docs were outdated.


sciencewhiz:
Looks like the artifactory link in https://github.com/wpilibsuite/frc-docs/blob/main/source/docs/software/python/pyproject_toml.rst also needs to be updated

sciencewhiz (approved):

PR #2987, opened 2025-02-23
Fix jvm-runtime.rst

arimb:
add missing space in code snippet


PeterJohnson:
A space is not required here, and various examples I’ve seen online don’t include one.

arimb:
Strange, it didn't work for me when I used it as written but did work when I added a space. Maybe it's only needed in certain terminals

sciencewhiz:
Both versions appear to work in CMD, but only the one with a space appears to work in powershell

sciencewhiz (approved):

PR #2990, opened 2025-03-07
Correct C++ example for Addressable LEDs page

GeovonniN:
Attempts to resolve https://github.com/wpilibsuite/frc-docs/issues/2991

Hi!

It's my understanding there's a typo in this C++ example. I assume the author wanted to have a lambda function that returned a double.


SamCarlberg (approved):
Good catch! Thanks for the fix

Issue #3002, opened 2025-03-18
Error in Command Compositions for Python


Suave101:
In the documentation for the [Repeating Command Composition](https://docs.wpilib.org/en/stable/docs/software/commandbased/command-compositions.html#repeating), the documentation references a function that [does not exist](https://robotpy.readthedocs.io/projects/commands-v2/en/stable/commands2.cmd/functions.html) in the RobotPy Commands2 API. The documentation is probably trying to cite the [repeatedly decorator](https://robotpy.readthedocs.io/projects/commands-v2/en/stable/commands2/Command.html#commands2.Command.repeatedly) of the [RobotPy commands2 Command Class](https://robotpy.readthedocs.io/projects/commands-v2/en/stable/commands2/Command.html#commands2.Command). The Java example is as follows:
```
// Will run forever unless externally interrupted, restarting every time command.isFinished() returns true
Command repeats = command.repeatedly();
```
Which uses a `Command` class that has already been created and adds the repeatedly decorator. To fix the Python example provided, one could change the code to:
```
# Will run forever unless externally interrupted, restarting every time command.IsFinished() returns true
repeatingCommand = nonRepeatingCommand.repeatedly()
```
Using this method, the new `repeatingCommand` variable now stores a [RepeatCommand](https://robotpy.readthedocs.io/projects/commands-v2/en/stable/commands2/RepeatCommand.html#commands2.RepeatCommand) as specified in the commands2 documentation [here](https://robotpy.readthedocs.io/projects/commands-v2/en/stable/commands2/Command.html#commands2.Command.repeatedly).

TheTripleV:
Good catch. Feel free to open a Pull Request changing https://github.com/wpilibsuite/frc-docs/blob/main/source/docs/software/commandbased/command-compositions.rst?plain=1#L54

PR #3003, opened 2025-03-18
Fix Error in Command Compositions for Python

Suave101:
Provides solution to issue #3002


source/docs/software/commandbased/command-compositions.rst
```diff
@@ -51,7 +51,7 @@ The ``repeatedly()`` decorator ([Java](https://github.wpilib.org/allwpilib/docs/
 
    ```python
    # Will run forever unless externally interrupted, restarting every time command.IsFinished() returns true
-   repeats = commands2.cmd.repeatedly()
+   repeatingCommand = nonRepeatingCommand.repeatedly()
```
Starlight220, line 54: Keep the old variable name for equivalence with Java/C++

Starlight220 (changes requested):

Starlight220 (approved):

PR #3012, opened 2025-04-14
Show where to select a language for XRP and Romi Reference examples

JonathanLindsey:
It seemed like the instructions for setting up the XRP and Romi projects were skiping the select a language step. This fixes that.

Not sure how one would create a Python project for XRP and Romi. Or a LabVIEW project for XRP.


sciencewhiz:
Thanks! Can you consolidate so the same image is used for both articles? Even better would be to consolidate the other VS code images in the articles, and even better would be to add those images to the zero to robot as well.

sciencewhiz (approved):

JonathanLindsey:
Thanks for completing!

PR #3019, opened 2025-04-23
WPILib installation guide: Minor tweaks

qwertychouskie:
Came across these when researching something else, decided to make a quick PR.

### DIFF
--- a/source/docs/zero-to-robot/step-2/wpilib-setup.rst
+++ b/source/docs/zero-to-robot/step-2/wpilib-setup.rst
@@ -1,6 +1,6 @@
 # WPILib Installation Guide
 
-This guide is intended for Java and C++ teams. LabVIEW teams can skip to :doc:`labview-setup`. Python teams can skip to :doc:`python-setup`. Additionally, the below tutorial shows Windows 11, but the steps are identical for all operating systems. Notes differentiating operating systems will be shown.
+This guide is intended for Java and C++ teams. LabVIEW teams can skip to :doc:`labview-setup`. Python teams can skip to the :doc:`python-setup`. Additionally, the below tutorial shows Windows 11, but the steps are identical for all operating systems. Notes differentiating operating systems will be shown.
 
 ## Prerequisites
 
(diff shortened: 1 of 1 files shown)


source/docs/zero-to-robot/step-2/wpilib-setup.rst
```diff
@@ -1,6 +1,6 @@
 # WPILib Installation Guide
 
-This guide is intended for Java and C++ teams. LabVIEW teams can skip to :doc:`labview-setup`. Python teams can skip to :doc:`python-setup`. Additionally, the below tutorial shows Windows 11, but the steps are identical for all operating systems. Notes differentiating operating systems will be shown.
+This guide is intended for Java and C++ teams. LabVIEW teams can skip to :doc:`labview-setup`. Python teams can skip to the :doc:`python-setup`. Additionally, the below tutorial shows Windows 11, but the steps are identical for all operating systems. Notes differentiating operating systems will be shown.
```
jasondaming, line 3: adding "the" isn't approriate here this is referring to a specific page.  It is the same in the previous sentence.

source/docs/zero-to-robot/step-2/wpilib-setup.rst
```diff
@@ -11,7 +11,7 @@ Supported Operating Systems and Architectures:
 
 .. warning:: The following OSes are no longer supported: macOS 12 or earlier, Ubuntu 18.04 & 20.04, Windows 7, Windows 8.1, and any 32-bit Windows.
 
-.. note:: [Windows 10 support from Microsoft will end in October 2025](https://www.microsoft.com/en-us/windows/end-of-support). We intend to continue supporting Windows 10 through the 2026 season, but may have to drop support in 2027. Teams should start planning their upgrade path to Windows 11.
+.. note:: [Windows 10 support from Microsoft will end in October 2025](https://www.microsoft.com/en-us/windows/end-of-support). We intend to continue supporting Windows 10 through the 2026 season, but may have to drop support in 2027. Teams should start planning their upgrade path to Windows 11, or switch to one of the supported Linux distributions listed.
```
jasondaming, line 14: I don't think switching to Linux is really a viable solution for most "zero to robot" teams.  Additionally support of components like the driver station are still unknown but switching a computer to Linux would have the additional downside of potentially not supporting the game tools.

jasondaming (changes requested):

source/docs/zero-to-robot/step-2/wpilib-setup.rst
qwertychouskie: The name of this page is "Python Installation Guide", so without the added "the" it reads "Python teams can skip to Python Installation Guide" which reads very poorly.  The other sentence currently reads as "LabVIEW teams can skip to Installing LabVIEW for FRC (LabVIEW only).", which isn't perfect, but there's not as clear of a quick tweak to make to make the sentence grammatically correct.

source/docs/zero-to-robot/step-2/wpilib-setup.rst
qwertychouskie: Some systems cannot be upgraded to 11 for one reason or another.  How about this? "Teams should start planning their upgrade path to Windows 11 if supported, or otherwise consider migrating to one of the supported Linux distributions listed."

jasondaming (approved):

PR #3023, opened 2025-04-26
Clean up WPILib setup

### LINKED ISSUE #2988: WPILib Installation Guide has misleading information
The first paragraph of the [WPILib Installation Guide](https://docs.wpilib.org/en/stable/docs/zero-to-robot/step-2/wpilib-setup.html) says that "[it] is intended for Java and C++ teams... Python teams can skip to Python Installation Guide." This is misleading because while teams using Python do not need to install everything, they still (to my knowledge) need to install the tools (Shuffleboard, SmartDashboard, etc.). I ran into this issue because my installation of the Driver Station kept complaining that SmartDashboard couldn't be found. I think multiple parts of the guide would need to be changed, including the step where it says to download everything.

sciencewhiz:
Add introduction
Clarify cases where python teams would want to install (fixes #2988) 
correct tools installed and add links
Put OS pictures note in a better location


jasondaming:
I broke your PR by merging the other one so I fixed it keeping your wording

PR #3040, opened 2025-05-27
Remove content for removed features in 2027

Gold856:
References to old Java tools, Labview, and other features removed in 2027 WPILib have been removed.


sciencewhiz:
I won't have a chance to review in detail for a while. However, from a usability perspective (and SEO) I think I'd tend to prefer replacing things with stubs that point to replacement rather then a wholesale delete for 2027. Since it's a massive PR, and I've only been able to glance at it, I can't tell if that's what you've done. 

See also #3041. I think it might be easier to reorganize first then delete if we're keeping stubs.

Gold856:
It's just wholesale deletes with some minor spot fixes to remove references or use alternate ones, e.g. replacing Shuffleboard with Elastic. And yeah, I think a reorganization first might be easier. Although I'm not exactly sure if stubs would be particularly useful compared to just using rediraffe. I suppose if we use one stub for every major removed component like SmartDashboard and Shuffleboard, that'd be fine.

What I might do is move the spot fixes (the first commit) into a separate PR, which would allow us to be more aggressive in moving/deleting stuff.

source/docs/software/advanced-gradlerio/gradlew-tasks.rst
```diff
@@ -23,9 +23,9 @@ This article aims to highlight the gradle commands supported by the WPILib team
 
 ## GradleRIO tasks
 
-``./gradlew $TOOL$`` - Runs the tool ``$TOOL$`` (Replace ``$TOOL$`` with the name of the tool. IE, Glass, Shuffleboard, etc)
+``./gradlew $TOOL$`` - Runs the tool ``$TOOL$`` (Replace ``$TOOL$`` with the name of the tool. IE, Glass, OutlineViewer, etc)
 
-``./gradlew $TOOL$Install`` - Installs the java tool ``$TOOL$`` (Replace ``$TOOL$`` with the name of the tool. IE, Shuffleboard, SmartDashboard, etc)
+``./gradlew $TOOL$Install`` - Installs the java tool ``$TOOL$`` (Replace ``$TOOL$`` with the name of the tool. IE, SysId, wpical, etc)
```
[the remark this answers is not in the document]
sciencewhiz: There also isn't an option to install C++ tools, so I think this whole line can be deleted.

source/docs/software/roborio-info/roborio-brownouts.rst
```diff
@@ -78,7 +78,7 @@ The NI roboRIO 2.0 adds the option for a software settable brownout level.  The
...
+In the meantime, teams can use their robot code and manual logging or AdvantageScope to visualize current draw on their robot as mechanisms are developed. For C++ and Java teams, use the PowerDistribution class as described in the :doc:`Power Distribution </docs/software/can-devices/power-distribution-module>` article. Plotting this information over time (easiest with AdvantageScope's Line Graph tab) can provide information to compare against and update your power budget or can locate mechanisms which do not seem to be performing as expected (due to incorrect load calculation, incorrect efficiency assumptions, or mechanism issues such as binding).
```
sciencewhiz, line 81: I believe Elastic has a PDP widget

source/docs/software/dashboards/index.rst
```diff
@@ -4,10 +4,7 @@
    :maxdepth: 1
 
    dashboard-intro
-   shuffleboard/index
```
sciencewhiz, line 7: I think not having things referenced in an index causes a build failure. Since they will eventually be stubs, probably just move them to the end rather then deleting

source/docs/software/roborio-info/roborio-brownouts.rst
Gold856: It explicitly says over time, and the previous wording pointed to graphs. I think it's better to recommend graphs instead of a widget that changes rapidly over time.

source/docs/romi-robot/hardware-support.rst
```diff
@@ -15,7 +15,6 @@ In general, the Romi is compatible with the following:
 Due to hardware limitations, the Romi Robot is not compatible with the following:
 
 - Encoders other than the Romi-integrated encoders
-- "Ping" style ultrasonic sensors (which require 2 DIO channels)
```
sciencewhiz, line 18: I think this is ok to leave here

Gold856:
I decided that I do want to do wholesale deletes and redirect to a stub. This mostly comes from the fact that since the reorg will necessitate articles to be moved around anyways, it will require all articles to find a new home, and a lot of articles that are currently up for removal (such as all the Java tools) are a mix of tutorial and reference, so we'd have to spend time to split them up, only to delete them. I think just deleting everything now, with all deletions pointing to one article listing all the removed features, and alternatives, will make things easier in the future.

source/docs/software/driverstation/driver-station.rst
```diff
@@ -13,8 +13,6 @@ For information on installing the Driver Station software see :ref:`this documen
 
 The FRC Driver Station can be launched by double-clicking the icon on the Desktop or by selecting Start->All Apps->FRC Driver Station.
 
-.. note:: By default the FRC Driver Station launches the :ref:`LabVIEW Dashboard <docs/software/dashboards/labview-dashboard/driver-station-labview-dashboard:FRC LabVIEW Dashboard>`.  It can also be configured on :ref:`docs/software/driverstation/driver-station:Setup Tab` to launch the other Dashboards: :ref:`SmartDashboard <docs/software/dashboards/smartdashboard/smartdashboard-intro:SmartDashboard Introduction>` and :ref:`Shuffleboard <docs/software/dashboards/shuffleboard/getting-started/shuffleboard-tour:Tour of Shuffleboard>`. WPILib must be :ref:`installed <docs/zero-to-robot/step-2/wpilib-setup:WPILib Installation Guide>` to use SmartDashboard and Shuffleboard.
```
sciencewhiz, line 16: This should be rewritten, rather then wholly deleted. It's still true that the LabVIEW dashboard is loaded by default, and can link to the documents for configuring it to launch Elastic

source/docs/software/hardware-apis/sensors/gyros-software.rst
```diff
@@ -156,7 +156,7 @@ Gyros are extremely useful in FRC for both measuring and controlling robot headi
 
 ### Displaying the robot heading on the dashboard
 
-:ref:`Shuffleboard <docs/software/dashboards/shuffleboard/getting-started/shuffleboard-tour:Tour of Shuffleboard>` includes a widget for displaying heading data from a gyro in the form of a compass.  This can be helpful for viewing the robot heading when sight lines to the robot are obscured:
+Shuffleboard includes a widget for displaying heading data from a gyro in the form of a compass.  This can be helpful for viewing the robot heading when sight lines to the robot are obscured:
```
sciencewhiz, line 159: I think this can be updated for Elastic's Gyro widget

source/docs/yearly-overview/removed-features.rst
```diff
@@ -0,0 +1,69 @@
...
+### PathWeaver
+Use Choreo or PathPlanner instead.
+
+## roboRIO Team Number Setter
+Config for SystemCore will be handled via the web interface.
+
+## Hardware
+
+### Relay
+No support.
```
sciencewhiz, line 28: I'd say something like SystemCore doesn't support X for all the hardware items below

source/docs/yearly-overview/removed-features.rst
```diff
@@ -0,0 +1,69 @@
...
+
+### Interrupts
+No support.
+
+### Axis Camera
+Not manufactured anymore.
+
+### Nidec Brushless
+
+## RamseteController/RamseteCommand
```
sciencewhiz, line 64: This should say something like "Use LTV Unicycle Controller which has more intuitive tuning"

source/docs/yearly-overview/removed-features.rst
```diff
@@ -0,0 +1,69 @@
...
+### DigitalSource
+
+### Interrupts
+No support.
+
+### Axis Camera
+Not manufactured anymore.
+
+### Nidec Brushless
+
```
sciencewhiz, line 63: Maybe something like support removed because it requires special software support and Usage Reporting show that it is no longer used

source/docs/yearly-overview/removed-features.rst
```diff
@@ -0,0 +1,69 @@
...
+## Java tools
+
+All Java GUI tools have been removed due to lack of maintenance.
+
+### Shuffleboard
+Use Elastic as a driver dashboard, use AdvantageScope for debugging and data visualization.
+
+### SmartDashboard
+Use Elastic as a driver dashboard, use AdvantageScope for debugging and data visualization.
```
sciencewhiz, line 14: I'd also say that it's removed because it only supports NetworkTables 3 and doesn't support modern platforms.

source/docs/software/hardware-apis/sensors/gyros-software.rst
Gold856: This shows use of the Shuffleboard API, which Elastic won't support. The whole article needs to be updated anyways, so I'm fine leaving this here, and fixing it when the APIs are ready.

source/docs/software/dashboards/glass/command-based-widgets.rst
```diff
@@ -25,8 +25,6 @@ The :guilabel:`Command Selector`  widget allows you to start and cancel a specif
    SmartDashboard.putData("My Command", command)
    ```
 
-.. note:: The ``MyCommand`` instance can also be sent using a lower-level NetworkTables API or using the :ref:`Shuffleboard API <docs/software/dashboards/shuffleboard/getting-started/shuffleboard-displaying-data:Displaying data from your robot>`. In this case, the ``SmartDashboard`` API was used, meaning that the :guilabel:`Command Selector` widget will appear under the ``SmartDashboard`` table name.
```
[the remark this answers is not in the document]
Gold856: To be honest, I've never liked the mention of a lower-level NT API. It fundamentally doesn't serve the same purpose at all as either the SmartDashboard API or Shuffleboard API. That statement can imply you can get similar functionality with the lower-level NT API, but doesn't tell you how, or that you'd need to plumb everything yourself. A lot of this will change with the telemetry API, so I'm hoping that we can be slightly more opinionated there. For now, I'm going to leave it like this.

sciencewhiz:
files deleted should be removed from the ignore list in `.github/workflows/inspector.json`

source/docs/controls-overviews/control-system-software.rst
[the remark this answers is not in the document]
sciencewhiz: People who happen to have a link to one of the deleted LabVIEW docs (not this one though) would get redirected to the [removed features](https://github.com/wpilibsuite/frc-docs/blob/4b4e207b62d670d9c5022b34fe15d0a0c43d59fd/source/docs/yearly-overview/removed-features.rst?plain=1). Thinking through the user experience in 2027 (when these docs are main), it probably would be helpful to have a link to to the 2026 docs from the removed features page (which would be a link to stable right now)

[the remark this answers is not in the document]
arbessette: that makes sense from user experience to me.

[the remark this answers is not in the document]
arbessette: Can you remind me why we are removing ultrasonics examples in exchange for encoders? Are teams using less ultrasonics now?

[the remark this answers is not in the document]
Gold856: That, and the fact that WPILib removed support for Ultrasonic hardware.

PR #3045, opened 2025-06-03
Add a special version warning for 2027

sciencewhiz:


source/conf.py
```diff
@@ -80,7 +80,14 @@
         <script>
             document.getElementById("versionwarning_href").setAttribute("href", location.href.replace("/latest/", "/stable/"))
         </script>
-        """
+        """,
+    "2027": """
+        The documentation you are currently viewing is for the SystemCore Alpha Test.
+        Please see the <a href="https://docs.wpilib.org" id="versionwarning_href">stable</a> version for the current release of WPILib.
+        <script>
+            document.getElementById("versionwarning_href").setAttribute("href", location.href.replace("/latest/", "/stable/"))
```
Starlight220, line 88: Won't this just not match anything, if the path is `/2027/`?

sciencewhiz: I'm not sure what that replace is actually doing, since there isn't any latest or stable either.

PR #3049, opened 2025-06-06
Add information about watching changes

KangarooKoala:


source/docs/contributing/watching-changes.rst
```diff
@@ -0,0 +1,44 @@
+# Watching changes
+
+This document describes different ways to stay up-to-date on the changes being made to the WPILib suite.
+
+## Stages of development
+
+To provide context about the different ways to stay up to date, below are the relevant stages of development:
+
+- Discord: The [Unofficial FIRST Robotics Competition Discord Server](https://discord.com/invite/frc) contains a programming discussion channel where people bring up their initial ideas and then receive feedback on them, either fleshing them out or learning why they might not be good ideas. Most of the major changes will have some discussion in here, but not all.
```
ArchdukeTim, line 9: I think it's important to mention:
- the number of major wpilib contributors that are active (and if this channel is considered the preferred place for ideas to being
- The reason the discord works i.e instant messaging makes the conversations move quickly, allowing ideas to get out of the idea phase and into the planning stage quicker

source/docs/contributing/watching-changes.rst
```diff
@@ -0,0 +1,44 @@
...
+- GitHub issue: Issues on the relevant GitHub repository are the next step toward a change being implemented. If the idea has gone through Discord, then this simply presents the fleshed out idea; Otherwise, this presents the idea and sometimes get initial discussion. Sometimes changes will skip this step and move straight to a pull request.
```
ArchdukeTim, line 10: nit: assuming this is meant to be an ordered list, saying this is the next step is superfluous

Probably needs a comparison to when something is ready for a GitHub issue (if ever? Changes can go from discord to PR, just like they can first pop up in PR). Github issues provide a single place for all discussion about a topic. Useful when the discussion spans a long period of time, or history of the discussion is useful

KangarooKoala: Fair point that "next step" is superfluous, and that reminds me that I should probably make the ordered nature more apparent. Maybe I'll change it to be a numbered list, but I'm hesitant since that could imply that everything needs to happen when that isn't the case. Maybe instead a wording change?

I agree that the wording should be clearer about the difference between GitHub issues and Discord conversations, but I do want to point out is that the stages of development are provided "To provide context about the different ways to stay up to date" (so the target audience is an observer), not as a guide for how someone should develop an idea.

source/docs/contributing/watching-changes.rst
KangarooKoala: > * (and if this channel is considered the preferred place for ideas to being

Could you finish this thought?

KangarooKoala:
Will we need a note about the Discord server being unofficial and not endorsed by FIRST? (There's a note in the [Unofficial Support](https://docs.wpilib.org/en/stable/docs/software/support/support-resources.html#unofficial-support) section of the docs on the support resources page) If so, where would we put that on the page?

KangarooKoala:
> Can you make the relevant updates first to allwpilib's contributing.md? That gets a lot more eyes from the development community so we can work out what exactly to say there, and then we can figure out what parts are best to leave only there and which are best to copy into frc-docs.

I agree that it'd get better visibility in allwpilib, but the existing wording in CONTRIBUTING.md is pretty strongly targeted towards people wanting to make changes instead of simple seeing what's going on and commenting:
https://github.com/wpilibsuite/allwpilib/blob/075cc4a20f44e2a5a61ed6139a855578cceb0b89/CONTRIBUTING.md?plain=1#L1-L3
```md
# Contributing to WPILib

So you want to contribute your changes back to WPILib. Great! We have a few contributing rules that will help you make sure your changes will be accepted into the project. Please remember to follow the rules in the [code of conduct](https://github.com/wpilibsuite/allwpilib/blob/main/CODE_OF_CONDUCT.md), and behave with Gracious Professionalism.
```

I'd also argue that instead of allwpilib having all of information and frc-docs having less, frc-docs should have more general information while allwpilib should have more targeted information, since frc-docs has a much wider audience.

> WRT to discord, I think it should be marked as unofficial, and because of that, it should be described as a good way to get inputs in a fast-paced manner, but not be described as a necessary step in the process.

Yep, 100% agree that it shouldn't be necessary or be described as necessary. Did something in the text suggest that it was necessary? I thought I worded it so it wouldn't come off that way, but if it did seem to imply that, then I should reword it.

PR #3059, opened 2025-06-15
Add WPILib development timeline

Gold856:


source/docs/software/dev_timeline.py
```diff
@@ -0,0 +1,174 @@
+#!/usr/bin/env python3
+
+import datetime
+
+from matplotlib.axes import Axes
+import matplotlib.dates as mdate
+import matplotlib.pyplot as plt
+
+START_YEAR = 2027
+END_YEAR = 2032
```
sciencewhiz, line 10: I feel like 3 years is enough to show the concept, and will probably be easier to read

Gold856: User feedback from the FRC Discord asked for more than 3 years. I'll lower it to 4 as a compromise.

Gold856:
Sure, but I’m gonna need help on that because I don’t know how to attach it as a Sphinx extension.

sciencewhiz:
There's examples here: https://github.com/wpilibsuite/frc-docs/tree/main/source/_extensions https://github.com/wpilibsuite/frc-docs/blob/main/source/conf.py#L64-L72

Gold856:
After experimenting around, it looks like Sphinx requires the file to exist in the source tree, so I'm just going to commit the file. It might be workable to make it into a directive or do something else to inject the file in, but I think that's too much effort and hackyiness compared to just commiting the file and updating it occasionally.

source/docs/software/what-is-wpilib.rst
```diff
@@ -39,3 +39,12 @@ While users are strongly encouraged to read the source code to resolve detailed
...
+
+.. image:: /assets/dev_timeline.svg
+  :alt: A timeline showing the different periods of WPILib's development for the next 4 years and how the different periods overlap between different, starting with development from FRC Kickoff to September, FRC beta from September to the next FRC Kickoff, FRC Support from FRC Kickoff to Champs, Updates/FTC Beta from Champs to FTC Kickoff, and FTC Support from FTC Kickoff to the next Champs.
+
+For year N, WPILib will create a branch for year N - 1, and actively do development on it until around October, when it enters the beta testing period for FRC, which continues until FRC Kickoff for year N, which is when the first stable release is published. FRC will receive active support until the FIRST Championship ends, which then begins the FTC beta testing period, which lasts until FTC Kickoff in September of year N. FTC will then receive active support until the next FIRST Championship ends, and then the branch will no longer be supported.
```
sciencewhiz, line 50: I had a little trouble following the years, I think this is a little easier to follow.

```suggestion
For year N, WPILib will create a branch during year N - 1, and actively do development on it until around October of year N-1, when it enters the beta testing period for FRC, which continues until FRC Kickoff for year N, which is when the first stable release is published. FRC will receive active support until the FIRST Championship ends, which then begins the FTC beta testing period, which lasts until FTC Kickoff in September of year N. FTC will then receive active support until the next FIRST Championship ends of year N + 1, and then the branch will no longer be supported.
```

source/docs/software/dev_timeline.py
sciencewhiz, line 1: The legend is hard to read, and also doesn't show the beta period stripes well. It should be significantly bigger. It would also be good to make the title larger to differentiate from the other text.
Is it possible to bring the year bars to the front?

PR #3068, opened 2025-06-21
Update to Sphinx 8

sciencewhiz:
Drop python before 3.11 (not supported by Sphinx 8)
Update all dependencies


sciencewhiz:
It looks like delta has a problem with Sphinx 8

sciencewhiz:
Done in #3269

PR #3080, opened 2025-07-15
Fix Ubuntu support timeline and formatting

Gold856:
This puts back the original wording that I had and shifts all the versions up by one LTS.


sciencewhiz:
Unfortunately, we still use rst table formatting, instead of md table formatting. https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#grid-tables

sciencewhiz (dismissed):

PR #3086, opened 2025-08-14
updated logo to include WPI full name 

arbessette:
WPI is working through a rebrand to use the full name when possible. This line allows that without changing the WPILib logo


Kevin-OConnor:
In wonderfully confusing fashion, the logo here is just a reference and isn't actually used anywhere (I don't think). The individual sized files in https://github.com/wpilibsuite/frc-docs/tree/main/source/_static are what are actually displayed. It looks like I was the last one to touch those, but I will admit I don't remember what process I used to generate them from the single base image.

arbessette:
Ok, I can make the same size images manually. Thanks!

arbessette:
Is everything ok with the new images or do I need to change the height?

PR #3099, opened 2025-10-05
2027 Docs Restructure

jasondaming:
initial restructure according to documentation template: https://docs.google.com/document/d/1hfguBWJAnKySKMACouyjNGWFuWJTAfZ_YTXg_Djp714/edit?tab=t.0

I marked this as a draft pull request because it is no where near ready to merge.  I just wanted to get something on paper to hopefully stir "criticism"


Gold856:
My immediate gut feeling is that it's too nested. Why is there a user manual index under a user manual heading? Why can't all of the contents be elevated up so it's all directly under the user manual heading? At the same time, too much information all at once can be problematic, but I wonder if this also hurts discoverability.

PR #3116, opened 2025-10-08
Fix OS detection for Firefox and Safari browsers

### LINKED ISSUE #3032: WPILib install page fails to detect OS in Firefox
On https://docs.wpilib.org/en/stable/docs/zero-to-robot/step-2/wpilib-setup.html I get the following error in the console.

```
Uncaught (in promise) TypeError: navigator.userAgentData is undefined
    <anonymous> https://docs.wpilib.org/en/stable/docs/zero-to-robot/step-2/wpilib-setup.html:926
```

screenshot

The offending source code line:

https://github.com/wpilibsuite/frc-docs/blob/f8306ce0759e9ce5eb504178a16ae2e49e3de5d3/source/_extensions/wpilib_release.py#L86

jasondaming:
## Summary
Adds fallback to userAgent parsing when userAgentData API is unavailable, fixing OS detection for Firefox and Safari browsers.

## Changes
- Check if `navigator.userAgentData` exists before using it
- Add fallback to traditional `navigator.userAgent` string parsing for Firefox/Safari
- Detect Windows by checking for "Windows" in user agent
- Detect macOS Intel vs ARM by checking for "Intel Mac OS X" in user agent string

Fixes #3032

### DIFF
--- a/source/_extensions/wpilib_release.py
+++ b/source/_extensions/wpilib_release.py
@@ -83,7 +83,32 @@ class WpilibRelease(SphinxDirective):
       <script>
         addEventListener('DOMContentLoaded', async (event) => {{
             let dlbutton = document.getElementsByClassName("wpilibrelease-dl-button")[0];
-            let ua = await navigator.userAgentData.getHighEntropyValues(['architecture', 'bitness', 'mobile', 'platform', 'platformVersion']);
+            let platform, architecture;
+
+            // Try modern API first (Chromium-based browsers)
+            if (navigator.userAgentData) {{
+                let ua = await navigator.userAgentData.getHighEntropyValues(['architecture', 'bitness', 'mobile', 'platform', 'platformVersion']);
+                platform = ua['platform'];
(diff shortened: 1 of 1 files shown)


TheTripleV:
Oh I think I tried to use LLMs on this one but they always hallucinated.

You can have Claude detect Safari and make it assume that Safari means Apple Silicon if you want.
Differentiating between Safari macOS x86-64 and Safari macOS arm64 is not possible and the claude prompt should be told that.

Firefox does not support the chrome api. It might work through
> Firefox doesn't currently work but could work in a hacky way: oscpu always returns MacIntel, even on arm macs. But loading webgpu and querying the gpu name will differentiate the 2 architectures.

but idk

I'm not sure why it's not rendering. js nested in html nested in rst nested in python is a bit cursed and there might be a syntax error somewhere or there might be a cors issue on pr builds.

TheTripleV:
funny enough this might be the hardest frc-docs issue for an llm to tackle

PR #3124, opened 2025-10-08
Draft of integrating driver station best practices into zero-to-robot path

stevep001:
I am thinking about integrating the driver station best practices into the zero-to-robot learning path.  Before I get too far on this want to check to see if folks think this is a good idea.


jasondaming:
Agreed on consulting with others as I know FIRST really likes to have tight control over what and how ZTR works/looks.

There is also likely a docs reorg coming up for 2027 so this change may get trampled in that.

PR #3148, opened 2025-10-12
Clarify Enable Desktop Support checkbox for all project types

### LINKED ISSUE #3146: XRP Java Programming Part 1 - no guidance for 'Enable Desktop Support' in WPILib Project Creator
Repro Steps
1) using Java Programming Part 1 course on firstinspires.org, go through XRP path to begin leveraging the Example Project...
2) while creating a new 
see also source/docs/xrp-robot/programming-xrp.rst

EXPECTED
The doc tells me whether I need to check "Enable Desktop Support."
Based on my limited understanding of how the XRP will work with WPILib, I believe it does use the simulation environment and somehow communicates that to some code on the Raspberry Pi in the XRP... so I expect there to be a statement like that in the tutorial (and if not a statement to that effect.)

ACTUAL
No guidance - so I'm not sure if I need to include Desktop Support or not.

jasondaming:
## Summary
Improves the documentation for the "Enable Desktop Support" checkbox in the New Project Creator, addressing confusion reported in issue #3146. This fix applies to all project types, not just XRP.

## Changes
Replaced the vague existing text with clear guidance:

**Before:**
> Enables unit test and simulation. While WPILib supports this, third party software libraries may not. If libraries do not support desktop, then your code may not compile or may crash. It should be left unchecked unless unit testing or simulation is needed and all libraries support it.

**After:**
- **Java**: This option has no effect and can be left checked or unchecked
- **C++**: Checking this option enables desktop compilation, which is required for simulation, unit tests, and certain development workflows (like Romi or XRP projects)
- Added note about third-party vendor library compatibility for C++ users

## Why This Matters
Users (especially those following tutorials) were confused about whether to check this box. The original text suggested it should be "left unchecked unless needed," but didn't explain that:
- For Java, it doesn't matter at all
- For C++, it's required for simulation/testing
- For Romi/XRP, C++ users must have it enabled

## Additional Changes
- Removed Python from the explanation (Python doesn't use this project creator)
- Added link to simulation documentation for conditional compilation guidance

Fixes #3146

### DIFF
--- a/source/docs/software/vscode-overview/creating-robot-program.rst
+++ b/source/docs/software/vscode-overview/creating-robot-program.rst
@@ -126,13 +126,13 @@ The elements of the New Project Creator Window are explained below:
 5. **Project Name**: The name of the robot project.  This also specifies the name that the project folder will be given if the Create New Folder box is checked.
 6. **Create a New Folder**: If this is checked, a new folder will be created to hold the project within the previously-specified folder.  If it is *not* checked, the project will be located directly in the previously-specified folder.  An error will be thrown if the folder is not empty and this is not checked.
 7. **Team Number**: The team number for the project, which will be used for package names within the project and to locate the robot when deploying code.
-8. **Enable Desktop Support**: Enables unit test and simulation support.
+8. **Enable Desktop Support**: Enables unit test and simulation support (see :doc:`/docs/software/wpilib-tools/robot-simulation/introduction`).
 
    - **Java**: This option has no effect and can be left checked or unchecked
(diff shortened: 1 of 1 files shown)


sciencewhiz:
> So it really sounds like that portion of the video needs to be removed

What video?

jasondaming:
I was referring to the Java Programming Part 1 video that triggered the original issue.  On reviewing them the Romi project erroneously instructs the user to check the box while the XRP project doesn't mention it and leaves it unchecked.  Probably not a big deal since it doesn't matter.

sciencewhiz (approved):

PR #3150, opened 2025-10-14
Adds driver station examples, note about non-English Windows

stevep001:
* adds pictures of three driver stations
* makes an editing pass to make driver station / driver station laptop / driver station application clearer
* adds recommendation about non-English Windows editions


Kevin-OConnor:
I agree that 6GHz is a convenience feature at best. It enables wireless connection between DS and a 2nd VH-109 as an AP, but plugging into that AP over Ethernet is perfectly viable. Along the direction Amanda is suggesting, even for new laptops you would probably end up cheaper hanging a 5GHz radio off your VH-109 (albeit with a little added complexity) than you would limiting your laptop search to 6G capable.

And starting in 2027, the SC 5Ghz Wifi becomes an option as well. While it will depend on the size of your practice space, existing 5GHz environment there, SC mounting location within your robot, and how good the radio in the laptop is, it has been working pretty reasonably for me so far.

stevep001:
> The images are huge (30+ mb). Since they're photos, jpg might be a more appropriate image type, and they should be scaled down.
> 
> Per the style guide, images should be 500kb or less. https://docs.wpilib.org/en/stable/docs/contributing/frc-docs/style-guide.html#images

I must have put the wrong images in the folder, as I did go and rescale them, but here they are now.

sciencewhiz (approved):

PR #3157, opened 2025-10-27
Build multipage HTML zip

Gold856:
Fixed up by overriding the htmlzip build: https://docs.readthedocs.com/platform/stable/build-customization.html#extend-or-override-the-build-process

This makes the htmlzip build command the same as before (instead of building singlehtml): <https://app.readthedocs.org/projects/frc-docs/builds/22213331/>

### DIFF
--- a/readthedocs.yml
+++ b/readthedocs.yml
@@ -14,7 +14,10 @@ build:
   jobs:
     post_checkout:
       - git fetch --unshallow || true
- 
+    build:
+      htmlzip:
+        - python -m sphinx -T -E -W --keep-going -b readthedocssinglehtmllocalmedia -d _build/doctrees -D language=$READTHEDOCS_LANGUAGE . $READTHEDOCS_OUTPUT/htmlzip
+
 formats:
   - htmlzip
   - pdf


TheTripleV:
This seems fine to send at first glance.

https://app.readthedocs.org/projects/frc-docs/builds/30087121/ shows the RTD currently auto builds an html zip using the `singlehtml` builder.

So if this doesn't work, the alternative is to keep the yaml the same but change https://github.com/wpilibsuite/frc-docs/blob/main/source/_extensions/rtd_patch.py#L13 to `singlehtml` instead.

Whichever you think is more maintainable.

Gold856:
I'd have questions about which builder would be used if we also called our own builder `singlehtml`, so let's see if we can stick with this.

sciencewhiz:
It failed: https://app.readthedocs.org/projects/frc-docs/builds/30087796/

Gold856:
I don't know why I didn't think of this before, but I'm gonna try again and just make sure a local build with the htmlzip build command succeeds, then PR that.

PR #3159, opened 2025-10-27
Fix downloadable HTML zip being one HTML page

Gold856:
One local build later, it turns out that RTD will actually run the build commands in the `source` directory, _unless_ you override the build command, where it just places you at the root repo, with the difference not being super obvious.

Just to be sure this works in RTD-land, I'm first going to add it as a `post_build` command, which _will_ execute on PRs and will actually give you a non-functional link that you can fix up to then use to download the docs. After we confirm that it spits out the ZIP correctly, then we can integrate it properly.

### DIFF
--- a/readthedocs.yml
+++ b/readthedocs.yml
@@ -17,8 +17,9 @@ build:
     build:
       htmlzip:
         - mkdir --parents _readthedocs/htmlzip
-        - python -m sphinx -T -W --keep-going -b readthedocssinglehtmllocalmedia -d _build/doctrees -D language=$READTHEDOCS_LANGUAGE source $READTHEDOCS_OUTPUT/out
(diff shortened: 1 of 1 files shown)


Gold856:
https://frc-docs--3159.org.readthedocs.build/_/downloads/en/3159/htmlzip/ (created by fusing the base PR docs domain with the path to the zip, which you can get by clicking the RTD flyout and copying the link from Downloads > HTML) is an archive that has the multipage HTML files. The `post_build` command had to be modified a tiny bit because I couldn't just build directly into the htmlzip directory, but that's where the normal `htmlzip` build output goes, so that should be fine. (emphasis on _should_) If you think it looks good, let me know, and I can uncomment out the `build` block and remove the `post_build` command.

PR #3160, opened 2025-10-28
Change the root zip directory name

Gold856:


sciencewhiz (approved):

sciencewhiz:
It extracts to frc-docs rather then frc-docs-latest. I think I'd rather change that on the installer though.

PR #3181, opened 2025-11-20
Migrate to uv

rzblue:
This migrates CI, readthedocs, and eventually the build documentation to use [uv](https://docs.astral.sh/uv/).

uv should have a better experience for both experienced devs and newcomers. The current documented build process has users installing packages globally, which is prone to package conflicts. It also uses poetry only for dependency resolution and a separate step to generate the requirements.txt for readthedocs. Using uv for everything makes building easy - after installing uv, `uv run make html` will install dependencies into a local venv and build like normal, and dependencies can be managed directly through pyproject.toml. uv will also automatically download and use a compatible python version if the local install is not compatible with the project. I've kept my system python at 3.12 because frc-docs needed it to build; with this that isn't necessary.

This isn't 100% done, I wanted to open this and see if this was something we'd be willing to do.

Readthedocs [documents using uv to install build dependencies](https://docs.readthedocs.com/platform/stable/build-customization.html#install-dependencies-with-uv). I've also tested it and it looks to be working fine: https://app.readthedocs.org/projects/rzblue-frc-docs/builds/

I've migrated the lockfile using `migrate-to-uv`, and generated a new requirements.txt. once hashes and comments are removed and contents sorted, the old and new files are line-for-line identical save for a one-line difference that doesn't change behavior (poetry includes `fonttools` and `fonttools[woff]` as well as their dependencies, while uv just resolves `fonttools[woff]`'s extra dependencies). I've also added a CI job to generate and make sure it's up to date.

I'd also like to make corresponding updates to frc-docs-translations for consistency and to eventually remove the requirements.txt. In order to do that without needing to merge PRs in lockstep, I've left the requirements.txt in this repo for the time being. Once this is merged, a PR can be made to -translations, and after that is merged, requirements.txt can be removed.

This also updates everything to ubuntu 24.04.

### DIFF
--- a/source/docs/contributing/frc-docs/build-instructions.rst
+++ b/source/docs/contributing/frc-docs/build-instructions.rst
@@ -40,7 +40,7 @@ Install the missing MikTex packages by navigating to the frc-docs directory, the
 
 ```console
 $ sudo apt update
-$ sudo apt install python3 python3-pip
+$ sudo apt install python3
 $ sudo apt install -y texlive-latex-recommended texlive-fonts-recommended texlive-latex-extra latexmk texlive-lang-greek texlive-luatex texlive-xetex texlive-fonts-extra dvipng librsvg2-bin
 ```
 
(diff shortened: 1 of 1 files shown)


rzblue:
> In uv, it will install different versions of scipy based on which version of python you're using.

It *should* be using the version in the lockfile (1.15.3 currently). What versions of python were you using?

source/docs/contributing/frc-docs/build-instructions.rst
```diff
@@ -32,17 +34,13 @@ Ensure that Python is in your Path by selecting the **Add Python to PATH** toggl
...
 
-Once Python is installed, open up Powershell. Then navigate to the frc-docs directory. Run the following command: ``pip install -r source/requirements.txt``
-
 Install the missing MikTex packages by navigating to the frc-docs directory, then running the following command from Powershell: ``miktex --verbose packages require --package-id-file miktex-packages.txt``
 
 ### Linux (Ubuntu)
 
 ```console
 $ sudo apt update
 $ sudo apt install python3 python3-pip
```
sciencewhiz: python3-pip can removed, right?

PR #3192, opened 2025-12-03
Refactor controls js generation

rzblue:
- Only regenerate output when input changes. This fixes issues with autobuilders that trigger rebuilds on filesystem events
- Use logging api instead of print. This makes the extension obey sphinx's verbosity settings
- Use pathlib
- use snake_case

### DIFF
--- a/source/_extensions/controls_js_sim/__init__.py
+++ b/source/_extensions/controls_js_sim/__init__.py
@@ -1,9 +1,10 @@
 from pathlib import Path
+from textwrap import dedent
 from typing import Any, Dict
 
-import os, glob
 from jsmin import jsmin
 from sphinx.application import Sphinx
+from sphinx.util import logging
 
 # Handle custom javascript
 # Groups, sorts, merges, and minifies the JS files associated with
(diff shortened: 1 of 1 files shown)


source/_extensions/controls_js_sim/__init__.py
```diff
@@ -25,74 +25,104 @@
...
+                    outf.write("//**    {}\n".format(source_file))
+                    outf.write(
+                        "//*******************************************************\n"
+                    )
+                    outf.write(
+                        "//*******************************************************\n"
+                    )
+                    outf.write("\n")
+                    outf.write(source.read())
+                    outf.write("\n")
```
katzuv:
```suggestion
output_text = ""
for source_file in source_files:
    text = ""
    if not debugJS:
        # Minify each file independently - again, low bar solution for now
        minified = jsmin(source_file.read_text())
        text += minified + "\n"
        continue
    
    # Verbose, no minify, and add debug markers.
    text = f"""\n\n\n
    //*******************************************************
    //*******************************************************
    //**    {source_file}
    //*******************************************************
    //*******************************************************
                              
    {source_file.read_text()}
    """
    output_text += textwrap.dedent(indented_string)

OUTPUT_FILE.write_text(output_text)
```
Further testing with `dedent` is needed, but I think it's more concise.

sciencewhiz (approved):

PR #3242, opened 2026-01-27
[WIP] add fire control simulation

Oblarg:
includes way too many features!  i am particularly proud of the cotangent geodesic and the convergence envelope ;)


ethanc8:
In "Static Shooting via Look-Up Table", there is a link to https://en.wikipedia.org/wiki/Artillery_fire_control#Firing_table, which is an article that does not exist.

PR #3282, opened 2026-05-02
Update to Node 24 versions of github actions and py 3.11

sciencewhiz:


Project516 (approved):
LGTM - I doubled checked and all actions are latest

PR #3296, opened 2026-05-09
Update changelog for Alpha 6

sciencewhiz:


jasondaming (commented):
The new Simulation section entries and the Romi/XRP Alpha 6 entry are missing the `- ` list prefix, which causes them to run together with surrounding content instead of rendering as a bullet list. Could you add the dash to these lines?

```rst
- 2027 Alpha 6: Don't wait on init for ``ExpansionHub`` in simulation.
- 2027 Alpha 6: Add Onboard IMU Simulation
- 2027 Alpha 6: GUI: fix game message string lifetime
- 2027 Alpha 6: Make HalSim DS extension a no-op, fixing crash when running with the real DS option. Support for the real DS simulation option will be added in a future release.
```

And for Romi/XRP:
```rst
- 2027 Alpha 6: Show IP and port of XRP/Romi to user on connection
```

jasondaming (approved):

PR #3299, opened 2026-05-13
Clarify instructions about building and deploying code

Gold856:
The instructions described how to build and deploy code, but only gave instructions for building, then inserted a large image, _then_ specified how to deploy, which is really easy to miss. Reorder it so it's harder to miss. Also replace roboRIO with Systemcore.


sciencewhiz:
I think we need to talk about the difference between build and deploy. For example, build is online, deploy is offline. Run build after updating vendordeps or after freshly cloning a repo. Build runs unit tests, deploy doesn't, etc. 

I could also see an argument for separating tutorial from reference documentation. This is linked from ZTR, but it would probably be better to add a quick deploy into ZTR, and leave this as reference build and deployt with all the options/differences.

sciencewhiz:
See also https://github.com/wpilibsuite/SystemcoreTesting/discussions/249#discussioncomment-16909766

sciencewhiz (approved):

PR #3304, opened 2026-05-15
Replace FRC logo with FIRST logo

arbessette:
Replaces the FRC logo with The FIRST logo


jasondaming (changes requested):
Assuming we want to target 2027?  We also need to modify the files in source/_static I believe.  The preview didn't show as working for me.

arbessette:
statics have been updated

PR #3307, opened 2026-05-17
Switch to native readthedocs uv support

sciencewhiz:


jasondaming:
Did we intentionally drop the --frozen parameter that looks at uv.lock?

sciencewhiz:
Based on the [documentation](https://docs.readthedocs.com/platform/stable/config-file/v2.html#uv) I don't think RTD supports parameters for the sync command.

It shouldn't matter though, since the `check-dependencies` CI check ensures the lock file is up to date, so we'd never get to the point where just uv sync would fail.

PR #3328, opened 2026-07-20
add opmoderobot framework

zachwaffle4:
I didn't include any Python examples because I don't even know if we have Python examples with OpModeRobot rn


ThadHouse:
Update the DS used for screenshots.

sciencewhiz:
PRs for 2027 should be targeted to main now. I retargetted the PR

source/docs/software/basic-programming/opmodes.rst
```diff
@@ -0,0 +1,244 @@
...
+
+**When the robot transitions from disabled to enabled**, ``start()`` is called exactly once.
+
+**While the robot is enabled**, ``periodic()`` is called repeatedly at ``OpModeRobot#getPeriod()`` (default 20 ms). Additional callbacks registered via ``addPeriodic()`` run at their own configured rates.
+
+**When the robot disables or a different opmode is selected while enabled**, ``end()`` is called first, then ``close()`` (Java) or the object is destroyed (C++/Python). The object is never reused.
+
+.. note:: Selecting a different opmode while the robot is enabled automatically disables the robot first, so ``end()`` is always called before the switch.
+
+**If a different opmode is selected while the robot is already disabled**, ``close()`` is called (or the object is destroyed) with no ``end()`` call, since the opmode was never enabled.
```
sciencewhiz: This is worded a little weird. Does the behavior change if the opmode was or wasn't enabled?

source/docs/software/basic-programming/opmodes.rst
```diff
@@ -0,0 +1,244 @@
...
+
+Callbacks are registered immediately at opmode construction and run even while the robot is disabled.
+
+## Migration from TimedRobot
+
+Teams switching from ``TimedRobot`` to ``OpModeRobot``:
+
+- Replace per-mode methods in ``Robot`` (``autonomousInit``, ``teleopPeriodic``, etc.) with separate ``@Autonomous`` and ``@Teleop`` opmode classes.
+- Replace ``SendableChooser`` with multiple ``@Autonomous`` classes.
+- Replace ``utilityInit``/``utilityPeriodic`` with ``@Utility`` opmode classes.
```
sciencewhiz: Any reason not to combine Utility into the Autonomous/Teleop bullet above?

source/docs/software/vscode-overview/creating-robot-program.rst
```diff
@@ -6,6 +6,28 @@ Once everything is installed, we're ready to create a robot program.  WPILib com
...
+- [C++](https://github.com/wpilibsuite/allwpilib/blob/v2027.0.0-alpha-6/wpilibc/src/main/native/include/wpi/framework/OpModeRobot.hpp)
+
+``OpModeRobot`` is the recommended base class for most users.  It structures robot code around individual *opmode* classes — one per autonomous routine, one per teleoperated configuration, and so on — annotated with ``@Autonomous``, ``@Teleop``, or ``@Utility`` (Java) or registered manually (C++).  The Driver Station displays a drop-down list of available opmodes for each mode, so operators can select routines without recompiling the robot program.  The ``Robot`` class holds shared hardware and subsystems, which opmodes receive through their constructors.
+
+See :doc:`/docs/software/basic-programming/opmodes` for full documentation, including the opmode lifecycle, multiple opmode selection, and DS integration.
+
+.. tab-set-code::
+
+   .. rli:: https://raw.githubusercontent.com/wpilibsuite/allwpilib/v2027.0.0-alpha-6/wpilibjExamples/src/main/java/org/wpilib/templates/opmode/Robot.java
+      :language: java
```
sciencewhiz, line 26: Use the `:lines:` and `:lineno-match:` directive to avoid boilerplate (like the copyright notice). See below for examples

sciencewhiz (commented):
I think it would be helpful to have a short section describing why opmodes (ie similar to the motivation section from the design document). It would also be helpful to have a section on differences from the FTC implementation, to help teams migrating.

zachwaffle4:
> It would also be helpful to have a section on differences from the FTC implementation, to help teams migrating.

I agree; Amanda's PR adds a special FTC section to the sidebar, so I was going to add some migration info there once that lands instead of putting program-specific transition info on a page for both programs.

source/docs/software/basic-programming/opmodes.rst
```diff
@@ -0,0 +1,269 @@
+# OpMode Framework
+
+## What is an OpMode?
+
+An opmode is an operator-selectable program that defines what the robot does during a particular mode of operation (autonomous, teleoperated, or utility). Robot code registers one or more opmodes for each mode; these appear in drop-down selectors on the Driver Station, filtered by the active mode. When the operator selects an opmode, a fresh instance of the corresponding class is constructed. The library then manages the full lifecycle: calling into it while disabled for pre-match setup, starting it when the robot is enabled, and tearing it down cleanly when the robot disables or the operator switches to a different opmode.
```
j5155: This is more wordy then necessary IMO. I would cut the entire second sentence (Robot code registers...), and I'm really not sure it's necessary to talk about the lifecycle.

source/docs/software/basic-programming/opmodes.rst
```diff
@@ -0,0 +1,269 @@
...
+
+Common use cases include:
```
j5155, line 7: Really like these examples for explaining it

source/docs/software/basic-programming/opmodes.rst
```diff
@@ -0,0 +1,269 @@
...
+
+- Multiple autonomous routines: follow different paths or perform different actions depending on match strategy
+- Multiple teleoperated behaviors: switch between drive styles (e.g. tank vs. arcade), different button mappings, or restricted controls for robot demonstrations or guest drivers
+- Testing and diagnostics: test the whole robot, an individual subsystem, a motor, or a sensor without modifying match code
+
+These use cases apply equally during competition matches and off-field testing.
```
j5155: I like this idea, but "These use cases apply equally" is a little hard to understand. I would just say something like "You can run one opmode at a time directly for your off field testing, and select an autonomous and teleop to run before each competition match."

source/docs/software/basic-programming/opmodes.rst
```diff
@@ -0,0 +1,269 @@
...
+
+   .. tab-item:: C++ (Source)
+      :sync: C++ (Source)
+
+      .. rli:: https://raw.githubusercontent.com/wpilibsuite/allwpilib/v2027.0.0-alpha-6/wpilibcExamples/src/main/cpp/templates/opmode/cpp/Robot.cpp
+         :language: c++
+         :lines: 5-25
+         :lineno-match:
+
+``OpModeRobot`` provides several overrideable lifecycle methods for robot-wide behavior:
```
j5155: I like this idea, but this sentence is hard to understand for new users. I would say something like "Methods in OpModeRobot run no matter the selected opmode."

source/docs/software/basic-programming/opmodes.rst
```diff
@@ -0,0 +1,269 @@
...
+
+- ``driverStationConnected()``: called once when the Driver Station first connects
+- ``robotPeriodic()``: called every loop iteration regardless of enabled state or selected opmode
+- ``disabledInit()`` / ``disabledPeriodic()`` / ``disabledExit()``: called when entering, during, and exiting disabled state
```
j5155: It would be nice to explain exactly when you enter and exit disable, and what enable and disable actually means for people coming from nothing or FTC. I like these explanations overall though.

source/docs/software/basic-programming/opmodes.rst
```diff
@@ -0,0 +1,269 @@
...
+         .. tab-item:: MyAuto.cpp
+
+            .. rli:: https://raw.githubusercontent.com/wpilibsuite/allwpilib/v2027.0.0-alpha-6/wpilibcExamples/src/main/cpp/templates/opmode/cpp/opmode/MyAuto.cpp
+               :language: c++
+               :lines: 5-33
+               :lineno-match:
+
+## OpMode Lifecycle
+
+**When the operator selects an opmode on the Driver Station**, a new instance of the opmode class is constructed. This is a fresh object — any state can be initialized in the constructor.
```
j5155: The second sentence of this (This is a fresh object...) doesn't make sense, and the em dash really makes it look AI generated. I would cut that sentence entirely and say something about "There's no separate initialization function. Any necessary state and hardware setup can happen in the class constructor directly." The first sentence is fine.

source/docs/software/basic-programming/opmodes.rst
```diff
@@ -0,0 +1,269 @@
...
+
+**While an opmode is selected but the robot is disabled**, ``disabledPeriodic()`` is called regularly at ``OpModeRobot#getPeriod()``. This is useful for updating dashboard displays, reading sensors, or previewing what the opmode is about to do. The library guarantees that ``disabledPeriodic()`` will be called at least once before the robot transitions to enabled, so any initialization logic placed here is guaranteed to run.
```
j5155: Totally fine explanation. It would be nice to mention the default period ideally.

source/docs/software/basic-programming/opmodes.rst
```diff
@@ -0,0 +1,269 @@
...
+
+**When the robot transitions from disabled to enabled**, ``start()`` is called exactly once.
```
j5155: Mention what you can use this for. For example, can you put blocking auto code in here like the FTC SDK? Should you?

source/docs/software/basic-programming/opmodes.rst
j5155: Also, in the example code's comment it says that disabled periodic runs when the bot recieves DS packets, but this says it runs at period. Which is it?

source/docs/software/basic-programming/opmodes.rst
```diff
@@ -0,0 +1,269 @@
...
+
+**While the robot is enabled**, ``periodic()`` is called repeatedly at ``OpModeRobot#getPeriod()`` (default 20 ms). Additional callbacks registered via ``addPeriodic()`` run at their own configured rates.
```
j5155: Totally fine start and I like the default period explanation. It would be nice to explain what you'd usually use this for (like teleop). Also, you could mention the hz as well (default 20 ms/50 hz).

source/docs/software/basic-programming/opmodes.rst
```diff
@@ -0,0 +1,269 @@
...
+
+**When the robot disables**, ``end()`` is called first, then ``close()`` (Java) or the object is destroyed (C++/Python). The object is never reused.
```
j5155: What's the difference between end and close, and why does close only exist in Java? What would you use this function for? What does it mean that the object is destroyed and never reused?

source/docs/software/basic-programming/opmodes.rst
```diff
@@ -0,0 +1,269 @@
...
+
+.. note:: Selecting a different opmode while the robot is enabled automatically disables the robot first, so ``end()`` is always called before the switch.
+
+**If a different opmode is selected while the robot is already disabled**, only ``close()`` is called, as the opmode was never started.
+
+After the old opmode is closed, a fresh opmode object is constructed based on the current DS selection. In teleop, autonomous, and utility modes the drop-down stays the same, so the same class is typically constructed again. In match mode (or when FMS-connected), only the selected autonomous opmode is constructed initially; once autonomous completes, the selected teleop opmode object is then constructed. Only one opmode object is ever alive at a time.
```
j5155: You should specify that the second sentence of this is only if you disable and re enable the same opmode. Also mention when exactly the new class construction and disabled periodic takes place for the same opmode. Is it immediately as soon as disable happens? How can you choose exactly when the opmode restarts, such as for pinpoint calibration?

source/docs/software/basic-programming/opmodes.rst
```diff
@@ -0,0 +1,269 @@
...
+
+## Accessing Robot Hardware
+
+Opmodes receive the ``Robot`` instance through their constructor. The library detects a single-argument constructor that accepts the ``Robot`` type and calls it automatically, passing the ``Robot`` object. If no such constructor exists, the no-argument constructor is used instead.
```
j5155: I like the idea of this paragraph but it's hard to understand for someone without a Java background. I'm also not sure it's necessary to explain the no argument constructor. The people writing opmodes need to understand that they need to have a constructor that takes a robot class. (Tbh I'm unconvinced that even allowing a no-argument constructor is a good idea.) Also, explain how they need to save the robot object to a variable.

source/docs/software/basic-programming/opmodes.rst
```diff
@@ -0,0 +1,269 @@
...
+public class DriveStraight extends PeriodicOpMode { ... }
+
+@Autonomous(name = "Score Cone", group = "Score")
+public class ScoreCone extends PeriodicOpMode { ... }
+
+@Autonomous(name = "Score Cube", group = "Score")
+public class ScoreCube extends PeriodicOpMode { ... }
+```
+
+The operator selects the desired routine in the DS before enabling. In match mode and while connected to the FMS, the operator selects both an autonomous and a teleop opmode. The driver station will automatically send the autonomous selection to the robot at the start of the match, and then send the teleop selection when autonomous ends.
```
j5155: I like this opmode example and setup explanation, but say OpMode here instead of routine. Also, does match modee really only work when connected to the FMS? I would also put "before a match" and the end of "selects both an autonomous and a teleop opmode".

source/docs/software/basic-programming/opmodes.rst
```diff
@@ -0,0 +1,269 @@
...
+
+## Custom Periodic Callbacks
+
+``PeriodicOpMode`` has an additional method, ``addPeriodic()``, allowing callbacks to run at rates different from the main loop:
```
j5155: Explain what this is for and the example use cases, particularly for offsetting it.

source/docs/software/vscode-overview/creating-robot-program.rst
```diff
@@ -6,6 +6,32 @@ Once everything is installed, we're ready to create a robot program.  WPILib com
...
+
+Documentation:
+[Java](https://github.wpilib.org/allwpilib/docs/beta/java/org/wpilib/framework/OpModeRobot.html)
+- [C++](https://github.wpilib.org/allwpilib/docs/beta/cpp/classwpi_1_1_op_mode_robot_base.html)
+
+Source:
+[Java](https://github.com/wpilibsuite/allwpilib/blob/v2027.0.0-alpha-6/wpilibj/src/main/java/org/wpilib/framework/OpModeRobot.java)
+- [C++](https://github.com/wpilibsuite/allwpilib/blob/v2027.0.0-alpha-6/wpilibc/src/main/native/include/wpi/framework/OpModeRobot.hpp)
+
+``OpModeRobot`` is the recommended base class for most users.  It structures robot code around individual *opmode* classes — one per autonomous routine, one per teleoperated configuration, and so on — annotated with ``@Autonomous``, ``@Teleop``, or ``@Utility`` (Java) or registered manually (C++).  The Driver Station displays a drop-down list of available opmodes for each mode, so operators can select routines without recompiling the robot program.  The ``Robot`` class holds shared hardware and subsystems, which opmodes receive through their constructors.
```
j5155: Em dash makes this look AI generated, use a comma instead
The rest of this explanation is totally fine though

source/docs/software/vscode-overview/creating-robot-program.rst
```diff
@@ -16,7 +42,7 @@ Source:
...
+The :code:`TimedRobot` class is an alternative base class that structures robot code around a single :code:`Robot` class with :code:`init()`, :code:`periodic()`, and :code:`exit()` methods for each mode (autonomous, teleoperated, utility), which are called by WPILib during the corresponding robot states. During these calls, your code typically polls each input device and acts according to the data it receives.  For instance, you would typically determine the position of the joystick and state of the joystick buttons on each call and act accordingly.  ``TimedRobot`` uses ``SendableChooser`` ([Java](https://github.wpilib.org/allwpilib/docs/beta/java/org/wpilib/smartdashboard/SendableChooser.html)/ [C++](https://github.wpilib.org/allwpilib/docs/beta/cpp/classwpi_1_1_sendable_chooser.html)) for autonomous routine selection via the dashboard.
```
j5155, line 45: Mention that FTC doesn't have sendablechooser

j5155 (changes requested):
Thanks for writing this.  I think this is overall a pretty great explanation already from the perspective of an experienced FRC and Java programmer, but I have some nitpicky feedback to make it understandable for new programmers. 

The biggest thing is that I'd love to see more explanations of the use cases for the functions so it's clear what they're actually for. Also, I would recommend avoiding semicolons and em dashes entirely, since they make the text look AI generated.

source/docs/software/vscode-overview/creating-robot-program.rst
sciencewhiz: This will need to be updated when SendableChooser is removed with the telemetry PR, doesn't need to be done here.

source/docs/software/basic-programming/opmodes.rst
zachwaffle4: this is an issue with the example comment, I just PR'd allwpilib (wpilibsuite/allwpilib#9156)

sciencewhiz:
> I didn't include any Python examples because I don't even know if we have Python examples with OpModeRobot rn

There's a Python OpMode example [PR](https://github.com/robotpy/mostrobotpy/pull/260), but it isn't a template like what's used for Java/C++ on this page.

source/docs/software/basic-programming/opmodes.rst
```diff
@@ -0,0 +1,271 @@
...
+            .. rli:: https://raw.githubusercontent.com/wpilibsuite/allwpilib/v2027.0.0-alpha-6/wpilibcExamples/src/main/cpp/templates/opmode/cpp/opmode/MyAuto.cpp
+               :language: c++
+               :lines: 5-33
+               :lineno-match:
+
+## OpMode Lifecycle
+
+**When the operator selects an opmode on the Driver Station**, a new instance of the opmode class is constructed. There is no separate initialization function; any necessary state and hardware setup can happen in the constructor directly.
+
+**While an opmode is selected but the robot is disabled**, ``disabledPeriodic()`` is called periodically at ``OpModeRobot#getPeriod()`` (default 20 ms). This is useful for updating dashboard displays, reading sensors, or previewing what the opmode is about to do. The library guarantees that ``disabledPeriodic()`` will be called at least once before the robot transitions to enabled, so any initialization logic placed here is guaranteed to run.
```
sciencewhiz, line 155: This is in the OpModeRobot, whereas everything else in this section is within the OpMode, right? That's worth clarifying

source/docs/software/basic-programming/opmodes.rst
```diff
@@ -0,0 +1,271 @@
...
+
+Callbacks are registered immediately at opmode construction and run even while the robot is disabled.
+
+.. warning:: Callbacks run regardless of enabled state. Any actuator commands inside a callback must be guarded with an ``isEnabled()`` check, or they will silently have no effect while the robot is disabled.
+
+## Migration from TimedRobot
+
+To switch to the OpMode framework from TimedRobot, replace per-mode methods in ``Robot`` (``autonomousInit``, ``teleopPeriodic``, ``utilityInit``, ``utilityPeriodic`` etc.) with separate ``@Autonomous``, ``@Teleop``, and ``@Utility`` opmode classes. Multiple opmodes of the same type replace ``SendableChooser``.
+
+``TimedRobot`` remains fully supported. Migration is not required.
```
sciencewhiz:
```suggestion
``.. note:: TimedRobot`` remains fully supported. Migration is not required.
```

source/docs/software/basic-programming/opmodes.rst
j5155: I was under the impression it was in OpMode. If it's not it should be added.

source/docs/software/basic-programming/opmodes.rst
```diff
@@ -0,0 +1,271 @@
...
+      MyTeleop::MyTeleop(Robot& robot) : robot{robot} {}
+
+      void MyTeleop::Periodic() {
+        robot.drive.ArcadeDrive(robot.joystick.GetY(), robot.joystick.GetX());
+      }
+      ```
+
+## Multiple OpModes and DS Selection
+
+Any number of classes can be annotated with the same type. All of them appear in the Driver Station's drop-down for that mode, organized alphabetically within their groups.
```
j5155: Nitpick: it would be nice here to have examples of opmodes not in groups, and the three types of opmodes. You could also  talk about how utility opmodes aren't available in match mode, but can be relabeled as teleops temporarily if needed such as for use in practice matches. Also, is it necessary to specify a name, or does it default to the class name?

source/docs/software/basic-programming/opmodes.rst
```diff
@@ -0,0 +1,271 @@
...
+.. tab-set::
+
+   .. tab-item:: Java
+      :sync: Java
+
+      ```java
+      public class MyAuto extends PeriodicOpMode {
+        public MyAuto(Robot robot) {
+          // Run an odometry update at 5 ms, offset 1 ms from the main loop
+          addPeriodic(robot.odometry::update, 0.005, 0.001);
```
j5155, line 253: Nitpick: are lambdas and this way of referring to functions explained anywhere? IMO it would be better to use the longer form () -> {function();} kind here so it's more obvious what it does.

j5155 (commented):
Couple last nitpicky things but overall this looks great.

source/docs/software/basic-programming/opmodes.rst
sciencewhiz: You're right

zachwaffle4: There are both robot-level and OpMode-level disabledPeriodic functions, which is why it's mentioned both here and in the "robot methods that are called regardless of current opmode" section

source/docs/software/basic-programming/opmodes.rst
sciencewhiz: Yes there is documentation on lambdas and method references. https://docs.wpilib.org/en/latest/docs/software/basic-programming/functions-as-data.html. Method references are also used in other places in the docs, like the zero to robot tutorial. https://docs.wpilib.org/en/latest/docs/zero-to-robot/step-4/creating-test-drivetrain-program-cpp-java-python.html#basic-drivetrain-example

As someone who isn't a professional software engineer, I find method references far more understandable then full lambda expressions.

j5155: Perfect! This is resolved then IMO (Github isn't letting me mark it as such)

sciencewhiz (approved):

PR #3341, opened 2026-08-07
Add new packet proposal for heartbeat and system information packet

ThadHouse:
This is a proposal for an updated packet setup. This does the following things. 

1. Most information is removed from the heartbeat, available for future use.
2. The heartbeat changes to be an FD frame when the device is configured for FD. This allows devices to know if they can send an FD frame by if this frame is received as FD or not. If there is a 2.0 device on the bus, this will automatically bring down the bus with error frames, instead of delaying until a vendor actually sends an FD frame.
3. Moves all of the information that was into the heartbeat into an information packet. This is send at a much slower rate. Additionally, this packet encodes the bus number. This will be used by the system to determine if users accidentally cross their buses.

### DIFF
--- a/source/docs/software/can-devices/can-addressing.rst
+++ b/source/docs/software/can-devices/can-addressing.rst
@@ -193,7 +193,9 @@ Systemcore provides a universal CAN heartbeat that any device on the bus can lis
 +-----------------------+------+--------------+
 | Description           | Byte | Width (bits) |
 +=======================+======+==============+
-| High Reserved Bits    | 7-5  | 23           |
+| Match Time (seconds)  | 7    | 8            |
++-----------------------+------+--------------+
+| High Reserved Bits    | 5-6  | 15           |
 +-----------------------+------+--------------+
 | FTC Motor Override    | 5    | 1            |
 +-----------------------+------+--------------+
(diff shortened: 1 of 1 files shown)


ThadHouse:
I realized we removed match time from the heartbeat, but that actually might be something useful. I can add that back in easily enough.

ThadHouse:
We should also create a low priority device type, and use that for the information packet.

source/docs/software/can-devices/can-addressing.rst
```diff
@@ -187,67 +188,99 @@ For CAN Nodes to be accepted for use in the FIRST System, they must:
...
++-----------------------+------+--------------+
+| Description           | Byte | Width (bits) |
++=======================+======+==============+
+| Match number          | 6-7  | 10           |
++-----------------------+------+--------------+
+| Replay number         | 6    | 3            |
++-----------------------+------+--------------+
+| Tournament type       | 6    | 3            |
++-----------------------+------+--------------+
+| Reserved.             | 4-5  | 16            |
```
sciencewhiz:
```suggestion
| Reserved              | 4-5  | 16           |
```

PR #3363, opened 2026-09-12
Remove readthedocssinglehtmllocalmedia builder

Gold856:
This builder doesn't do anything except for manually disable photofinish, which can also be done by setting `READTHEDOCS` environment variable to true.


sciencewhiz:
I didn't think that had to do with photofinish, but instead to override the way RTD built the downloadable files.

That's backed up by this comment:
```
# Force RTD to build zip with regular html builder, not single html
# The builder is preloaded after all extensions are loaded, so we can
# change the registered class here.
```

Gold856:
Yeah, I got confused over what this builder was for. Either way, now that we control the builder for RTD builds, the only thing this builder does is manually disable photofinish, but I think that's better done by setting the RTD env var. Updated PR description to match.

